PR #1, opened 2018-07-30
Custom block post type

lukecarbis:
Setup for a custom block post type.


rheinardkorf (changes requested):
Text domain translation strings need to match the plugin slug.  Plugin submission requirement.

Also, lets try to get menu slug consistent with the naming too.

rheinardkorf (approved):

PR #42, opened 2018-08-02
Update menu icon image

lukecarbis:


lukecarbis:
@rheinardkorf If I tried to update the SVG embedded inside `js/blocks/icons.js`, all my blocks break. Could you take a look?

Ideally, we'd use the SVG from `assets/images/admin-menu-icon.svg`.

lukecarbis:
Actually, never mind, I figured it out. Over to you for review.

rheinardkorf (approved):

PR #44, opened 2018-08-02
Feature/preview

### LINKED ISSUE #27: Block preview
In the editor, once the fields have been completed, a preview of the block is show.

This should include the ability to return to "edit" mode to change the fields again.

### LINKED ISSUE #30: Warning in Gutenberg if no template can be found.

rheinardkorf:
Where there is a will there is a way.  Or as we say in Afrikaans... "'n Boer maak 'n plan"

Resolves #30.
Resolves #27.


lukecarbis (changes requested):
This looks great!

ACB should look for the file `blocks/preview-$slug.php` first, and if it doesn't exist, then look for `blocks/block-$slug`.

This way, if I've got a complicated or wild block template that heavily depends on CSS, I can create a "preview" version of it for displaying in the editor.

`acb_template_part( $slug, $part )` already has a provision for this – just pass `'preview'` as the second param? Or make the second param work with an array of "parts", ordered by priority.

lukecarbis:
Resolves #30.
Resolves #27.

rheinardkorf:
@lukecarbis I like the idea for 'preview'.  I will push back on the array of types (for now) to reduce filesystem operations.  But please do add it as a separate issue because I do see a valid use case, but lets try keep this simple for now.
UPDATE: Hmm... I change my mind. We need it if we want to support `preview` and the `block` fallback. :/ But I will drop the `generic_file` feature... so no `block.php` or `preview.php` fallbacks.

rheinardkorf:
This now properly solves #30 as well.

lukecarbis (approved):

PR #52, opened 2018-08-05
Update control architecture.

rheinardkorf:
Abstract `control.js` out to instead use separate JS files for each control type.

These files now live in `js/blocks/controls/*.js` with `js/blocks/controls/index.js` with each control "registered" for use.

See `formControls` methods in `edit.js` and `inspector.js` to see how it gets used.

Adding new controls now simply requires a new component to be created in `js/blocks/controls` and having the new control added to `js/blocks/controls/index.js`.

Controls should be written in ES6, but this method does allow for using `createElement` to use ES5. Regardless, you will still need to transpile the source with npm.


js/blocks/controls/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+export default {
+	text: ACBTextControl,
+	textarea: ACBTextareaControl,
+	checkbox: ACBCheckboxControl,
+	radio: ACBRadioControl,
+	range: ACBRangeControl,
+	select: ACBSelectControl,
+	toggle: ACBToggleControl,
+}
```
lukecarbis, line 17: If I were a third-party plugin wanting to extend this list, how would I do that?

rheinardkorf:
You won't. Not yet.... but this gets us closer.

js/blocks/loader/edit.js
```diff
@@ -12,14 +12,16 @@ const formControls = ( props, block ) => {
 	const fields = simplifiedFields( block.fields ).map( field => {
 
 		// If its not meant for the inspector then continue (return null).
-		// if ( !field.location ) {
 		if ( !field.location || !field.location.includes( 'editor' ) ) {
 			return null
 		}
 
+		const controlFunction = controls[ field.control ];
```
lukecarbis: So this line is saying that whatever my control is called, there needs to be a matching JS function?

rheinardkorf: @lukecarbis correct... but it doesn't allow for extensibility in the way you might think. Lets sync.

lukecarbis (approved):

PR #55, opened 2018-08-06
Reorganise things.

lukecarbis:
I started working on abstracting the `controls`, but then realised that there was a bit of renaming and reorganising I wanted to do.

- Use underscores in classnames and namespacing
- Update the autoloader to replace underscores with dashes in directory names
- Use plural directory names (Views, not View)
- Simplify class names since they're namespaced so well (`Post_Types\Block_Post_Type` to `Post_Types\Block`).
- Move the `Custom_Field` and `Custom_Block` models to a models directory
- Start stubbing out a `Control_Abstract` class, for further work in a different branch

Open questions:
- Should we rename the `AdvancedCustomBlocks` namespace to `Advanced_Custom_Blocks`, or even `ACB`? (personally I like ACB, but I understand the minute potential for conflict)
- ~~Should we remove the `Plugin_Interface` and `Component_Interface`? I don't see the point in having these.~~


rheinardkorf (approved):
Rename all the things!

PR #57, opened 2018-08-07
Add ability to load blocks/blocks.json

rheinardkorf:
Users can create blocks by specifying a `blocks/blocks.json` file with the block definitions.  The file follows the same order of preference as template files... 1. filtered path first, 2. then current theme, 3. then parent theme if current theme is a child theme.

Although this was a roadmap item, I felt it was important to add this feature now.

I will allow me to define blocks in `blocks.json` which is not yet supported by the UI.  This will help me to continue the work required on the Gutenberg-JSX side of development.

### DIFF
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -145,5 +145,12 @@ function abc_locate_template( $template_names, $path = '', $single = true ) {
 		}
 	}
 
-	return $single ? array_shift( $located ) : array_values( array_unique( $located ) );
+	// Remove duplicates and re-index array.
+	$located = array_values( array_unique( $located ) );
+
+	if ( $single ) {
+		return array_shift( $located );
+	}
+
+	return $located;
 }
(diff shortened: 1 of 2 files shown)


php/helpers.php
```diff
@@ -96,37 +96,54 @@ function acb_template_part( $slug, $type = 'block' ) {
...
+		if ( file_exists( ABSPATH . WPINC . '/theme-compat/' . $template_name ) ) {
+			$located[] = ABSPATH . WPINC . '/theme-compat/' . $template_name;
+			if ( $single ) {
+				break;
+			}
 		}
 	}
 
-	return $located;
+	return $single ? array_shift( $located ) : array_values( array_unique( $located ) );
```
lukecarbis: My only suggestion is here on the last line of code for my review!

This line seems like Clever Code. I'd recommend rephrasing to multiple lines. Something like:

```
// Remove duplicates, and reindex numerically
$located = array_values( array_unique( $located ) );

if ( $single ) {
    // Return the first value in the array
    return array_shift( $located );
}

return $located;
```

lukecarbis (changes requested):

rheinardkorf:
@lukecarbis implemented the changes. I will merge based on the fact that this was the only point you called out, and provided a solution for too.

Issue #17, opened 2018-08-01
Toggle "True / False" Control Type


lukecarbis:

rheinardkorf:
This is called a ToggleControl in Gutenberg.

rheinardkorf:
Example JSON:

```
"is-it-on": {
        "name": "is-it-on",
        "label": "Is this thing on?",
        "control": "toggle",
        "location": "editor",
        "type": "boolean",
        "order": "2"
      },
```

Issue #16, opened 2018-08-01
"Radio" control type


lukecarbis:
- Choices

rheinardkorf:
Example JSON:
```
"i-choose": {
        "name": "i-choose",
        "label": "I choose...",
        "control": "radio",
        "location": "editor",
        "options": [
          {
            "label": "Option 1",
            "value": "option-1"
          },
          {
            "label": "Option 2",
            "value": "option-2"
          },
          {
            "label": "Option 3",
            "value": "option-3"
          },
          {
            "label": "Option 4",
            "value": "option-4"
          }
        ],
        "default": "option-3",
        "order": "3"
      },
```

Issue #14, opened 2018-08-01
Additional text control and textarea control field options


lukecarbis:
- Placeholder text
- Character limit

rheinardkorf:
Attribute for placeholder is unsurprisingly: `placeholder`.
The default HTML attribute of `maxlength` has been added to supply a character limit.

lukecarbis:
Resolved in #61 .

Issue #13, opened 2018-08-01
Additional global field options


lukecarbis:
- Required
- Default value
- Field instructions

rheinardkorf:
Required does not work for "SelectControl" "CheckboxControl", "ToggleControl" and "RadioControl".

rheinardkorf:
Required is a boolean: `"required": true` , false by default.
Default is supplied using `"default": <value>`.
Field instructions is called `help` in Gutenberg, so option is `"help": "Some help"`

lukecarbis:
That's a good point. Default value may not always work, either. I think that these should all be implemented outside of the abstract class, for two reasons:

1. More flexibility for controls to choose which "defaults" they include
2. More flexibility for controls to put options in the right order

PR #58, opened 2018-08-09
feature/extra-constols-jsx

rheinardkorf:
Addresses JS requirements for:
#10 
#11 
#12 
#13 
#14 
#16 
#17 

Further comments will be made on each of those tickets.


rheinardkorf:
Please see correction, this does not yet implement #15 (the image control).

php/endpoints/class-preview.php
```diff
@@ -60,8 +60,11 @@ public function mock_attributes( $block ) {
 				case 'textarea':
 					$value = '<' . wp_json_encode( $field['name'] ) . '>';
 					break;
+//				case 'toggle':
+//					$value = '<' . wp_json_encode( $field['name'] ) . '>';
+//					break;
```
lukecarbis: Why are these lines commented out?

lukecarbis (approved):
Looks good. If there's a good reason for the lines to be commented out, then go ahead and merge. Otherwise, uncomment, or just delete the lines?

php/endpoints/class-preview.php
rheinardkorf: I will actually remove this block. Its only there for when there are significant variances which I have not yet come across as I was building out the controls.

PR #61, opened 2018-08-10
Abstract controls (for real this time)

### LINKED ISSUE #13: Additional global field options

lukecarbis:
Resolves #13 and #14.

All this is New Block UI only, since I noticed you were doing work on the JS side. This PR won't cause the fields in the Gutenberg block to reflect the field settings.

On naming:
- A `Block` is a storage model for custom blocks, that has the ability to convert to and from JSON
- A `Block_Post` is the WP custom post, and handles all the UI rendering
- A `Field` is a storage model for the fields that an author sees when adding a block in the editor. `Block`s have a `$fields` param that accepts an array of `Field`s.
- A Control (see `Control_Abstract`) is the type of field, for example Text, Textarea, or Select. All possible Controls are registered by the `Block_Post`, and a `Field` has a `$control` parameter, that accepts a Control
- A `Control_Option` is a storage model for the various settings that can be assigned to a Control. For example, a Text Control would have a Placeholder `Control_Option`. Controls have an `$options` parameter, that accepts an array of `Control_Option`s.

One last note – `Field`s have an `$options` parameter, but this is a simple `$key => $value` array, it is not made up of `Control_Option` instances.


rheinardkorf:
> All this is New Block UI only, since I noticed you were doing work on the JS side. This PR won't cause the fields in the Gutenberg block to reflect the field settings.

I expected that and am aligned with it. When testing this PR I will test it in a merged state with the JS work.

php/blocks/class-block.php
```diff
@@ -9,6 +9,8 @@
 
 namespace Advanced_Custom_Blocks\Blocks;
 
+use Advanced_Custom_Blocks\Blocks\Controls\Control_Option;
```
rheinardkorf, line 12: Unused.

rheinardkorf (approved):

PR #62, opened 2018-08-10
Bugfix for "required" field being always checked

lukecarbis:


php/blocks/class-block.php
```diff
@@ -9,8 +9,6 @@
 
 namespace Advanced_Custom_Blocks\Blocks;
 
-use Advanced_Custom_Blocks\Blocks\Controls\Control_Option;
```
lukecarbis, line 12: Another unused Use.

rheinardkorf (approved):
Good catch.

Issue #60, opened 2018-08-09
Review escaping functions.


rheinardkorf:
E.g. https://github.com/rheinardkorf/advanced-custom-blocks/blob/develop/php/post-types/class-block-post.php#L173

Should be `esc_attr_e`.

lukecarbis:
Never heard of it! Can't find any docs either: 
https://duckduckgo.com/?q=%22esc_attr_er%22

lukecarbis:
Oh, you mean `esc_attr`, not `esc_html`.

pedro-mendonca:
Hi, just made a PR #64 that solves this issue.

rheinardkorf:
thanks @pedro-mendonca

PR #64, opened 2018-08-10
Review escaping functions according to Issue #60

pedro-mendonca:
Solves Issue #60

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -209,11 +209,11 @@ class Block_Post extends Component_Abstract {
 					<p>
 						<input
 							name="acb-properties-category-custom"
-							title="<?php esc_html_e( 'Custom Category', 'advanced-custom-blocks' ); ?>"
+							title="<?php esc_attr_e( 'Custom Category', 'advanced-custom-blocks' ); ?>"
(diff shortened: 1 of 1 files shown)


rheinardkorf:
@pedro-mendonca awesome! were there any others you found or are these the only ones?

pedro-mendonca:
@rheinardkorf I've made a search on all the code for `echo` and `esc_` within attributes and only found these.

rheinardkorf (approved):

pedro-mendonca:
One thing I found inconsistent was the use of either `echo esc_attr` or `esc_attr_e`. But that wasn't the intent of this PR so I didn't touch it :)

lukecarbis:
I just went through and checked every instance of `esc_attr`, `esc_attr_e`, `esc_html`, and `esc_html_e`, and they're all correct. 

@pedro-mendonca  the `_e` functions are used for translatable text.

pedro-mendonca:
@lukecarbis I meant that I've found cases where it's used `echo esc_attr` and others where its `esc_attr_e`, that is basically the same. Still, I didn't change those.

You can see an example in the commitdiff of this PR.

Issue #74, opened 2018-08-25
Fatal Error on Plugin Activation


RobStino:
`Plugin could not be activated because it triggered a fatal error.`

`Fatal error: Uncaught Error: Class 'AdvancedCustomBlocks\View\Admin_Menu' not found in /var/www/vhosts/acb.number11.co/httpdocs/wp-content/plugins/advanced-custom-blocks/advanced-custom-blocks.php:95 Stack trace: #0 /var/www/vhosts/acb.number11.co/httpdocs/wp-admin/includes/plugin.php(1897): include() #1 /var/www/vhosts/acb.number11.co/httpdocs/wp-admin/plugins.php(178): plugin_sandbox_scrape('advanced-custom...') #2 {main} thrown in /var/www/vhosts/acb.number11.co/httpdocs/wp-content/plugins/advanced-custom-blocks/advanced-custom-blocks.php on line 95`

- This is on a new WP install.
-  Plugin is v0.1 from the wp.org repo
- Default theme
- Not other plugins active
- PHP 7.1.20
- Has Let's Encrypt SSL Cert
- https://acb.number11.co/
- Apache

lukecarbis:
Does the same error occur on the current `develop` branch? We changed a few things that might be related.

RobStino:
This was on v1.0 that is up on wp.org. I'll test with the dev version.

timbutler:
Can confirm the same error occurs with the wp.org 0.1 version but the activation works if using the dev branch from GitHub.

rheinardkorf:
Thanks @timbutler

RobStino:
Interestingly/coincidently, my dev server is with Conetix, which is @timbutler 's hosting company. :D So the bug may have been related to the hosting env, but if the dev branch sorts it out, then we're all good.

PR #75, opened 2018-08-27
Admin menu refinement

### LINKED ISSUE #49: Add "Add New" to fly-out menu in WP Admin under Blocks
The experience with what you get when you hover over a CPT menu item where you get the "Add New" option, we should have that. :)

lukecarbis:
- Moves the menu definition to when the post type is registered, so that the default options such as "All Blocks" are included in the menu
- Makes some small changes to the labels
- Adds custom capabilities (easier to extend)

Resolves #49.


rheinardkorf:
@lukecarbis does this address #74 as well?

lukecarbis:
@rheinardkorf It might, but we should test that separately.

php/post-types/class-block-post.php
```diff
@@ -112,11 +112,14 @@ public function register_post_type() {
 			'show_ui'       => true,
 			'show_in_menu'  => true,
 			'menu_position' => 100,
+			// This SVG icon is being included from the plugin directory, so using file_get_contents is okay.
+			// @codingStandardsIgnoreStart
```
rheinardkorf: This is fine, but we need to keep our eyes open for when the WPCS minimum requirements bumps the PHPCS version.  As of PHPCS 3.2+ `@codingStandards*` comments will be replaced with `phpcs:*` comments and eventually dropped entirely in PHPCS 4.0.

rheinardkorf (approved):

PR #76, opened 2018-08-27
Empty fields fix

### LINKED ISSUE #28: Placeholder row when there are no fields

### LINKED ISSUE #71: PHP Notices when fields not provided.
Steps:
- Create new custom block.
- Provide title, description, slug.
- DON'T add any fields.
- Publish the block.

Result:
```
Notice: Undefined index: fields 
```
Need to check for empty value.

lukecarbis:
Add user notice when no fields exist, and make sure that at least an empty array is stored when saving a block.

Resolves #71, Resolves #28


js/admin.block-post.js
```diff
@@ -108,6 +113,12 @@
 		}
 	};
 
+	let blockFieldInit = function() {
```
rheinardkorf, line 116: At some stage we need to get this into a build process so that we can start using ES6+ syntax, but this works for now.

rheinardkorf (approved):

PR #78, opened 2018-08-29
Slug as a Block Property meta box input

### LINKED ISSUE #45: Move "Slug" meta box into the Properties meta box.
Might have to do this with a little JS magic.

lukecarbis:
Remove the slug meta box and put the slug input into the Block Properties meta box instead.

Resolves #45.


rheinardkorf (approved):

rheinardkorf:
Much better with the slug!

PR #81, opened 2018-08-30
PHPCS Fixes

lukecarbis:
@rheinardkorf Here's what you get for being picky about coding standards.

This resolves all the outstanding PHPCS issues currently in `develop`, before I go ahead and fix them in the branches.

The one exception is the `php/views/class-admin-menu.php` file, because I know that file is being deleted in #75.

This will likely create a few conflicts, but nothing I can't deal with.


lukecarbis:
If you think it's worth it, you might like to sort out the `npm run dev` process to include `.min` in the minified js and css files, so that they can be more easily excluded from PHPCS (rather than specifying them individually).

rheinardkorf (approved):

PR #85, opened 2018-08-31
Bugfix: Invalid characters in slug

### LINKED ISSUE #53: Can't use Number in Block Slug
I noticed that although I could successfully save a custom block with a number character in the slug, as well as set up the template file, the block was not accessible in the GB editor. After changing the slug (removing number), and updating the template to match I was able to access it in the editor.

lukecarbis:
Resolves #53.


lukecarbis:
Accidentally started with the wrong branch here, so please merge #78 first.

php/blocks/class-loader.php
```diff
@@ -111,6 +111,14 @@ public function dynamic_block_loader() {
 
 			$attributes['acb_block_name'] = $block_name;
 
+			// sanitize_title() allows underscores, but register_block_type doesn't.
```
rheinardkorf, line 114: Good catch!

rheinardkorf (approved):

PR #87, opened 2018-09-03
Fix a bug causing the save method to interfere with other post types

### LINKED ISSUE #86: Error when publishing a CPT
When tryinh to publish a CPT from another plugin, I get the WordPress admin error
`The link you have followed has expired`
When I disable this plugin, I am able to publish as normal.
Have tested on a few environments and observed the same issue.

lukecarbis:
Resolves #86.

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -621,6 +621,11 @@ class Block_Post extends Component_Abstract {
 			return $data;
 		}
 
+		// Exits script if not the right post type.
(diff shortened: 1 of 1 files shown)


php/post-types/class-block-post.php
```diff
@@ -621,6 +621,11 @@ public function save_block( $data ) {
 			return $data;
 		}
 
+		// Exits script if not the right post type.
```
rheinardkorf, line 624: That makes sense.

rheinardkorf (approved):

Issue #90, opened 2018-09-04
Lost our Dashicon Icons


RobStino:
With the update to the icons in GB (Material icons), it seems we've lost access, in the current format, to dashicons.

RobStino:
Note: Dashicons are still visible in the block editor, but are no longer seen in Gutenberg.
Note 2: Conversation in slack with @rheinardkorf showing the reasoning why the Core team moved away from Dashicons. We should probably follow suit.

PR #91, opened 2018-09-04
Add support for 'Material Icons'.

rheinardkorf:
Replaces Dashicons with Material Icons.

Resolves: #90


rheinardkorf:
@lukecarbis just requires the updates in the block editor now, so handing this ticket over to you.

php/blocks/class-loader.php
```diff
@@ -93,6 +93,16 @@ public function editor_assets() {
...
 			filemtime( $this->assets['path']['editor_style'] )
 		);
+
+		// Enqueue 'material-icons'.
+		if ( ! wp_style_is( 'material-icons', 'enqueued' ) ) {
+			wp_enqueue_style(
+				'material-icons',
+				'https://fonts.googleapis.com/icon?family=Material+Icons',
+				[],
+				'20180904'
```
lukecarbis, line 103: @rheinardkorf Is this the best date to use? Should we be including the material icons with the plugin, rather than loading them remotely?

rheinardkorf: Its the date we decided to add it 😄  I have mixed feelings about adding it to the plugin or via CDN.  Getting it from CDN means all the styles are loaded too.  Doing it locally means a few more enqueues.

rheinardkorf: We could do it via... https://google.github.io/material-design-icons/#setup-method-2-self-hosting

lukecarbis:
@rheinardkorf I've done some more work… can't get the outline theme to work though. The more I play, the more I'd like to move to SVGs instead of a font, but I can't figure out how to include them from the JSON file. 

One option would be to add all the SVGs from our list to `icons.js`, maybe.

lukecarbis:
@rheinardkorf I can't even find the SVG files for the outline versions in the Github repo. If this is too hard to figure out for you too, then I suggest removing the [WIP] tag and merging this PR with the standard icons, until the outline versions are available as a font.

Relevant issue: https://github.com/google/material-design-icons/issues/773

PR #96, opened 2018-09-05
Rename the plugin

### LINKED ISSUE #88: Rename the plugin
~~Advanced Custom Blocks~~ **Block Lab**

lukecarbis:
Resolves #88.

This PR will cause you to lose all your existing blocks.


lukecarbis:
After merging this, the Github repo should be renamed (moved?) and the plugin resubmitted to .org.

advanced-custom-blocks.php
```diff
@@ -1,20 +1,20 @@
...
+ * @package   Block_Lab
  * @copyright Copyright(c) 2018, Rheinard Korf
  * @license http://opensource.org/licenses/GPL-2.0 GNU General Public License, version 2 (GPL-2.0)
  *
- * Plugin Name: Advanced Custom Blocks
- * Plugin URI: https://github.com/rheinardkorf/advanced-custom-blocks
- * Description: Create Advanced Custom Blocks effortlessly with no Gutenberg development know-how required.
+ * Plugin Name: Block Lab
+ * Plugin URI: https://github.com/rheinardkorf/block-lab
+ * Description: Create Block Lab effortlessly with no Gutenberg development know-how required.
```
lukecarbis: Fix

js/blocks/controls/index.js
```diff
@@ -1,17 +1,17 @@
...
-import ACBRangeControl from './range'
-import ACBSelectControl from './select'
-import ACBToggleControl from './toggle'
+import BLTextControl from './text'
+import BLTextareaControl from './textarea'
+import BLCheckboxControl from './checkbox'
+import BLRadioControl from './radio'
+import BLRangeControl from './range'
+import BLSelectControl from './select'
+import BLToggleControl from './toggle'
```
lukecarbis: @rheinardkorf Should these be expanded to `BlockLabTextControl `, etc.?

php/blocks/class-loader.php
```diff
@@ -119,14 +119,14 @@ public function dynamic_block_loader() {
 		foreach ( $blocks as $block_name => $block ) {
 			$attributes = $this->get_block_attributes( $block );
 
-			$attributes['acb_block_name'] = $block_name;
+			$attributes['block_lab_block_name'] = $block_name;
```
lukecarbis: @rheinardkorf Is this too verbose? Can we rename it to something better?

php/post-types/class-block-post.php
```diff
@@ -24,7 +24,7 @@ class Block_Post extends Component_Abstract {
 	 *
 	 * @var string
 	 */
-	public $slug = 'acb_block';
+	public $slug = 'block_lab_block';
```
lukecarbis: This slug causes a lot of weird class names and IDs. I can't think of a better option though. `bl_block` is too short and weird looking, and it's inconsistent. Is `custom_block` too generic? I'd like to go with that.

php/post-types/class-block-post.php
```diff
@@ -221,26 +221,26 @@ public function enqueue_scripts() {
 	 */
 	public function add_meta_boxes() {
 		add_meta_box(
-			'acb_block_properties',
-			__( 'Block Properties', 'advanced-custom-blocks' ),
+			'block_lab_block_properties',
```
lukecarbis: These can be shortened to `block_properties`, `block_fields`, etc.

lukecarbis (commented):
Self-review.

php/post-types/class-block-post.php
RobStino: I think custom_block works. It's descriptive.
However, being so generic, is there risk of conflict with other plugins/themes?

lukecarbis: Yes, if there's a theme or plugin that also uses that name then there would be a conflict. Pretty low chance of that, though.

lukecarbis: Another (more generic) option is just `block`. What's the chance someone will want to use that CPT name in their theme or plugin?

RobStino: I'm not 100% on the use case of the slug and how it's used, so my only insight is that the more generic the slug the more likely it will be used.

RobStino: Another option could be lab_block

rheinardkorf: Yeah Block Lab doesn't lend itself to nice shortening.  As this is a Post Type, why not just call it `block_lab` for the slug?  I know its not very descriptive, but where else will this post type actually be used?

lukecarbis:
@rheinardkorf Ready for further review and merge.

RobStino:
Have switched to the `rename` branch and have tried creating a new block. The general block building experience (New block in WP Editor + Block template) is working fine, however the created block is not showing up as available Gutenberg.

RobStino (approved):
All looks good to me.

PR #105, opened 2018-11-03
Autoloader uncaught error

pharmarin:
Adding strtolower() to $namespace to avoid Uncaught Error.


lukecarbis:
@pharmarin Having trouble reproducing the error you mentioned. I can't see a problem with adding `strtolower`, but it would be good to check it out first. What's your development environment look like?

lukecarbis (approved):

PR #115, opened 2018-11-12
Remove the required field setting

### LINKED ISSUE #113: Maybe required doesn't make sense.
If I mark a field as required, I can still save my post. Do required fields even make sense? I'm not sure they do. Any thoughts @RobStino ?

lukecarbis:
Resolves #113


RobStino (approved):
Checked the branch locally and confirmed that the `required` field is gone from the Block edit screen.

PR #116, opened 2018-11-12
Prevent field from being added if the slug is blank

### LINKED ISSUE #107: Empty / Blank fields appear in GB Editor as text fields with no label
Noticed this after accidentally added a field, but didn't configure it, and hit publish.
The field appeared in the Gutenberg editor as a text field with not label or anything else.
Assuming it was a text field as this is the default field when adding a new field.
Noting that, without configuration, or at least a name/slug, the field values can't be interacted with in the template.
With this in mind, perhaps it makes sense to not show fields in the GB editor unless minimum requirements (name and slug) are configured.

lukecarbis:
Resolves #107

### DIFF
--- a/js/blocks/loader/fields.js
+++ b/js/blocks/loader/fields.js
@@ -12,6 +12,8 @@ const simplifiedFields = ( fields ) => {
 
 	for ( let fieldName in fields ) {
 
+		if ( '' === fieldName ) continue;
+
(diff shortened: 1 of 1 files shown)


RobStino (approved):
Checked branch locally and confirmed the empty/blank field was not added to the GB editor screen.

PR #117, opened 2018-11-12
Hide template not found error if user is not logged in

### LINKED ISSUE #108: The "Template not found" error displays for unauthenticated users
Noticed that the front end error for when no template file is found is displayed for unauthenticated users. Makes sense for logged in users, but perhaps not so much for others.

lukecarbis:
Resolves #108


RobStino (approved):
Tested locally and confirmed that:
- the error appears on the front end when logged in
- the error does not appear on the front end when not logged in

PR #119, opened 2018-11-13
Ensure that Gutenberg respects field types

### LINKED ISSUE #104: True/False state of toggle & Checkbox input not being passed through to function.
Screen cap walkthrough of issue -> https://d.pr/v/aywe4l

Regardless of if a toggle is switched to true or false, it's being picked up as a false by the function.

lukecarbis:
This one took quite a lot of work! Got there in the end, but please test carefully.

This fixes broken checkboxes, toggle controls, and range controls (which previously only returned their default value).

Resolves #104.

### DIFF
--- a/php/blocks/class-block.php
+++ b/php/blocks/class-block.php
@@ -123,7 +123,7 @@ class Block {
 
 		if ( isset( $config['fields'] ) ) {
 			foreach ( $config['fields'] as $field ) {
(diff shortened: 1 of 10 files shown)


RobStino (approved):
Checked out branch and observed that the value of the checkbox and toggle fields were now being passed correctly to the template.
Tested both true and false states.
Also checked that var_dump no longer gave is NULL. Which it didnt.

PR #136, opened 2018-11-15
Add better auto-slug generation

### LINKED ISSUE #134: Change the auto-slug generation behaviour
The automatic generation of the slug when adding the title for the first time makes sense. However after this point, it's not so ideal. 2 scenarios that come to mind that this is annoying for:

1. I edit the slug, i.e. shorten "Call to action" to "CTA". If I decide afterwards change the Title. I.e. capitilise the 'a' in 'action', the slug gets overwritten.
2. I have an established block that I've been using. I change the title, for whatever reason, and the slug gets updated. This actually "breaks" the block as the plugin is now looking for a different template file.

*Suggestion*
Auto-generate the slug _only_ when the title field is filled out for the first time.

lukecarbis:
Resolves #134

### DIFF
--- a/js/admin.block-post.js
+++ b/js/admin.block-post.js
@@ -11,14 +11,10 @@
 (function( $ ) {
 
 	$(function() {
+		blockTitleInit();
 		blockCategoryInit();
 		blockIconInit();
 		blockFieldInit();
-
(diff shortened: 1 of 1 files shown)


RobStino (approved):
Perfect!

PR #137, opened 2018-11-15
Add Save Draft button to unpublished blocks

### LINKED ISSUE #100: Add 'Save Draft' Button to Block Edit Page
Working with the plugin, there were a few times saving as draft felt like a natural step. Typically when the block I was working with was unfinished.
Reasons for adding this feature include:
- Save as draft is a natural workflow action in WordPress. It _feels_ like it should be there.
- Having Draft vs Published Blocks will help with organisation/filtering on the Block list page
- I imagine having Draft, Published and other statuses will help with exporting (future feature)
- Sometimes you want to remove an old block from the workflow (get it out of GB), but want to keep it around for future reference.
- If you are working on a block, the only way to leave the screen is to publish. If the block is unfinished, as could easily be the case for complex blocks, being published it will appear in GB in its unfinished state.

lukecarbis:
Resolves #100

### DIFF
--- a/css/admin.block-post.css
+++ b/css/admin.block-post.css
@@ -1,6 +1,7 @@
 #minor-publishing-actions,
 #misc-publishing-actions #visibility,
-#misc-publishing-actions .curtime {
+#misc-publishing-actions .curtime,
+#major-publishing-actions #delete-action {
 	display: none;
 }
 #block_fields .handlediv,
--- a/css/admin.block-post.css
+++ b/css/admin.block-post.css
@@ -1,7 +1,6 @@
 #minor-publishing-actions,
 #misc-publishing-actions #visibility,
-#misc-publishing-actions .curtime,
-#major-publishing-actions #delete-action {
+#misc-publishing-actions .curtime {
 	display: none;
 }
 #block_fields .handlediv,
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -245,6 +245,12 @@ class Block_Post extends Component_Abstract {
 	 * @return void
 	 */
 	public function remove_meta_boxes() {
+		$screen = get_current_screen();
+
+		if ( ! is_object( $screen ) || $this->slug !== $screen->post_type ) {
+			return;
+		}
+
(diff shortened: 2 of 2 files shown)


lukecarbis:
Note: This solution moves the Save Draft button into the main actions area (instead of above it), and removes the "Move to Trash" link normally found next to the Publish button, since I couldn't find a nice place for it to go.

RobStino:
- I think we need the "move to trash" option somewhere
- I had a look how normal posts handle this and noticed 2 things:
1. The "Save Draft" button is up the top.
2. This PR breaks this in the classic editor for posts (maybe pages as well) by adding the "Save Draft" at the bottom as well. See the below screen cap.
image

lukecarbis:
1. The Save draft button looks bad in the top, since we're only showing status (I tried it)
screen shot 2018-11-15 at 2 14 26 pm

2. I'll fix the classic editor thing – that was an oversight, but nice pickup

lukecarbis:
Fixed the classic editor problem – any advice on where to put the Move to Trash link? You can still do this from the main list table, so I'm not really worried about it not being available here.

RobStino:
It does look odd, but my gut is to stick with standard UI.
The other option is adding it as another li under `Status`. This could work as it's just text.

lukecarbis:
I don't think it's an important enough feature to warrant additional UI clutter. Let's merge without it, and revisit if needed later.

lukecarbis:
Wait, I've got an idea…

css/admin.block-post.css
```diff
@@ -1,6 +1,7 @@
 #minor-publishing-actions,
 #misc-publishing-actions #visibility,
-#misc-publishing-actions .curtime {
+#misc-publishing-actions .curtime,
```
RobStino: Is hiding a UI action component by CSS a good idea/best practices?
_actually reviews code_

lukecarbis: Unfortunately, it's the only way. The other option would be to create our own version of the meta box, which would be a mess of duplicate code (and ongoing support).

lukecarbis:
Okay, @RobStino – try it now. I think it's a good compromise.

RobStino (approved):

PR #148, opened 2018-11-17
mixin exposing the full block parameters like labels, types …

violy:
Hi there.

I've discover recently Guntenberg, an your plugin. It's helpful ! 
Thank's

So I've to simply retrieve the label, an the only available mixin concerns _value_ of fields. 
And so, I've expose a new mixin `block_params()` returning all params of current block, available in the template. 

Usage example in my theme block template:
```php
<?php
	$block = block_params();

	foreach ($block['fields'] as $field_name => $field) {

		?>
		<h2><?= $field['label'] ?></h2>
		<p><?php block_field( $field_name ) ?></p>
		<?php

	}
```

Thank you. 
Arthur.

PR #151, opened 2018-11-19
Feature/148 violy

lukecarbis:
Based on @violy's work in #148. Added methods for viewing block config, or a fields config.

Note – after merging this PR, docs must be added for the `block_config` and `block_field_config` functions.

### DIFF
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -66,13 +66,28 @@ function block_value( $key ) {
 }
 
 /**
- * Convenience method to return the block object.
+ * Convenience method to return the block configuration.
  *
- * @return mixed|null
+ * @return array
  */
-function block_params( ) {
-	global $block_lab_block;
-	return $block_lab_block;
+function block_config() {
+	global $block_lab_config;
+	return $block_lab_config;
+}
+
+/**
+ * Convenience method to return a field's configuration.
+ *
+ * @param string $key The name of the field as created in the UI.
+ *
+ * @return array|null
+ */
+function block_field_config( $key ) {
+	global $block_lab_config;
+	if ( ! isset( $block_lab_config['fields'][ $key ] ) ) {
+		return null;
+	}
+	return $block_lab_config['fields'][ $key ];
 }
 
 /**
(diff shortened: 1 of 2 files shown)


violy:
Thank's ;)

php/helpers.php
```diff
@@ -65,6 +65,31 @@ function block_value( $key ) {
...
+ * Convenience method to return a field's configuration.
+ *
+ * @param string $key The name of the field as created in the UI.
+ *
+ * @return array|null
+ */
+function block_field_config( $key ) {
+	global $block_lab_config;
+	if ( ! isset( $block_lab_config['fields'][ $key ] ) ) {
+		return null;
```
rheinardkorf, line 88: @lukecarbis should this return a new WP_Error or an empty array instead?  `null` feels a bit uneasy.

rheinardkorf (changes requested):
Consider alternative to returning `null`.

php/helpers.php
lukecarbis: @rheinardkorf I chose `null` for the following reasons:

1. `block_field()` also returns `null` if the field value is not set
2. I don't want to return a `WP_Error`, since I believe it's reasonable to call `block_field_config()` to check whether the config is set or not. e.g. `if ( isset( block_field_config( 'foo' ) )`
3. I don't want to return `false`, since that might be mistaken for an expected value (for example, an unchecked checkbox). This is more true for `block_field()`, but the same _sort of_ applies here

I'm open to other suggestions, though.

rheinardkorf: We'd have to document this then. Just to avoid warnings and possible exceptions that can be thrown because of the `null`.  If you're happy with that course of action I'll update my review.

RobStino (approved):
Tested with the follow and all looks good.
`<?php echo '<pre>' . print_r( block_config(), true ) . '</pre>'; ?>`
`<?php echo '<pre>' . print_r( block_field_config( 'text' ), true ) . '</pre>'; ?>`

rheinardkorf (approved):

php/helpers.php
lukecarbis: I will make sure I call this out in the documentation.

PR #154, opened 2018-11-21
Fix problems with quotes and slashes in the post title or field data

lukecarbis:
This should resolve all of these issues once and for all!

Here's some of the strings I used to test:
> "keywords",'keywords',\keywords\

> Testing: A "complex' title* 💅🏿 with special&nbsp; \&כעיח ©haracters

> a ' b " c \ d \' e \" f \\ g

Resolves #154

### DIFF
--- a/php/blocks/class-block.php
+++ b/php/blocks/class-block.php
@@ -148,8 +148,6 @@ class Block {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


RobStino (approved):
Tested and all looks good.

PR #177, opened 2018-11-27
[WIP] Bugfix/incorrect field value returned

### LINKED ISSUE #160: Toggle fields not returning the correct result on frontend
Noticed this when testing today. The same is _probably_ true for checkboxes.

### LINKED ISSUE #162: Calculation with block_value() - Placeholder instead of value!
Hi,

I've added a "Range" field with values 0 to 10. I want use this value for further calculations in the template file. Unfortunately this doesn't work. I'm not shure if this is an problem in my enviroment, or a template placeholder replacement issue.

Example: I selected the range slider value 8;

If I use `block_field( 'rating');` in the template file, everything looks fine. **Output is 8.**

Also `echo block_value('rating');` just output the correct value. **It's again 8.**

Problem: I can't calculate with the value. 

`echo block_value('rating') + 2;` will output 2. 

The error log contains an entry "PHP Warning:  A non-numeric value encountered".

If I output the value using WordPress' esc_html(), It shows a placeholder instead of the real value.

`echo esc_html(block_value('rating'));` outputs **["rating"]**

lukecarbis:
(Maybe) Resolves #160.
(Maybe) Resolves #162.

@RobStino Can you please test this thoroughly against those issues?

There is one outstanding issue, which is going to be hard to solve, and should be addressed separately:

If I use `block_value()` to evaluate the content inside a block template, I will always get the block's name as the result when Previewing in Gutenberg (it should work fine on the frontend).

For example, in preview mode, `$toggle = block_value( 'toggle-me' ) ? 'Yes' : 'No'` will always return 'Yes' because the result will be a string (which evaluates as true).


RobStino:
@lukecarbis re the fix for #160, I'm not sure I understand the issue and how I should go about testing. See this screen recording for what I have done. https://d.pr/v/uUswqN

RobStino:
@lukecarbis for issue #162 I'm seeing a problem with what's being displayed in the editor.
Screen recording --> https://d.pr/v/wx2ucc

RobStino (approved):

PR #180, opened 2018-11-29
Workaround for mock attributes being replaced with JS

lukecarbis:
Okay…

So the way things currently work is that when a block is updated in the admin, it requests a "preview" version of the template. The preview version is unaware of the block's current state, so it can't output the values. Instead it inserts placeholders into the template (`["block-name"]`), which are subsequently replaced by the correct values once they're returned.

There's a fundamental problem with this approach, which is that I can't use any of the fields in my block programmatically. For example, suppose I have a Toggle block called "sunshine". My template looks like this:

```PHP
if ( block_value( 'sunshine' ) ) {
    echo "Buon Giorno!";
} else {
    echo "Buona Serra!";
}
```

When this is sent to the template, the value of `sunshine` is unknown, so the template essentially sees:

```PHP
if ( '["sunshine"]' ) {
    echo "Buon Giorno!";
} else {
    echo "Buona Serra!";
}
```

Which of course means we evaluate as true every time, and the sun never sets on Rome.

This PR sends the field values along with the template request, so that we never even need to use mock attributes. By taking this approach, I was able to remove the `updatePreview` component entirely.

(As an aside, we were previously updating the preview with every keystroke and value change in the block's fields. This PR stops doing that, and only updates the preview after all the settings have been changed).

**Note 1**: This is a significant refactoring, so please review and test thoroughly
**Note 2**: Please re-save all your blocks before testing them.
**Note 3**: Don't forget to `npm run build`


js/blocks/controls/checkbox.js
```diff
@@ -18,9 +16,6 @@ const BlockLabCheckboxControl = ( props, field, block ) => {
 				attr[ field.name ] = checkboxControl
 				setAttributes( attr )
 			}}
-			onBlur={
```
lukecarbis, line 21: This is removed entirely since it seemed like overkill to fetch a new preview via the REST API every time a value changed.

js/blocks/loader/edit.js
```diff
@@ -34,14 +33,22 @@ const formControls = ( props, block ) => {
 }
 
 const previewData = ( props, block ) => {
-	if ( typeof props.attributes.block_template !== 'undefined' && props.attributes.block_template.length > 0 ) {
```
lukecarbis, line 37: I believe this line was intended to prevent the preview from updating when the block was blurred, but since we're no longer updating with every single keystroke and change in the field settings, I removed the line.

lukecarbis: I could be wrong about its purpose, though. In the end, this code doesn't work if this line exists.

js/blocks/loader/edit.js
```diff
@@ -34,14 +33,22 @@ const formControls = ( props, block ) => {
...
+
+		if ( 'block_template' === attribute || 'block_preview' === attribute ) {
+			continue;
+		}
+
+		params.append( attribute, props.attributes[ attribute ] );
 	}
 
-	wp.apiFetch( { path: `/block-lab/v1/block-preview?slug=` + block.name } ).then(
+	wp.apiFetch( { path: '/block-lab/v1/block-preview?slug=' + block.name + '&' + params.toString() } ).then(
```
lukecarbis, line 49: Is there a better way of sending data along with the API fetch besides just appending it to the URL with `params.toString()`?

RobStino:
Great work Luke!
I've tested using the same setup as yesterday afternoon. See the recording below for what I observed. From a QA point, all looks good. Over to Rheinard or Koop for code review.
https://d.pr/v/Su3zpq

camya:
Just tested it too. This works really good.

RobStino (approved):
QA done.
Code review done to the best of my ability.  :)

PR #185, opened 2018-12-05
New controls

lukecarbis:
- Number
- Email
- URL
- Multiselect


kopepasah:
@lukecarbis this looks good to me.

kopepasah (dismissed):

RobStino:
Hey @lukecarbis 
Noticed that when you enter an invalid entry in to the URL field, you are unable to do anything else until you fix the validation issue. This feels a bit _off_, but want your thoughts before I proceed.
Here's a demo of what I mean. https://d.pr/v/vK9IWu

kopepasah:
@lukecarbis looks like this is happening because of the `setTimeout` on blur. Is the intention here to validate a user's input?

lukecarbis:
@kopepasah @RobStino No, the timeout is to get around a weird Gutenberg focus behaviour. It's a bit complicated, but the issue occurs because the `reportValidity` function focuses the field. So even without the weird workaround in effect here, this happens:

1. User blurs the field
2. Browser checks if valid
3. If not valid, browser focuses the field

I'd love to find a better way, but since there's no "form" element at work here, I'm struggling to find one.

kopepasah:
@lukecarbis @RobStino I pushed up some changes which display a nicer error. Let me know what you think.

text-control-errors

js/blocks/controls/email.js
```diff
@@ -1,31 +1,46 @@
...
 	const attr = { ...props.attributes };
+
+	/**
+	 * Sets the Error Class for the Text Control.
+	 *
+	 * @param {string} element The HTML element.
+	 *
+	 * @return {void}
+	 */
+	const setErrorClass = ( element, valid ) => {
```
lukecarbis, line 16: Great call. Let's move this to it's own component and import it!

kopepasah: @lukecarbis I want to hold off on this, as I have a different idea of abstracting the entire TextControl into a wrapper component.

RobStino:
Great solution @kopepasah 
2 bugs
1. For some weird-ass reason, you can enter the letter `e` in the number field. :shrug:
2. I feel like the red border should persist if the block is closed then reopened.
Video explanation -> https://d.pr/v/UmRK79

RobStino:
TIL that the letter `e` is part of the spec for the html input field type `number`.

kopepasah:
> 2. I feel like the red border should persist if the block is closed then reopened.

@RobStino this will take a bit of reworking how the control is rendered and I have not dove deep into that section yet. Happy to give it a go, but don't want to hold up these controls for now.

RobStino:
@lukecarbis If you're happy with the above, so am I. Will merge with your ok.

RobStino:
Sorry. Further testing.
Unexpected URL Field behaviour with the template.
Demo of issue -> https://d.pr/v/QjWo3t
Basically block_field() on a URL field displays the "Site Name" on the front end and then links to the site.
E.g. I entered `https://getblocklab.com` in the block editor, and on the front end it gave me `Block Lab` that was linked to the URL.

lukecarbis:
@RobStino I think that might be a WordPress "feature". WP tries to be smart and convert urls to links.

Out of interest, could you try this in your template, see what happens?

```HTML+PHP
<?php echo block_value( 'url' ); ?>

<a href="<?php block_field( 'url' ); ?>">This text shouldn't change</a>

<pre><?php block_field( 'url' ); ?></pre>

<code><?php block_field( 'url' ); ?></code >

<p>The url is: <?php block_field( 'url' ); ?></p>
```

RobStino:
Intersting...
The above code gives me what I'd expect. See the screenshot for the frontend.
Backend is good as well.
image
If that "feature" is expected behaviour by WP, then everything looks ok.

RobStino (approved):
QA tested and all looks good.

PR #188, opened 2018-12-05
Add Select2 Functionality.

kopepasah:
This functionality will be needed when searching through a large list (like users). It is also implemented for the Field Type control, because it will grow over time.


RobStino:
Hey @kopepasah 
To confirm, this isn't adding a new `select2 field type` to block building is it? Just adding the functionality so when we want to implement it, we can. I.e. I can see it being used with the `field type` option when block building.

lukecarbis:
@kopepasah Thanks so much for your attention here! I'm unsure we actually need select2, though.

- There's never going to be _that many_ control types
- For things like searching for users or posts in Gutenberg, we should use the [autocomplete component](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/autocomplete)

kopepasah:
> Hey @kopepasah
> To confirm, this isn't adding a new `select2 field type` to block building is it? Just adding the functionality so when we want to implement it, we can. I.e. I can see it being used with the `field type` option when block building.

@RobStino that is correct, just implemented for now.



> @kopepasah Thanks so much for your attention here! I'm unsure we actually need select2, though.
> 
> * There's never going to be _that many_ control types
> * For things like searching for users or posts in Gutenberg, we should use the [autocomplete component](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/autocomplete)

@lukecarbis that makes sense. Will investigate when building out the Users control.

lukecarbis:
Closing for now, then. We can add `select2` later if we find we need it.

PR #189, opened 2018-12-06
Simplify WordPress version check

lukecarbis:
- [x] Simplify the WordPress version check, which was failing when trying to do a `version_compare` of `5.0-RC3-43968`. Instead
- [x] Remove mentions of the Gutenberg plugin, since it's not going to be relevant for much longer


kopepasah:
@lukecarbis looks good to me.

kopepasah (approved):

PR #190, opened 2018-12-06
Bugfix/178 merged html tags

### LINKED ISSUE #178: HTML Tags being merged in the Gutenberg Editor
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that you're using the latest version of Block Lab: https://github.com/getblocklab/block-lab/releases
    * Checked the FAQs for common solutions: https://github.com/getblocklab/block-lab#frequently-asked-questions
    * Checked that your issue isn't already filed: https://github.com/getblocklab/block-lab/issues

### Description

Observing that when 2 HTML elements of the same type (e.g. 2 H1 siblings) are being merged together in the Gutenberg editor. They are fine on the front end i.e. on seperate lines, but are merged together in a single line in the editor.

Screen recording of issue--> https://d.pr/v/bqOBgz

### Steps to Reproduce

1. Create 2 text fields
2. Create template that renders these 2 fields as siblings and uses the same html tag (i.e H1) for each
3. Check the editor preview

**Expected behavior:** 
The two entries to be on their own lines as expected of H1's

**Actual behavior:** 
The 2 H1's are merged in to one.

**Reproduces how often:** 
100%. Tested on `develop` branch

### Versions

Latest of all

### Additional Information

NA

lukecarbis:
Resolves #178 

This is a big one!

This replaces the `RichText` component hack we were using, which was causing strange behaviour with `p` tags. Using the RichText's previewData method, we made a REST API call to an endpoint we'd setup, and that endpoint would return the correct result.

In this PR, I use the `ServerSideRender` component instead, which comes with a built-in fetch of the default template. This means we can strip out all that login around Previews! It even comes with its own fancy loading screen.

In short, this is the way we should have done it from the start, except it wasn't available when we were first putting this together.


kopepasah (dismissed):
@lukecarbis this looks good to me and can confirm all previous blocks rendered as expected in the editor.

RobStino:
Thanks @lukecarbis for working on this one!
Not sure if this is a bug, but I was getting some unexpected behaviour with the (new) select field.
With the single select field, I'm not getting any returned values in the editor or on the front end. I am with the new multiselect though.
Demo of what I'm seeing -> https://d.pr/v/otlsUj

The bugfix is working well though! No more merged tags.

RobStino:
Just realised the background noise on that recording is terrible! Let me know if I need to re-record.

RobStino (approved):
QA done and all good.

PR #199, opened 2018-12-11
Add empty default value for selects

### LINKED ISSUE #196: Multi-Select Control: Automatically adds "Default Value"
Hi,

1. I create a "Multi-Select" field and add choices. 
2. I don't want to add any default values.

### Problem:

At the moment, the system **automatically adds a default value**, if there is no one assigned.

### Expected behavior:

A user can setup a "Multi-Select" field without default value.

### Example data for choices:

```
news : News
info : Info
update : Update
```

lukecarbis:
Resolves #196

### DIFF
--- a/js/blocks/controls/select.js
+++ b/js/blocks/controls/select.js
@@ -3,6 +3,15 @@ const { SelectControl } = wp.components;
 const BlockLabSelectControl = ( props, field, block ) => {
 	const { setAttributes } = props
 	const attr = { ...props.attributes }
+	const { __ } = wp.i18n;
+
+	if ( '' === field.default ) {
+		field.options = [
+			// @see https://github.com/WordPress/gutenberg/issues/11270 Disabled attribute not currently supported.
+			{ label: __( '– Select –', 'block-lab' ), value: '', disabled: true },
+			...field.options
+		]
+	}
 
 	return (
 		<SelectControl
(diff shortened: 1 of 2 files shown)


js/blocks/controls/select.js
```diff
@@ -3,6 +3,15 @@ const { SelectControl } = wp.components;
 const BlockLabSelectControl = ( props, field, block ) => {
 	const { setAttributes } = props
 	const attr = { ...props.attributes }
+	const { __ } = wp.i18n;
+
+	if ( '' === field.default ) {
+		field.options = [
+			// @see https://github.com/WordPress/gutenberg/issues/11270 Disabled attribute not currently supported.
+			{ label: __( '– Select –', 'block-lab' ), value: '', disabled: true },
```
lukecarbis, line 11: This "– Select –" label was copied from WordPress core settings pages.

RobStino:
Tested the PR branch,
- Noted that the `- Select -` option was the first available option in the single select field
- When `- Select -` is "selected" in the _select_ field (so meta!), nothing is output on the frontend or in the editor when using the `block_field()` function. E.g. `block_field( 'select' );`
- Could still set a default value as expected.
Unless I should be testing for something else, all looks good to me.

RobStino (approved):
Tested and reviewed

PR #201, opened 2018-12-11
Add Color Picker control.

kopepasah:
Adds a color picker control.

### DIFF
--- a/js/blocks/controls/color-picker.js
+++ b/js/blocks/controls/color-picker.js
@@ -11,23 +11,33 @@ const BlockLabColorPickerControl = ( props, field, block ) => {
 	// Set a default color of White, if nothing is set.
 	attr[ field.name ] = attr[ field.name ] ? attr[ field.name ] : '#fff';
 
+	const currentColor = {
+		display: 'inline-block',
+		width: '120px',
+		height: '30px',
+		borderRadius: '5px',
+	}
+
 	return (
 		<BaseControl
 			label={ field.label }
 			help={ field.help }
 			className="block-lab-color-picker"
 		>
-			<span class="block-lab-color-picker__current-color" style={ {
-				display: 'inline-block',
-				width: '120px',
-				height: '30px',
-				marginRight: '15px',
-				borderRadius: '5px',
+			<span className="components-color-picker__alpha block-lab-color-picker__current-color" style={ {
+				...currentColor,
+				overflow: 'hidden',
 				border: `1px solid ${ Color( attr[ field.name ] ).darken( 0.2 ) }`,
-				backgroundColor: attr[ field.name ],
-			} }></span>
+				marginRight: '15px',
+			} }>
+				<span style={ {
+					...currentColor,
+					backgroundColor: attr[ field.name ],
+				} }></span>
+			</span>
 			<Dropdown
 				className="components-color-palette__item-wrapper components-color-palette__custom-color block-lab-color-picker__custom-color"
+				contentClassName="block-lab-color-picker__picker"
 				renderToggle={ ( { isOpen, onToggle } ) => (
 					<Tooltip text={ customColorPickerLabel }>
 						<button
@@ -45,7 +55,7 @@ const BlockLabColorPickerControl = ( props, field, block ) => {
 					<ColorPicker
 						color={ attr[ field.name ] }
 						onChangeComplete={ value => {
-							attr[ field.name ] = value.hex;
+							attr[ field.name ] = `rgba( ${ value.rgb.r }, ${ value.rgb.g }, ${ value.rgb.b }, ${ value.rgb.a } )`;
 							setAttributes( attr )
 						} }
 					/>
--- a/js/blocks/controls/color-picker.js
+++ b/js/blocks/controls/color-picker.js
@@ -1,0 +1,58 @@
+import Color from 'color';
+
+const { __ } = wp.i18n;
+const { BaseControl, ColorPicker, Dropdown, Tooltip } = wp.components;
+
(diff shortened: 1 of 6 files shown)


RobStino:
@kopepasah 
Testing. Noticed a few bugs.
1. There is too much padding on the rgba input fields. Means you can't see the input values.
image
2. The input fields & the arrow button at the end are different heights. 
image
3. The position of the arrow button that cycles through the color types shifts around depending on the width of the other input fields. Expecting the button to remain fixed to the right.
4. The little white arrow at the top of the modal is about 1px out of line with the top of the modal. Expecting it to be flush, where it kinda overlaps.
image
5. Toggling through the different color types doesn't change the output. The output, in the editor and frontend is always HEX. This is particularly noticeable with rgba as the alpha value doesn't do anything.

1 & 5 are the only real bugs. 2-4 are minor design flaws.

In general, this is a great start! Really cool to see you pull together this particular implementation. I'm interested in whether you still think it's a good idea doing this as you mentioned concern with that it departs from the native controls.

kopepasah:
@RobStino addressed numbers 1 and 5 in the two new commits.

RobStino:
@kopepasah 
Looking good now. I think I've found one more bug, and have one more thought.
**Bug(?)**
 If you set the color with HSL, it switches back to RGBA. The color is correct, but I feel like if you choose HSL, that should persist through the experience. Including what is displayed next time you open the control as well as what is outputted in the preview and and the frontend. Noting however that this same issue occurs with the native color control, so may not be something we solve ourselves.
**Thought**
When adding the field to the block, should you be able to set a default?

lukecarbis:
@RobStino This is a known issue in Gutenberg. https://github.com/WordPress/gutenberg/issues/11791

package.json
```diff
@@ -38,5 +38,7 @@
     "uglifyjs-webpack-plugin": "^1.2.7",
     "webpack": "^3.11.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "color": "^3.1.0"
```
lukecarbis, line 42: Is this required? Isn't a colour picker included in WordPress by default? https://github.com/WordPress/gutenberg/tree/master/packages/components/src/color-picker

js/blocks/controls/index.js
```diff
@@ -22,4 +23,5 @@ export default {
 	select: BlockLabSelectControl,
 	multiselect: BlockLabMultiselectControl,
 	toggle: BlockLabToggleControl,
-}
\ No newline at end of file
+	'color-picker': BlockLabColorPickerControl,
```
lukecarbis, line 26: Minor gripe here, but can we wrap all the keys in `'single quotes'` if we're going to wrap `'color-picker'`?

php/blocks/controls/class-color-picker.php
```diff
@@ -0,0 +1,50 @@
...
+		$this->settings[] = new Control_Setting(
+			array(
+				'name'     => 'help',
+				'label'    => __( 'Field instructions', 'block-lab' ),
+				'type'     => 'textarea',
+				'default'  => '',
+				'sanitize' => 'sanitize_textarea_field',
+			)
+		);
+	}
```
lukecarbis, line 58: There should also be a `default` setting.

js/blocks/controls/index.js
lukecarbis: Actually, a better, simpler solution would be to refer to it simply as `color`. e.g.:
`color: BlockLabColorControl`

This would require renaming the PHP class as well.

lukecarbis (changes requested):
Great feature! Some minor suggestions included here.

js/blocks/controls/color-picker.js
```diff
@@ -0,0 +1,68 @@
...
+		borderRadius: '5px',
+	}
+
+	return (
+		<BaseControl
+			label={ field.label }
+			help={ field.help }
+			className="block-lab-color-picker"
+		>
+			<span className="components-color-picker__alpha block-lab-color-picker__current-color" style={ {
```
lukecarbis, line 27: Could this be replaced with the `ColorIndicator` component? https://github.com/WordPress/gutenberg/tree/master/packages/components/src/color-indicator

package.json
kopepasah: @lukecarbis this package is not the color picker, but rather used to adjust the darkness of the border in the select color display, this way the border can be a similar color variation.

js/blocks/controls/index.js
kopepasah: I think your first suggestion is better, as there may be other uses of the word color in the future. Since this is specific a `color-picker`, I think that that name is the best.

js/blocks/controls/color-picker.js
kopepasah: It could, but would require some adjusting non-the-less. Do you see any advantages of using the component over the current implementation?

lukecarbis: Relying on core components means less maintainability.

package.json
lukecarbis: It would be really nice if we could find a workaround for this without introducing a third-party dependency (our only one!).

kopepasah: We could use `tinycolor`, which is used in WordPress, but I did not see it exposed as usable. So it would have to be imported too.

Do you see an issue with introducing dependencies?

js/blocks/controls/color-picker.js
kopepasah: While I do agree, in this case we are only referencing a span and, even if we did use the component, we still need the wrapper component and style adjustments (which the core component does not allow to alter on the component level).

Issue #187, opened 2018-12-05
Add ESLint Config and Standards


kopepasah:
This issue will track the adding of ESLint and JavaScript coding standards checks.

kienstra:
Applied in #426

PR #202, opened 2018-12-11
Update PHPCS exclude patterns and add lint script.

kopepasah:
- [x] Add a Composer script to lint from within plugin directory, eliminating any conflicts with local PHPCS installs.
- [x] Update PHPCS config to exclude CSS and JS files.


kopepasah:
Noting that CSS and JS will be checked in #187.

lukecarbis (approved):

PR #204, opened 2018-12-12
Allow blank default for numbers, and fix incorrect values in editor

### LINKED ISSUE #203: Number field reverts to default value after refresh
When using a block with the Number Field Type, it works as expected upon entering a number, updating, previewing etc.

But when I refresh the page/come back to edit the post, the number field has reverted to the default value.

I tried leaving the default value empty, but then it's set to 0 automatically (Kinda like #196 with multi-selects)

**Version:** I'm using WordPress 5.0 and BlockLab 1.1.0.

lukecarbis:
Resolves #203.

### DIFF
--- a/js/blocks/controls/number.js
+++ b/js/blocks/controls/number.js
@@ -28,7 +28,7 @@ const BlockLabNumberControl = ( props, field, block ) => {
 			defaultValue={field.default}
 			value={attr[ field.name ]}
 			onChange={numberControl => {
-				attr[ field.name ] = numberControl
+				attr[ field.name ] = Number( numberControl )
 				setAttributes( attr )
 			}}
 			onFocus={ event => {
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -118,13 +118,14 @@ class Loader extends Component_Abstract {
 			}
 
 			register_block_type(
-				$block_name, [
+				$block_name,
+				array(
 					'attributes'      => $attributes,
 					// @see https://github.com/WordPress/gutenberg/issues/4671
 					'render_callback' => function ( $attributes ) use ( $block ) {
 						return $this->render_block_template( $block, $attributes );
 					},
-				]
+				)
 			);
 		}
 	}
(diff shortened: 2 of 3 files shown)


php/blocks/class-loader.php
```diff
@@ -118,13 +118,14 @@ public function dynamic_block_loader() {
 			}
 
 			register_block_type(
-				$block_name, [
+				$block_name,
+				array(
```
lukecarbis, line 122: This is just a coding standards fix I snuck in.

js/blocks/controls/number.js
```diff
@@ -28,7 +28,7 @@ const BlockLabNumberControl = ( props, field, block ) => {
 			defaultValue={field.default}
 			value={attr[ field.name ]}
 			onChange={numberControl => {
-				attr[ field.name ] = numberControl
+				attr[ field.name ] = Number( numberControl )
```
lukecarbis, line 31: Incorrect values were being saved because they were being passed to the validation function as a string.

php/blocks/controls/class-number.php
```diff
@@ -59,7 +59,9 @@ public function register_settings() {
 				'label'    => __( 'Default Value', 'block-lab' ),
 				'type'     => 'number',
 				'default'  => '',
-				'sanitize' => 'intval',
+				'sanitize' => function ( $value ) {
+					return filter_var( $value, FILTER_SANITIZE_NUMBER_INT );
```
lukecarbis, line 63: This allows a blank value to be saved as the default.

RobStino:
@lukecarbis 
Tested the saving issue, and all good there.
RE the "blank value to be saved as the default" fix, this does work now. However the editor preview never resolves (spinner) if the only field in the block is a number and has an empty/blank default. Guessing this is because it's looking for something to preview and there isn't anything.
Submit as a seperate issue?

lukecarbis:
Are you sure you ran `npm run build`?

RobStino (approved):
Tested and reviewed

PR #205, opened 2018-12-12
Update the field help setting

lukecarbis:
@RobStino For your UX consideration.

This PR changes the Field Instructions textarea that comes with every field so that it's a standard text input instead. It also changes the label from "Field instructions" to "Help Text".

Reason:
- 90% (or more) of "Field instructions" will be short, one sentence
- This encourages users to keep the instructions brief
- Significantly reduces the amount of space taken up by field settings
- The label "Help Text" is a better description (and it's capitalised like the other labels)


RobStino (commented):
Heaps better!

RobStino (approved):
Tested and reviewed.

PR #211, opened 2018-12-20
Bugfix for 1.1.1 release

lukecarbis:
This should resolve any outstanding issue in the 1.1.1 release. Specifically:

- No more "Undefined index" errors
- Correct values for checkboxes and toggles
- Correct preview template behaviour


RobStino (approved):
Tested and can confirm it fixes:
- Toggle/Checkbox incorrect values
- No more undefined errors
- Iss ue with preview template

PR #218, opened 2019-01-11
Version bump 1.1.2

lukecarbis:


RobStino (approved):
Did a sanity check.
Tested all fields in a new block in both the editor and the inspector. Took careful note if the new color field.
Look all good.

PR #220, opened 2019-01-21
Image upload control

### LINKED ISSUE #15: "Image" control type
- Return value
  - Image object
  - Image url
  - Image ID

lukecarbis:
Resolves #15


RobStino (approved):
Looks great!!!!
Good work!
Tested all other field types in editor and inspector as well and all good.

RobStino:
I noticed that the Rich Text controls disappear when typing. This felt weird, but then I checked the normal paragraph block and noticed that the behaviour is the same there.
See this vid for what I mean
https://d.pr/v/Ja6liq

Also, is it possible to get the text alignment options?
image

PR #221, opened 2019-01-21
Add rich text control

### LINKED ISSUE #20: "WYSIWYG" control type
Shows a rich editor, including formatting controls.

lukecarbis:
Resolves #20.


RobStino:
@lukecarbis See above for conflicts notice.

RobStino:
@lukecarbis Without this option checked, I still get new paragraphs when I create a new line.
image
image

RobStino:
I noticed that the Rich Text controls disappear when typing. This felt weird, but then I checked the normal paragraph block and noticed that the behaviour is the same there.
See this vid for what I mean
https://d.pr/v/Ja6liq

Also, is it possible to get the text alignment options?
image

RobStino:
Hey @lukecarbis 
Testing this further this morning and saw this bug.
Basically, I can't seem to go back to an empty state for the rich text field.
https://d.pr/v/0k0AuY

kienstra:
**Looks Good**
*Alignment Buttons*

Hi @lukecarbis,
This PR looks good. If it would help, I'm working on a commit that adds alignment buttons:

alignment-buttons

davidallenlewis:
This is awesome. Looking forward to it being merged.

I would like to see a format menu for headers though and a list option. You can use separate blocks for headings and lists but if your rich text is part of a larger "format" style block (say a 2 column image + text block with a background colour / image) it would be nice to have lists and headers. No more than that though. Keep it simple.

dave-smyth:
This might be a nightmare to implement, but it would be amazing if there was some way to control the options that are available in the editor. For instance, in some cases it would be useful to limit them to multiple lines of text only (no formatting other than bold and italic, perhaps), other times headings might be useful and in some cases alignment might be good, too.

Perhaps the editor could have three settings (minimal, standard, full) that would enable different levels of control.

Minimal: multiple lines, bold, italic, links
Standard: minimal + headings, lists
Full: everything else

AndrewSheetMetal:
Hey guys, just a quick question:

We need this feature for a project. How is the progress? When can we expect this feature to be distributed in a release?
Thanks!

lukecarbis:
@AndrewSheetMetal We're hoping to release it in our 1.3 milestone, which we're working on right now. :)

kienstra:
**Question About Alignment Controls**

Hi @lukecarbis and @RobStino,
Hope you had a great weekend.

Would you like the Rich Text control to have alignment controls?

Below is the current Rich Text control, with the possible location of alignment controls.

I liked the [location here](https://github.com/getblocklab/block-lab/pull/221#issuecomment-471322765) better, maybe the API changed in the meantime.

rich-text-current

I'm not sure if it's possible to get the `RichText` to display the alignment selected, though. For example, on clicking "Align text right," it'd be best if the `RichText` showed everything aligned to the right:

alignment-here

If you'd like alignment controls, I'll look at this more.

Thanks!

kienstra:
This PR already worked pretty well, even before I made some commits above. The [.gif above](https://github.com/getblocklab/block-lab/pull/221#issuecomment-496096523) shows the current state.

All of the controls shown above are included in the `RichText` by default, including Code, Inline Image, and Strikethrough. Except for the alignment controls shown above.

lukecarbis:
@kienstra I like the earlier version of the alignment controls better too (where they're not hidden behind a dropdown). But if that's no longer easily possible, I don't have that strong feelings about it.

Showing how the alignment effects the text, however, is very important. Without it, I'm sure I would think it was a bug.

js/blocks/controls/index.js
```diff
@@ -18,6 +19,7 @@ import BlockLabUserControl from './user'
 export default {
 	text: BlockLabTextControl,
 	textarea: BlockLabTextareaControl,
+	rich_text: BlockLabRichTextControl,
```
lukecarbis, line 22: Our first two word control!

php/blocks/controls/class-rich-text.php
```diff
@@ -0,0 +1,78 @@
...
+		$this->settings[] = new Control_Setting(
+			array(
+				'name'     => 'help',
+				'label'    => __( 'Help Text', 'block-lab' ),
+				'type'     => 'text',
+				'default'  => '',
+				'sanitize' => 'sanitize_text_field',
+			)
+		);
+		$this->settings[] = new Control_Setting(
```
lukecarbis: Hmm… what do you think… should we?

a) Remove this option entirely, and force new paragraphs. This is "rich text" and so paragraphs should be expected
b) Enable this by default. I would expect paragraphs in rich text
c) Use the same setting labels / options as we recently implemented for the textarea (doesn't work quite the same though since the textarea isn't an imported control with a `multiline` option)
d) Leave as it (but maybe better align the Help text and Label to the textarea setting?)

lukecarbis: @RobStino I'd love your thoughts on this, too. I'm leaning toward option A. What's a use case for not automatically creating paragraphs with a rich text field?

php/blocks/controls/class-rich-text.php
```diff
@@ -0,0 +1,78 @@
...
+		$this->label = __( 'Rich Text', 'block-lab' );
+	}
+
+	/**
+	 * Register settings.
+	 *
+	 * @return void
+	 */
+	public function register_settings() {
+		$this->settings[] = new Control_Setting(
```
lukecarbis: We should also find a way of disabling the "Field Location" setting, or forcing the Editor option, since it simply doesn't work when loaded in the Inspector.

Screen Shot 2019-05-28 at 8 26 46 am

php/post-types/class-block-post.php
```diff
@@ -128,7 +133,8 @@ public function get_control( $control_name ) {
 			return $this->controls[ $control_name ];
 		}
 
-		$control_class = 'Block_Lab\\Blocks\\Controls\\' . ucwords( $control_name );
+		$class_name    = ucwords( $control_name, '_' );
+		$control_class = 'Block_Lab\\Blocks\\Controls\\' . $class_name;
```
lukecarbis, line 137: Nice.

tests/php/blocks/controls/test-class-image.php
```diff
@@ -22,7 +22,7 @@ class Test_Image extends \WP_UnitTestCase {
 	/**
 	 * Instance of the setting.
 	 *
-	 * @var Controls\Control_setting
+	 * @var Controls\Control_Setting
```
lukecarbis, line 25: Great pickup!

kienstra:
Hi @lukecarbis,
Thanks for looking at this.

>@kienstra I like the earlier version of the alignment controls better too (where they're not hidden behind a dropdown). But if that's no longer easily possible, I don't have that strong feelings about it.

Yeah, that display was better. I'll look at if it's still possible.

>Showing how the alignment effects the text, however, is very important. Without it, I'm sure I would think it was a bug.

Good point. I'll also work on this.

lukecarbis (commented):
I can't approve / request changes since this is my PR… but so far it looks great. Haven't yet been able to test alignment controls. Is that committed yet? I couldn't find them.

Screen Shot 2019-05-28 at 8 16 17 am

A few minor suggestions, plus getting the alignment controls working is important. Alignment + Paragraphs is the challenging part, I believe!

php/blocks/controls/class-rich-text.php
kienstra: Ah, good catch.

tests/php/blocks/controls/test-class-image.php
kienstra: Haha, thanks!

js/blocks/controls/index.js
kienstra: Haha, a big occasion!

kienstra:
>Haven't yet been able to test alignment controls. Is that committed yet?

No, I just have it locally in a messy commit 😄It's not working in the sense of applying the alignment in the editor.

RobStino:
Hey all.
Testing this and it's looking good. Everything so far is behaving as expected.
Re the field setting for having `enter` start a new paragraph. I think this should be the default and actually be removed as a setting altogether. I can't think of a good use case otherwise. Also, `shift` + `enter` creates a break tag anyway so I think we're covered.
Looking forward to testing the alignment controls.

RobStino:
Seeing that the control just doesn't work in the Inspector, I think we don't make it an option for this particular field. For now at least.

kienstra:
>Re the field setting for having enter start a new paragraph. I think this should be the default and actually be removed as a setting altogether. I can't think of a good use case otherwise. Also, shift + enter creates a break tag anyway so I think we're covered.

Good point. Unless there are other ideas, I'll remove the 'Use Paragraphs' option for now.

kienstra:
**Alignment Controls Are In PR**
*Though They're Not Working Yet*

With https://github.com/getblocklab/block-lab/pull/221/commits/881d85706e4efd79bf69ee53ff6b5795241eeca7, there are now alignment controls, though I still have to get them to work 😄 I'll let you know how it goes.

alignment-in-pr

kienstra:
**WIP Alignment Controls**

The latest commits added alignment controls. But I still need to do some refactoring and bugfixes.

align-controls-here

kienstra:
**Current State**
*Alignment Controls Work*

rich-text-here

js/blocks/controls/rich-text.js
```diff
@@ -0,0 +1,120 @@
...
+const BlockLabRichTextControl = ( props, field, block ) => {
+	const { setAttributes } = props
+	const attr = { ...props.attributes }
+
+	return (
+		<BaseControl label={field.label} className="block-lab-rich-text-control " help={field.help}>
+			{
+			/*
+			* @todo: Resolve known issue with toolbar not disappearing on blur
+			* @see: https://github.com/WordPress/gutenberg/issues/7463
```
kienstra, line 13: It looks like there was some progress with this in https://github.com/WordPress/gutenberg/pull/14640, but there might not be a clear path to fixing this yet.

php/blocks/controls/class-checkbox.php
```diff
@@ -44,15 +44,8 @@ public function __construct() {
...
-		$this->settings[] = new Control_Setting(
-			array(
-				'name'     => 'help',
-				'label'    => __( 'Help Text', 'block-lab' ),
-				'type'     => 'text',
-				'default'  => '',
-				'sanitize' => 'sanitize_text_field',
-			)
-		);
+		$this->settings[] = new Control_Setting( $this->settings_config['location'] );
```
kienstra, line 47: @lukecarbis, your idea of adding the location setting only when needed was good. This does that.

php/blocks/controls/class-color.php
```diff
@@ -37,23 +37,8 @@ public function __construct() {
...
-				'name'     => 'default',
-				'label'    => __( 'Default Value', 'block-lab' ),
-				'type'     => 'text',
-				'default'  => '',
-				'sanitize' => 'sanitize_text_field',
-			)
-		);
+		foreach ( array( 'location', 'help', 'default' ) as $setting ) {
+			$this->settings[] = new Control_Setting( $this->settings_config[ $setting ] );
+		}
```
kienstra, line 42: This is intended to only refactor these controls, not change them. In brief testing of a few of them, there wasn't a backwards-compatibility problem.

But it'd be good if I added unit tests for the remaining controls before the next release. Maybe this will be in a different PR, for simplicity.

lukecarbis:
@kienstra I've got a bug that's blocking me from testing. When editing a block, if I choose any field type, no options load.

kienstra:
Hi @lukecarbis,
Sorry about that. https://github.com/getblocklab/block-lab/pull/221/commits/50361cadcc659329f437f707667ba6dea5e931a5 should fix the issue with the settings not loading.

Ah, just a minute, I didn't push the commit yet.

kienstra:
@lukecarbis, https://github.com/getblocklab/block-lab/pull/221/commits/5bf3c02c4088c0722785aab00e14fca781443bbd should fix the issue.

kienstra:
Hi @lukecarbis,
Thanks for reviewing this. 

As [shown above](https://github.com/getblocklab/block-lab/pull/221#issuecomment-496818207), there might still be some styling work needed here.

The `<ServerSideRender>` of the Rich Text control doesn't look exactly like the `<RichText>`. I'm not sure how close it's possible to get, but there could definitely be work remaining on that.

lukecarbis:
Cool – thanks! Got another bug now. The block edit state isn't showing anything. `npm run build` runs without any problems, but doesn't help.

kienstra:
OK, I see that also. I'm working on it.

kienstra:
Hi @lukecarbis, https://github.com/getblocklab/block-lab/pull/221/commits/ba38e356ae8075b27397a3491d6904706b8227a0 should allow you to continue testing. Of course, I'll have to revert that and come up with a better solution.

RobStino:
Just did some testing. Overall, great work! This is looking sharp!
Couple of things I noticed that may need some attention.

- The alignment action requires you to select the whole line. I believe normal editor behaviour doesn't need this. It just requires the cursor to be within the paragraph you are aligning.
- Pressing enter adds a `</br` tag rather than a new paragraph as we have done in the textarea
- Some styling improvements on the buttons.

Walkthrough if you need examples. Sorry about the _hushed_ voice. Kids just got to bed. :D
https://d.pr/v/qv2w8p

kienstra:
Hi @RobStino,

>Pressing enter adds a </br tag rather than a new paragraph as we have done in the textarea

Good catch. With https://github.com/getblocklab/block-lab/pull/221/commits/135afefe5f5f7fb033c86acc86b1c866ded55acf, this uses `wpautop()` on the front-end output to create `<p>` tags.

>The alignment action requires you to select the whole line. I believe normal editor behaviour doesn't need this. It just requires the cursor to be within the paragraph you are aligning.

>Some styling improvements on the buttons.

Good points. I'll work on them.

kienstra:
Hi @RobStino,

>Some styling improvements on the buttons.

Could you please elaborate a little on that? There definitely could be some improvements needed 😄

lukecarbis:
+1 for not selecting the whole line of text to realign it.

Buttons looks good to me. They match the paragraph block. Just one very minor glitch:
Screen Shot 2019-05-30 at 7 53 10 am

I'm seeing some screwy behaviour with new lines after alignment. Are we intending to break aligned text into a div? Should it be a paragraph?



Also, is it possible to style the text in the rich text area to better match a standard textarea? Something as simple as reducing the font size would makes a big difference.

Screen Shot 2019-05-30 at 8 00 02 am

js/blocks/controls/rich-text.js
```diff
@@ -0,0 +1,120 @@
...
+						value={ value }
+						onChange={ ( newAlignment ) => {
+							if ( newAlignment ) {
+								props.onChange(
+									applyFormat(
+										props.value,
+										{
+											type: ALIGNMENT_CONTROL_NAME,
+											attributes: {
+												align: getAlignmentStyle( newAlignment ),
```
lukecarbis: Heavy nesting… but I'm okay with it.

lukecarbis: You could potentially spin out the `onChange` handler into a separate method. But no real need.

js/blocks/controls/rich-text.js
```diff
@@ -0,0 +1,120 @@
...
+			* @see: https://github.com/WordPress/gutenberg/issues/7463
+			*/
+			}
+			<RichText
+				placeholder={field.placeholder || ''}
+				keepPlaceholderOnFocus={true}
+				defaultValue={field.default}
+				value={attr[ field.name ]}
+				className='input-control'
+				multiline={false}
```
lukecarbis: If we change this to true, what effect does that have? Would a keyboard "return" create a new paragraph (instead of a line break)?

php/blocks/controls/class-email.php
```diff
@@ -37,15 +37,8 @@ public function __construct() {
...
-		$this->settings[] = new Control_Setting(
-			array(
-				'name'     => 'help',
-				'label'    => __( 'Help Text', 'block-lab' ),
-				'type'     => 'text',
-				'default'  => '',
-				'sanitize' => 'sanitize_text_field',
-			)
-		);
+		$this->settings[] = new Control_Setting( $this->settings_config['location'] );
```
lukecarbis, line 40: Nice! ❤️

lukecarbis:
Just watched @RobStino's video, and I just want to agree with one thing he was trying to say: it should be possible to _toggle_ alignment, with the alignment controls.

So if I click "left align",   the text should explicitly align left. If I click "left align" again, the text should loose its specified alignment.

js/blocks/controls/rich-text.js
kienstra: Yeah, good point.

php/blocks/controls/class-email.php
kienstra: Thanks!

js/blocks/controls/rich-text.js
kienstra: Good point, https://github.com/getblocklab/block-lab/pull/221/commits/f7983f6ae0bf97d362a197153cc0fedd27e9791f changes it to `true`. That helped a lot, as the `RichText` now uses `<p>` for line breaks, just like the front-end.

js/blocks/controls/rich-text.js
kienstra: https://github.com/getblocklab/block-lab/pull/221/commits/4afb3f90c1d420195ed2980c77d20671ed812306#diff-b64eb5a514d6a8f70b18938222cd7a4fR45 applies your suggestion of creating a separate method for this.

kienstra:
Hi @lukecarbis,
>Just watched @RobStino's video, and I just want to agree with one thing he was trying to say: it should be possible to toggle alignment, with the alignment controls.

>So if I click "left align",   the text should explicitly align left. If I click "left align" again, the text should loose its specified alignment.

Sure, toggling alignment should work now. Also, it's no longer necessary to select text, simply placing a cursor on a line and clicking the 'center' button should center-align it.

kienstra:
Hi @lukecarbis,

>+1 for not selecting the whole line of text to realign it.

Sure, that's applied now. 

>Buttons looks good to me. They match the paragraph block. Just one very minor glitch:

How does the fix in https://github.com/getblocklab/block-lab/pull/221/commits/d3901de844de557d9fd7a28b768dd1cce6138798 look?

border-here

>I'm seeing some screwy behaviour with new lines after alignment. Are we intending to break aligned text into a div? Should it be a paragraph?

Good point, https://github.com/getblocklab/block-lab/pull/221/commits/f7983f6ae0bf97d362a197153cc0fedd27e9791f makes aligned text and new lines `<p>` tags.

>Also, is it possible to style the text in the rich text area to better match a standard textarea? Something as simple as reducing the font size would makes a big difference.

Yeah, `13px` would make this look more like the Text Area control. 

To me, it looks a little small, though:

13-px-font-size

lukecarbis:
@RobStino Can we get a third opinion on the text size? I prefer it smaller (matching a textarea).

lukecarbis:
Looks great! Some thoughts here: https://www.fromsmash.com/n-8KJbnS-f-c0

RobStino:
Hey friends
Epic work Ryan btw!
Couple of things:
- I agree with 13px. Keeping font size consistent between similar style fields is important I think. 
- the grey highlight you get is the same experience you get when formatting text in a regular text Block. I’m guessing this is where this has come from. 
- this carry over behaviour is probably what is causing the issue Luke spotted when he added a character to the same line as the centre aligned text.

kienstra:
The commits above partially address the testing comments. There's still more work needed, though.

kienstra:
Finishing the alignment controls (left, center, and right) could take at least a few more days. The behavior for alignment controls is a little different from the other controls, like bold. 

For example, as mentioned above, placing a cursor at the end of a centered line and typing line should result in more centered text, not a new line.

But the built-in behavior of the `RichText`, like for bold or italic, is that it doesn't continue the formatting. It wouldn't be bold on placing a cursor at the end of bold text and typing. So I've [been overriding](https://github.com/getblocklab/block-lab/pull/221/commits/c630cd125bed025cd76c49e8ae924113c3536341#diff-6a451811c6282091d849a83572f58a36R16) the `RichText` component to accommodate the alignment controls.

lukecarbis:
Just had a look through the code. Obviously having to extend the rich text is not ideal, but your implementation is really neat, there's obviously no other way to get alignment working the way we'd like it to.

Nice work!

kienstra:
>Obviously having to extend the rich text is not ideal, but your implementation is really neat, there's obviously no other way to get alignment working the way we'd like it to.

Thanks, @lukecarbis! Didn't want it to seem like I was complaining, just explaining it was taking a while 😄

kienstra:
**Extending RichText**

Extending the [RichText component](https://github.com/WordPress/gutenberg/blob/788a3d2ea7d5490f149bf5ebe191cbbcb140c6f1/packages/block-editor/src/components/rich-text/index.js#L105) isn't working locally so far.

Extending it looks to be necessary for [certain behaviors](https://github.com/getblocklab/block-lab/pull/221#issuecomment-498601148) of the alignment controls to work.

For example, I've been working on a class:

```js
const { RichText } = wp.editor;
const { LEFT, RIGHT } = wp.keycodes;
const { getComputedStyle, isCollapsed } = wp.richText;

/**
 * Browser dependencies
 */
const { getSelection } = window;

class BlockLabRichText extends RichText {


	/**
	 * Constructs the component class.
	 */
	constructor() {
		super( ...arguments );
		this.onSelectionChange = this.onSelectionChange.bind( this );
		this.onKeyDown = this.onKeyDown.bind( this );
		this.handleHorizontalNavigation = this.handleHorizontalNavigation.bind( this );
		this.onPointerDown = this.onPointerDown.bind( this );
	}

	/**
	 * Overrides the parent implementation of the `selectionchange` handler.
	 *
	 * Mainly copies the parent implemetation.
	 * The main difference is that this allows selecting the entire format by clicking on its border.
	 */
	onSelectionChange() {
		...
	}

	/* etc... */
}
```

But this class isn't using `onSelectionChange` above or the other extending methods. It's simply using the [onSelectionChange()](https://github.com/WordPress/gutenberg/blob/788a3d2ea7d5490f149bf5ebe191cbbcb140c6f1/packages/block-editor/src/components/rich-text/index.js#L458) method in the parent [RichText](https://github.com/WordPress/gutenberg/blob/788a3d2ea7d5490f149bf5ebe191cbbcb140c6f1/packages/block-editor/src/components/rich-text/index.js#L105) class when I debug this.

I'll keep looking at this, but it might not be possible to override these methods in the class to get the expected behavior for the alignment controls.

lukecarbis:
@kienstra It might be best to pull alignment controls entirely in that case, and put them in a separate issue.

kienstra:
**Alignment Controls Removed**

Hi @lukecarbis,
Thanks a lot for your ideas for the alignment controls.

I couldn't get them to work, including with the `onChangeAlign()` method.

This would probably need to extend [RichText](https://github.com/WordPress/gutenberg/blob/788a3d2ea7d5490f149bf5ebe191cbbcb140c6f1/packages/block-editor/src/components/rich-text/index.js#L105) somehow, but like we talked about, that's not recommended in React.

The behavior of the default formats like bold is different from the alignment formats in a few ways. I don't see a way to accommodate the alignment formats.

For example:
centered-this

lukecarbis:
Without alignment controls, this works brilliantly. I can't approve, since I started the request, but this has my green tick.

Just one topic of discussion. What do you think about reducing the gap between paragraphs from `0.1em` to something a little lower? `0.67em` looks a little nicer, in my opinion, and still provides a clear distinction between paragraph break and line break. Very minor difference:

 
Screen Shot 2019-06-12 at 9 33 29 am
Screen Shot 2019-06-12 at 9 33 06 am

kienstra:
Hi @lukecarbis,
Good idea. How does https://github.com/getblocklab/block-lab/pull/221/commits/a5856e53156952ffe8913363ba82d8ebc259fbac look? 

It uses you idea of reducing the margins:

paragraph-tags

Also, it removes the `FORMATTING_CONTROLS` constant, as it was simply the [default value](https://github.com/WordPress/gutenberg/blob/0cc2e02811b3fc64e5365927d16a105debc29895/packages/block-editor/src/components/rich-text/index.js#L1106) in `RichText`.

lukecarbis:
Great work Ryan.

✅ Review passed.

kienstra:
Thanks, Luke! Thanks a lot for your work on this.

RobStino (approved):
Hi everyone.
Sorry for the big delays on reviewing this PR.
Tested in my local and everything looks really good!

*Adding a Rich Text Field to a new Block*
image

*Adding the block in the Editor*
image

*Previewing in the Editor*
image

*On the front end*
image

kienstra:
**Thanks!**

Hi @RobStino, 
All good, thanks a lot for testing this!

alancwoo:
Is this block type only available in early access/pro? I just installed `1.3.4` and don't see this as an option.

kienstra:
Hi @alancwoo, yes, the Rich Text field is only available in the Pro version.

PR #224, opened 2019-01-25
Version bump 1.1.3

lukecarbis:


RobStino (approved):
Created a new block using all fields in both locations.
All looks good in the editor and the frontend
Then added default values to all fields
Retested and everything behaved as expected

PR #233, opened 2019-02-21
Add a "User" control type

### LINKED ISSUE #84: "User" control type
A searchable list of users.

kienstra:
* For now, this only has helper text and a `<select>` to determine how to display the name.

Closes #84


lukecarbis:
Remind me to update [the docs](
https://github.com/getblocklab/block-lab/wiki) to include this field, once we merge.

kienstra:
**User Input Works, But Still WIP**

The `<input>` to select the user now works. But this PR is still WIP.

This will need a way to output the field, and possibly more controls in the block.

testing-user-block

lukecarbis:
Looks great – will be able to re-use a lot of this for the Posts control as well.

kienstra:
**Front-end Output**

What should this control output on the front-end?

**Strategy A**
This could output the user slug (Username), and let the front-end template get whatever user data it needs. 

username-users

This would be flexible. For example, a template could output:

1. The username directly (though this might [have a security risk](https://clarkwp.wordpress.com/2014/02/07/how-to-change-your-wordpress-display-name/), as someone could attempt to hack the site with this):
```php
block_field( 'my-example-username' );
```

2. The [display_name](https://clarkwp.wordpress.com/2014/02/07/how-to-change-your-wordpress-display-name/) 

```php
$user = get_user_by( 'slug', block_value( 'my-example-username' ) );
if ( $user ) {
        echo get_avatar( $user->get( 'display_name' );
}
```

3. The avatar
```php
$user = get_user_by( 'slug', block_value( 'my-example-username' ) );
if ( $user ) {
        echo get_avatar( $user->ID );
}
```

**Strategy B**
Otherwise, if the idea of the user control is to simply output text, maybe echoing the 'displayname' would be enough:

```php
block_field( 'my-example-username' ); // Outputs displayname, like 'Alex Doe'
```

display-name-publicly

lukecarbis:
@kienstra What we do with the `multi-select` control type is relevant here, since it's value is an array.

If want to _output_ the value of a multi-select by using the `block_field()` function, you get a comma separated string. e.g.
```
block_field( 'interests' );
```
will output:
> Camping, Hiking, Fishing

But if you use the _return_ function `block_value()`, you just get the array. e.g.
```
$interests = block_value( 'interests' );
```
will cause `$interests` to be equal to an `Array` containing the values.

---

So, my suggestion with this field is:
- The _output_ function `block_field()` should output the Display Name.
- The _return_ function `block_value()` should return the `WP_User` object

Why the `WP_User` object? If we're just returning the User ID or username, the developer will be passing that on to a `get_user` function 99% of the time. For the other 1%, the `WP_User` object contains what they need anyway.

kienstra:
**Testing Steps**

1. In `/wp-admin` > Block Lab, create a new block
2. Add a new field
3. As the "Field Type," select "User":
select-user

4. Create a new post
5. Add the block and enter a username:
example-adding-user-block
6. Add this to a template in your active theme, like `blocks/block-test-user.php`:
```php
<?php
block_field( 'test-user' );

$user = block_value( 'test-user' );
if ( $user ) {
	echo get_avatar( $user->ID );
}
```
7. Expected: this should output the `display_name` and avatar on the front-end
avatar-front-end

kienstra:
**Request For Code Review**

Hi @lukecarbis,
Good [ideas above](https://github.com/getblocklab/block-lab/pull/233#issuecomment-468049316) on what values to make available on the front-end. This PR applies your ideas.

Could you please review this PR? There are [steps to test](https://github.com/getblocklab/block-lab/pull/233#issuecomment-468193797) above.

Please run `npm install` again, as this adds a dependency to it.

Thanks, Luke!

js/blocks/controls/user-input.js
```diff
@@ -0,0 +1,298 @@
...
+		this.state = {
+			users: [],
+			showSuggestions: false,
+			selectedSuggestion: null,
+		};
+	}
+
+	componentDidUpdate() {
+		const { showSuggestions, selectedSuggestion } = this.state;
+		// only have to worry about scrolling selected suggestion into view
```
lukecarbis: Hey! This PR is _amazing_. But I challenged myself to find _something_ to review you on. :P

Let's be consistent with our comments – always start with a Capital Letter, and finish with a full stop.

js/blocks/controls/user-input.js
```diff
@@ -0,0 +1,298 @@
...
+			if ( this.suggestionsRequest !== request ) {
+				return;
+			}
+
+			this.setState( {
+				users,
+				loading: false,
+			} );
+
+			if ( !! users.length ) {
```
lukecarbis: Cool. The `!!` operator is new to me.

php/blocks/controls/class-user.php
```diff
@@ -0,0 +1,48 @@
...
+	 *
+	 * @return void
+	 */
+	public function register_settings() {
+		$this->settings[] = new Control_Setting(
+			array(
+				'name'     => 'placeholder',
+				'label'    => __( 'Placeholder Text', 'block-lab' ),
+				'type'     => 'text',
+				'default'  => __( 'Enter a username', 'block-lab' ),
```
lukecarbis: This is the first time we are suggesting a default Placeholder. Just wanted to make sure we want to do that. @RobStino thoughts?

screen shot 2019-03-03 at 7 26 51 pm

js/blocks/controls/user-input.js
```diff
@@ -0,0 +1,298 @@
...
+		// Move focus to the input field when a link suggestion is clicked.
+		this.inputRef.current.focus();
+	}
+
+	render() {
+		const { value = '', autoFocus = true, instanceId, className, placeholder } = this.props;
+		const { showSuggestions, users, selectedSuggestion, loading } = this.state;
+		/* eslint-disable jsx-a11y/no-autofocus */
+		return (
+			<div className={ classnames( 'editor-url-input', className ) }>
```
lukecarbis: This seems to be missing the field's label? For other controls, I've used a `<BaseControl>` instead of a `<div>` here, like so:
```
<BaseControl label={field.label} className="{ classnames( 'editor-url-input', className ) }" help={field.help}>
    …
</BaseControl>

Comment (lukecarbis) [reply js/blocks/controls/user-input.js]: screen shot 2019-03-03 at 7 33 26 pm

File: php/blocks/controls/class-user.php [hunk 5]
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	/**
+	 * Register settings.
+	 *
+	 * @return void
+	 */
+	public function register_settings() {
+		$this->settings[] = new Control_Setting(
+			array(
```
lukecarbis, line 39: It would also be good to have a `help` option for help text.
```
		$this->settings[] = new Control_Setting(
			array(
				'name'     => 'help',
				'label'    => __( 'Help Text', 'block-lab' ),
				'type'     => 'text',
				'default'  => '',
				'sanitize' => 'sanitize_text_field',
			)
		);
```

js/blocks/controls/user-input.js
```diff
@@ -0,0 +1,298 @@
...
+			showSuggestions: false,
+			selectedSuggestion: null,
+		};
+	}
+
+	componentDidUpdate() {
+		const { showSuggestions, selectedSuggestion } = this.state;
+		// only have to worry about scrolling selected suggestion into view
+		// when already expanded
+		if ( showSuggestions && selectedSuggestion !== null && ! this.scrollingIntoView ) {
```
lukecarbis: This seems pretty edge case. What would be the effect of removing this code? If we just replaced it with a `focus` event on the input, would that basically achieve the same thing?

I'm a little allergic to external dependencies, sorry.

package.json
```diff
@@ -22,6 +22,7 @@
     "cross-env": "^5.1.5",
     "css-loader": "^0.28.11",
     "del": "^3.0.0",
+    "dom-scroll-into-view": "^1.2.1",
```
lukecarbis: Isn't this already packaged with WordPress?

lukecarbis (changes requested):
😻 Love this. Great work.

- [x] Can you please adjust so that it only loads if `block_lab()->is_pro()`?
- [ ] If the input loses focus, the popover should disappear
- [x] Adding the user control to the _inspector_ causes problems, including the field being too wide, required by default, and the popover not appearing
- [x] Correct the popover position (see screenshot below)
- [x] Correct the popover padding (see screenshot below)

screen shot 2019-03-03 at 7 36 54 pm

screen shot 2019-03-03 at 7 37 58 pm

package.json
kienstra: Good question. I couldn't find evidence that it's packaged with WordPress, though maybe it is.

For example, the `@wordpress/components` [package](https://github.com/WordPress/gutenberg/blob/3397ae408a870ded1516fcea77b63cdafb72f873/packages/components/package.json) includes from an external source, not WordPress:

```js
"dom-scroll-into-view": "^1.2.1",
```

js/blocks/controls/user-input.js
kienstra: Haha, thanks! 

Good point, 2e3f04b9059c1b4096135ff04606bc4bb537645a corrects this.

php/blocks/controls/class-user.php
kienstra: Now that there's help text in https://github.com/getblocklab/block-lab/pull/233/commits/d80465db25a7c78261801ab85f92e9675cf06b13, maybe removing the default for the placeholder would be good.

lukecarbis: Great. Let's keep it consistent then. I don't mind adding default placeholder text (or even help text), but if we do we should do it for all controls for consistency.

js/blocks/controls/user-input.js
kienstra: Great point.

https://github.com/getblocklab/block-lab/pull/233/commits/d80465db25a7c78261801ab85f92e9675cf06b13 uses your suggestion of changing the `<div>` to a `<BaseControl>`. Now, the label displays:

label-now-exists

php/blocks/controls/class-user.php
kienstra: Good suggestion, https://github.com/getblocklab/block-lab/pull/233/commits/d80465db25a7c78261801ab85f92e9675cf06b13#diff-7a9a8c2db176ed99ec78aa840392c3e0R39 adds your code above.

kienstra:
The popover now appears below, not below and to the right:

popover-under-input

lukecarbis:
@kienstra Just looking through everything again. How hard would it be to augment the UserInput component (`user-input.js`) to be a more generic "searchable" input component? Maybe `fetchInput` or `apiInput` or `searchInput` or something like that?

If we could do that, and pass along the API query path and query args when implementing it, then it could be re-usable for things like a Post control, or Taxonomy control.

lukecarbis:
An optional suggestion – instead of relying on the `editor-url-input`, `editor-url-input__suggestions`, and related classes, what if we used our own styles instead? Might turn out better in the long run, especially if the core editor styles change. I'm not sure though, so consider this more asking for your opinion than a requested change.

kienstra:
Hi @lukecarbis,
>If we could do that, and pass along the API query path and query args when implementing it, then it could be re-usable for things like a Post control, or Taxonomy control.

Great idea, [these commits](https://github.com/getblocklab/block-lab/compare/f58b8a3847c1dd6b86c007093e945d4d35a3d30c...fc89a0e1df7522fdc549795cfdd56888b314459e) abstract the component into `<FetchInput>`. This can be reused for taxonomies and posts.

For example, by simply changing the `getValueFromAPI` and `apiSlug`, this searches for posts:

```es6
const getValueFromAPI = apiResponse => apiResponse.title ? apiResponse.title.rendered : ''; 

return (
	<FetchInput
		field={field}
		placeholder={field.placeholder}
		value={attr[ field.name ]}
		apiSlug="posts" /* Changed */
		getValueFromAPI={getValueFromAPI} /* Changed */
		onChange={value => {
			attr[ field.name ] = getValueFromAPI( value )
			setAttributes( attr )
		}}
	/>
);
```

With those changes, this component can display posts:
displays-posts

php/blocks/controls/class-user.php
kienstra: Sounds good, 7bbb0ed97d923dc921218821df0a50eb1cbf0e53 removes the placeholder's default value.

kienstra:
Hi @lukecarbis, if it's alright, I'll continue working on this tomorrow (your Tuesday).

RobStino:
For these type of fields (user, post, taxonomy?), field parameters start to make more sense. E.g. I imagine with posts you may want to restrict the options by things like:
- Status (published, draft, etc)
- Author
- Post Type

lukecarbis:
I don't think a Post control is in scope for this PR (this is just focused on the User one), but I can see a "User Role" option working. Not for this PR, but maybe it could be a feature request.

js/blocks/controls/user-input.js
kienstra: Good question, I'm not sure what it does. It's from `URLInput` in Gutenberg. 

At least so far, it looks like deleting `componentDidUpdate()` entirely is fine. Here's what it looks like when all of the logic in `componentDidUpdate()` is deleted:

comp-did-update

https://github.com/getblocklab/block-lab/pull/233/commits/c464479dd004e23cbb19eda7b9c0504da4b9936d removes `componentDidUpdate()`, and https://github.com/getblocklab/block-lab/pull/233/commits/c6954ecf4b482f151e95a26faed945769f014f32 removes the dependency from `package.json`. If we later find a bug, we can address it then.

kienstra:
**License Expiration**

This should probably handle pro license expiration better. Right now, there's a PHP notice if a 'pro' control like 'user' isn't available:

control-not-available

kienstra:
Hi @lukecarbis,
All of your points here are applied except for 2 in [this comment](https://github.com/getblocklab/block-lab/pull/233#pullrequestreview-209901121).

The styling now [doesn't depend](https://github.com/getblocklab/block-lab/pull/233/commits/37f9e06b5f175d015f6165a59881de1bf05de637) on Gutenberg styling. That should help when we use this component for other controls, like posts or taxonomies.

 I'll continue with this tomorrow (your Wednesday) if that's alright.

lukecarbis:
> **License Expiration**
> 
> This should probably handle pro license expiration better. Right now, there's a PHP notice if a 'pro' control like 'user' isn't available:

Let's fix this in a separate PR. I'll create an issue.

kienstra:
>Adding the user control to the inspector causes problems, including the field being too wide, required by default, and the popover not appearing

Good points. Here's how the user control looks in the inspector now:
input-not-required

kienstra:
**Status**

Hi @lukecarbis,
Sorry this isn't finished yet. The only point remaining from your initial review is:
>If the input loses focus, the popover should disappear

I'll keep working on this tomorrow if that's alright. It's pretty easy to hide the `<Popover>` when the `<input>` loses focus. 

But the `<input>` also loses focus when selecting an item from the Popover. So I might throttle that somehow, so that the handler for selecting an item isn't overridden.

lukecarbis:
Ah yes, I remember running into that with the Colour popover. Don't feel rushed, though – this work is brilliant.

kienstra:
**onBlur Handler Added**

Hi @lukecarbis,
Thanks for waiting 😄 

https://github.com/getblocklab/block-lab/pull/233/commits/3147da88b37846b98da48c15d47e2d959f32c962 applies the last outstanding suggestion here. It hides the `<Popover>` on clicking outside of the `<input>`.

It's mainly taken from your [onBlur handler](https://github.com/getblocklab/block-lab/blob/b57612b5de7f98b0863bf701edf68c4e0932be51/js/blocks/controls/color.js#L32) in the color control:

losing-focus

Thanks, Luke!

js/blocks/controls/fetch-input.js
```diff
@@ -0,0 +1,304 @@
...
+
+		this.suggestionsRequest = request;
+	}
+
+	/**
+	 * On clicking the block outside the <input>, hide the Popover.
+	 * Mainly taken from the color control onBlur handler.
+	 */
+	onBlur( event ) {
+		console.log( event.relatedTarget.classList );
```
lukecarbis: Is this debug code?

js/blocks/controls/fetch-input.js
```diff
@@ -0,0 +1,304 @@
...
+		if (
+			event.relatedTarget
+			&&
+			(
+				event.relatedTarget.classList.contains( 'wp-block' ) // The block editor.
+				||
+				event.relatedTarget.classList.contains( 'edit-post-sidebar' ) // The inspector.
+			)
+		) {
```
lukecarbis, line 156: I just realised that this doesn't work if the new target is another input inside the same block. For example, consider a block with a user field and a text field. Type a few search characters in the user field, see the popover, then select the text field. The expected behaviour is that the popover would disappear, but it doesn't.

I remember this being a really tough problem to solve. If you feel that you've spent too long on it, feel free to just leave this for now and create a new issue to address it. It would be good if there were a way, though.

php/post-types/class-block-post.php
```diff
@@ -61,38 +62,40 @@ public function register_hooks() {
...
+		if ( block_lab()->is_pro() ) {
+			$controls = array_merge(
+				$controls,
+				array(
+					'user' => new Controls\User(),
+				)
+			);
+		}
+
 		$this->controls = apply_filters(
```
lukecarbis: Does this fit on one line now?

php/helpers.php
```diff
@@ -68,17 +68,38 @@ function block_field( $name, $echo = true ) {
 			$value = __( 'No', 'block-lab' );
 		}
 
+		/*
+		 * The user value is stored as the slug (user_login), which cannot change.
+		 * But it's echoed as the display_name, which can change via a setting in /wp-admin/user-edit.php.
+		 */
+		if ( has_user_control_type( $block_lab_config['fields'][ $name ] ) ) {
```
lukecarbis: Instead of creating a new function `has_user_control_type()`, we could do this:

Just below where `$value` is defined on line 33—36, we could define a new variable called `$field`:
```
if ( array_key_exists( $name, $block_lab_config['fields'] ) ) {
    $field = new Field( $block_lab_config['fields'][ $name ] );
} else {
    $field = new Field();
}
```

Once we have that, we don't need that additional helper function anymore, because to check if it's a `user` control type, we can just do:

```
if ( 'user' === $field->control ) {
    …
}
```

I don't know for sure that `$block_lab_config['fields'][ $name ]` is properly mapped to the `$args` a `Field` constructor is expecting, but if it is that would be neat.

lukecarbis (changes requested):
@kienstra Once again, a brilliant set of changes. Can't properly express how impressed I am with this PR, although I do have a couple of changes to suggest.

- [ ] Right now it's possible to input and save an invalid username. Is there a way of preventing invalid users? Or of checking if the user exists in the `block_field()` helper function? Or should we leave this validation step to the user? I tend to prefer keeping the user's job as simple as possible.

- [x] Type a username, then press the down key. The username text turns white, and can't be read.
Screen Shot 2019-03-11 at 8 06 12 am

- [x] The popover isn't aligned with the input field (try resizing your browser window to see the effect)
Screen Shot 2019-03-11 at 7 59 19 am

Getting closer!

php/helpers.php
kienstra: Good idea, https://github.com/getblocklab/block-lab/pull/233/commits/c9c0476380c5433b420c5fb831135cb39b7e3955 uses your code above.

php/post-types/class-block-post.php
kienstra: Good point, https://github.com/getblocklab/block-lab/pull/233/commits/f53ba5cf764dadeb84554803502ebc2d94e08a58 moves it to one line.

js/blocks/controls/fetch-input.js
kienstra: Ah, I should have removed this. https://github.com/getblocklab/block-lab/pull/233/commits/f53ba5cf764dadeb84554803502ebc2d94e08a58 deletes it.

kienstra:
Hi @lukecarbis,
Thanks for reviewing this again. Really good points, including:

>Is there a way of preventing invalid users?

If it's alright, I'll continue applying your suggestions tomorrow.

js/blocks/controls/fetch-input.js
kienstra: Hi @lukecarbis, good catch. https://github.com/getblocklab/block-lab/pull/233/commits/1b6c13b68fd38f2e4de66423c5a666499c56af07 looks to fix this locally:

text-field-clicking

kienstra:
**Question About Invalid User**

Hi @lukecarbis,
All of your points above are applied, except for this:

>Right now it's possible to input and save an invalid username. Is there a way of preventing invalid users? Or of checking if the user exists in the block_field() helper function? Or should we leave this validation step to the user? I tend to prefer keeping the user's job as simple as possible.

 What do you think of the rough `.gif` below? It needs styling, of course:

not-a-valid-user

The 'Not a valid user' Popover needs to be styled, and moved closer to the `<input>`. But what do you think of the concept?

At the moment, it still allows saving an invalid user. But that Popover will always display.

lukecarbis:
@kienstra That looks great. Had you considered using `setCustomValidity()` ([MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/setCustomValidity))?

I _think_ that should turn on the field's `:invalid` CSS pseudo-selector, thus enabling the "invalid" styling we already have. Try out the "email" field as see how it works with an invalid email.

kienstra:
Thanks, @lukecarbis. Great idea, `setCustomValidity()` looks to [have good browser support](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/setCustomValidity#Browser_compatibility).

kienstra:
**Validation Of User Value**

Hi @lukecarbis,
https://github.com/getblocklab/block-lab/pull/233/commits/cfa9100819e64898a45728732d0f606d2f46b9e7 uses your suggestion of the `setCustomValidity()` function.

It mainly works. But if you type part of a valid username, like 'ale' instead of the full 'alejandro', then click out of the `<input>`, there's no error. But ideally there would be.

beginning-input

`setCustomValidity()` doesn't look to block updating the post when there's an invalid value in the `<input>`. With a traditional form, if an `<input>` has an invalid state, it will block submitting the form. Maybe there's a way to do this, though.

lukecarbis:
Do you think it's worth adding the extra step of validating the username at the point it's being output? Maybe in the helper function?

What should the user expect if:
- Using `block_field()` and the user is valid? The display name?
- Using `block_field()` and the user is invalid? An empty string?
- Using `block_value()` and the user is valid? A `WP_User`?
- Using `block_value()` and the user is invalid? A `WP_Error`? `false`? `null`?

kienstra:
@lukecarbis, good points. I'll look at what this would output with `block_field()` and `block_value()`.

Also, it'd be good if there were a better way to validate the user in the editor. Maybe a notice on updating a post when the user doesn't exist. At the very least, maybe that could be a backlog item.

kienstra:
**Latest Commits Filter The Output**

Hi @lukecarbis,
What do you think about the last 2 commits? They create a [filter](https://github.com/getblocklab/block-lab/pull/233/files#diff-db54e597c9cdb4b241d129437fc64bc6R73) of the output in `block_field()`.

Then, the filter's callback delegates the output to a possible [output method](https://github.com/getblocklab/block-lab/pull/233/files#diff-7a9a8c2db176ed99ec78aa840392c3e0R65) in the control class. Right now, only the `User` class has that method.

Adding a filter might not be the best approach. By releasing this, we would of course be committing to keeping that filter 😄

kienstra:
Also, the commits above apply your [suggested outputs](https://github.com/getblocklab/block-lab/pull/233#issuecomment-472306385) for the 4 cases.

kienstra:
**Changed Filter And Method Names**

Hi @lukecarbis,
Like we talked about, https://github.com/getblocklab/block-lab/pull/233/commits/24764ec8466aa0b75c719d3e588bbff145589c03 makes the method and filter name changes we talked about:

`block_lab_output_value` filter changed to `block_lab_field_value` 
`get_output_value()` changed to `get_field_value()`
`output()` changed to `validate()`

kienstra:
Also, it would be great if you could test this PR on a few different browsers. I should have done that 😄

js/blocks/controls/fetch-input.js
```diff
@@ -63,28 +64,14 @@ class FetchInput extends Component {
...
-		if ( value.length < 2 ) {
-			this.setState( {
-				showSuggestions: false,
-				selectedSuggestion: null,
-				loading: false,
-			} );
-
-			return;
-		}
-
```
lukecarbis: Removing this allows the dropdown to show immediately.

js/blocks/controls/fetch-input.js
```diff
@@ -63,28 +64,14 @@ class FetchInput extends Component {
...
 		this.setState( {
-			showSuggestions: true,
-			selectedSuggestion: null,
```
lukecarbis: This corrects a strange behaviour where suggestions loaded twice after selecting one and then refocusing the field.

js/blocks/controls/fetch-input.js
```diff
@@ -108,6 +96,13 @@ class FetchInput extends Component {
 					results.length,
 					'block-lab'
 				), results.length ), 'assertive' );
+
+
+				if ( null === this.state.selectedSuggestion && '' !== this.props.value ) {
+					this.setState( {
+						selectedSuggestion: 0
```
lukecarbis, line 102: If there's no selectedSuggestion, make the first one selected. This allows the user to just type the first few letters of a name and press Enter (no down arrow).

js/blocks/controls/fetch-input.js
```diff
@@ -141,8 +136,11 @@ class FetchInput extends Component {
 		} else {
 			this.inputRef.current.setCustomValidity( '' );
 		}
-	}
 
+		this.inputRef.current.className = classNames( 'bl-fetch__input', {
+			'text-control__error': ! isValid
```
lukecarbis, line 140: Also add a red border if the input is invalid (which is what we do with other fields).

js/blocks/controls/fetch-input.js
```diff
@@ -152,13 +150,34 @@ class FetchInput extends Component {
 	 * That has its own handler, which will eventually hide the Popover.
 	 */
 	onBlur( event ) {
-		if ( event.relatedTarget && ! event.relatedTarget.classList.contains( 'bl-fetch-input__suggestion' ) ) {
+		if (
+			event.relatedTarget &&
+			! event.relatedTarget.classList.contains( 'components-popover__content' ) &&
+			! event.relatedTarget.classList.contains( 'bl-fetch-input__suggestion' )
```
lukecarbis, line 155: Different browsers pick up different parent elements as the `event.relatedTarget`, so including both classes in the check here is necessary.

js/blocks/controls/fetch-input.js
```diff
@@ -152,13 +150,34 @@ class FetchInput extends Component {
...
+				return;
+			}
+
+			const matchingResults = this.state.results.filter(
+				suggestions => ( suggestions.slug === this.props.value )
+			);
+
+			if ( ! matchingResults.length ) {
+				this.selectLink( this.state.results[ this.state.selectedSuggestion ] );
+			}
```
lukecarbis, line 171: If the user blurs the field with only a partial name match, then populate the field with the full suggestion (otherwise we get an invalid user).

js/blocks/controls/fetch-input.js
```diff
@@ -152,13 +150,34 @@ class FetchInput extends Component {
...
 		}
 	}
 
+	onFocus() {
+		const inputValue = this.props.value;
+		this.updateSuggestions( inputValue );
```
lukecarbis, line 177: Show the dropdown immediately when focusing on the field.

js/blocks/controls/fetch-input.js
```diff
@@ -238,34 +257,33 @@ class FetchInput extends Component {
...
 
 	handleOnClick( result ) {
 		this.selectLink( result );
-		// Move focus to the input field when a suggestion is clicked.
-		this.inputRef.current.focus();
 	}
 
 	render() {
-		const { value = '', autoFocus = true, instanceId, className, placeholder, field, getValueFromAPI } = this.props;
+		const { value = '', autoFocus = false, instanceId, className, placeholder, field, getValueFromAPI } = this.props;
```
lukecarbis, line 279: Since we're showing the dropdown immediately, we should autofocus the field.

js/blocks/controls/fetch-input.js
```diff
@@ -320,7 +343,7 @@ class FetchInput extends Component {
 						</div>
 					</Popover>
 				}
-				{ ! showSuggestions || ! results.length && this.setInputValidity( false ) }
+				{ ! showSuggestions || '' === this.props.value || ! results.length && this.setInputValidity( false ) }
```
lukecarbis, line 345: An empty input should be valid.

lukecarbis:
@kienstra Could you please code review c20166c?

@RobStino It would be great to get another UX review, too. :) I haven't tested output functions.

I've checked this in Firefox, Safari, and Chrome.

RobStino:
Those UX fixes are great.
I did observe this. I _think_ it's ok though, just wanted to point it out.
https://d.pr/v/gP8T8t

lukecarbis:
@RobStino I implemented that intentionally. Can easily remove it, but I think it's a better experience than getting the error.

lukecarbis:
@RobStino to test the alternative, change line `170` in `js/blocks/controls/fetch-input.js` to:

```js
this.setInputValidity( false );
```

RobStino:
Tested, but was unable to even close the block as the autoselect kept reactivating the field.
Either way though, I'm happy to leave it as it is.

lukecarbis:
Exactly why I changed it. That refocus effect is created entirely by the browser and there's nothing we could do to prevent it.

lukecarbis (dismissed):

js/blocks/controls/fetch-input.js
```diff
@@ -98,6 +85,7 @@ class FetchInput extends Component {
 
 			this.setState( {
 				results,
+				showSuggestions: true,
```
kienstra, line 88: Great idea to always show suggestions of users.

js/blocks/controls/fetch-input.js
kienstra: Great idea!

top-selection

js/blocks/controls/fetch-input.js
kienstra: Nice, that looks good.

js/blocks/controls/fetch-input.js
kienstra: Nice improvement here. That definitely makes more sense than leaving the `<input>` empty.

js/blocks/loader/editor.scss
```diff
@@ -242,6 +242,24 @@ div[class^="wp-block-block-lab-"],
   }
 }
 
+.bl-fetch-input {
+  >.components-base-control__field {
```
kienstra: This is a nitpick, but this could probably use a space between the `>` and `.`:

```css
> .components-base-control__field {
```

js/blocks/loader/editor.scss
```diff
@@ -242,6 +242,24 @@ div[class^="wp-block-block-lab-"],
...
+    display: flex;
```
kienstra, line 247: It's nice how the plugin has an `.scss` build system, even though there aren't many styles. It keeps this more DRY.

js/blocks/controls/fetch-input.js
kienstra: Nice, that's a great improvement. People might not know the exact username they're looking for.

kienstra (commented):
**Approved**

Hi @lukecarbis,
Thanks, you made some great improvements to this. It's much more intuitive, now that it displays users on clicking the `<input>`.

Because I opened the PR, I can't officially "approve" this. But you get the idea 😄

lukecarbis (approved):

PR #234, opened 2019-02-21
🅿️®️🅾️❗️

### LINKED ISSUE #18: Import / Export JSON file of blocks
This is especially useful for theme authors who wish to use ACB to build blocks, and want to distribute the blocks with their themes.

lukecarbis:
This PR introduces:

- [x] A Settings screen with offsite links to Docs and Help
- [x] Licensing for Pro registration
- [x] Export individual blocks
- [x] Bulk export blocks
- [x] Import blocks (via Tools > Import) (note: we can improve visibility of this feature in a future release)
- [x] A "Go Pro" feature (visible when there is no active license) (To be completed by @RobStino)

@kienstra I have a question about how the plugin is structured in this PR. I've stored additional classes (`class-admin.php`, `class-import.php`, `class-settings.php`, `class-settings.php` and `class-license.php`) in the `php` directory, but I'd like your opinion on whether there should be a (namespaced?) child directory within `php` where these files are kept.

Similarly, are you happy with `php/views/` for front-endy template files?

Closes #18.


php/class-settings.php
```diff
@@ -0,0 +1,148 @@
...
+
+	/**
+	 * Render the Settings page header.
+	 */
+	public function render_page_header() {
+		?>
+		<h2><?php echo esc_html( get_admin_page_title() ); ?></h2>
+		<h2 class="nav-tab-wrapper">
+			<a href="<?php echo esc_url( add_query_arg( 'tab', 'license' ) ); ?>" title="License" class="nav-tab nav-tab-active dashicons-before dashicons-nametag">
+				<?php esc_html_e( 'License' ); ?>
```
kienstra: Should this have the textdomain?
```php
<?php esc_html_e( 'License', 'block-lab' ); ?>
```
And likewise for the `esc_html_e()` calls below.

php/class-upgrade.php
```diff
@@ -0,0 +1,82 @@
...
+	 *
+	 * @var string
+	 */
+	public $slug = 'block-lab-pro';
+
+	/**
+	 * Initialise the Upgrade component.
+	 */
+	public function init() {
+	}
```
kienstra: This PR is marked WIP, so maybe this method will later have logic.

php/post-types/class-block-post.php
```diff
@@ -1017,6 +1019,28 @@ public function post_row_actions( $actions = array() ) {
 			unset( $actions['inline hide-if-no-js'] );
 		}
 
+		// Add the Export link.
+		if ( block_lab()->is_pro() ) {
+			$export = array(
+				'export' => sprintf(
+					'<a href="%1$s" aria-label="%2$s">%3$s</a>',
```
kienstra, line 1026: It's nice how this uses `sprintf()` to construct the `<a ...>`.

php/views/license.php
```diff
@@ -0,0 +1,108 @@
...
+					if ( isset( $license['expires'] ) ) {
+						$expiry = date( get_option( 'date_format' ), strtotime( $license['expires'] ) );
+					}
+
+					echo wp_kses_post(
+						sprintf(
+							'<p>%1$s %2$s</p>',
+							sprintf(
+								// translators: A number, wrapped in <strong> tags.
+								__( 'Your license includes %1$s%2$s%3$s site installs.', 'block-lab' ),
```
kienstra: This is fine as is. But it might be simpler to include the `<strong>` in the `__()` call:
```php
__( 'Your license includes <strong>%s</strong> site installs.', 'block-lab' )
```

This [happens in Core](https://github.com/WordPress/wordpress-develop/blob/9448313544e5d6db16ac640d552258405e8bc3d8/src/wp-admin/revision.php#L128) sometimes.

Likewise for the similar `__()` calls below.

php/views/upgrade.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
kienstra, line 1: Nice organization of this in `views/`

php/class-upgrade.php
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Add submenu pages to the Block Lab menu.
+	 */
+	public function add_submenu_pages() {
+		add_submenu_page(
+			'edit.php?post_type=block_lab',
+			__( 'Block Lab Pro', 'block-lab' ),
+			__( 'Go Pro', 'block-lab' ),
```
kienstra: This isn't my domain 😄 

But do you think some people might confuse this with the [GoPro](https://gopro.com) camera?

go-pro

php/views/license.php
```diff
@@ -0,0 +1,108 @@
...
+					);
+				} else {
+					echo wp_kses_post(
+						sprintf(
+							'<p>%1$s %2$s</p>',
+							__( 'No Pro license was found for this installation.', 'block-lab' ),
+							sprintf(
+								// translators: Opening and closing anchor and emphasis tags.
+								__( '%1$s%2$sPurchase Block Lab Pro now!%3$s%4$s', 'block-lab' ),
+								'<a href="' . add_query_arg( array( 'page' => 'block-lab-pro' ) ) . '">',
```
kienstra: Wrapping `add_query_arg()` in `esc_url()` might be a little safer.

php/class-license.php
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	/**
+	 * Admin notice for incorrect license details.
+	 *
+	 * @return string
+	 */
+	public function license_error_message() {
+		$message = __( 'There was a problem activating your Block Lab license.', 'block-lab' );
+		return sprintf( '<div class="notice notice-error"><p>%s</p></div>', esc_html( $message ) );
```
kienstra: This works well when entering a wrong license key:
 
feedback-license

php/class-import.php
```diff
@@ -0,0 +1,247 @@
...
+			__( 'Block Lab', 'block-lab' ),
+			__( 'Import custom blocks created with Block Lab.', 'block-lab' ),
+			array( $this, 'render_page' )
+		);
+	}
+
+	/**
+	 * Render the import page. Manages the three separate stages of the JSON import process.
+	 */
+	public function render_page() {
```
kienstra: The import UI for the pro version looks good, though I didn't test it with a JSON file:


import-ui

css/admin.css
```diff
@@ -0,0 +1,3 @@
+#adminmenu ul > li > a[href="edit.php?post_type=block_lab&page=block-lab-pro"] {
```
kienstra, line 1: It's nice how you organized the stying in `css/`. Sometimes I'll output inline `<style>` tags in the markup, but it's cleaner how they're organized here in `css/`.

php/class-import.php
```diff
@@ -0,0 +1,247 @@
...
+	 *
+	 * @var string
+	 */
+	public $slug = 'block-lab';
+
+	/**
+	 * Initialise the Upgrade component.
+	 */
+	public function init() {
+	}
```
kienstra: Maybe this function will be populated later.

php/class-license.php
```diff
@@ -0,0 +1,156 @@
...
+				'body'      => $api_params,
+			)
+		);
+
+		if ( is_wp_error( $response ) ) {
+			return false;
+		}
+
+		$license    = json_decode( wp_remote_retrieve_body( $response ), true );
+		$expiration = 24 * 12 * 12; // 24 hours in seconds.
```
kienstra: How about `DAY_IN_SECONDS` here?

It looks like [it's defined in Core](https://codex.wordpress.org/Easier_Expression_of_Time_Constants) (I just found out about this).

php/class-settings.php
```diff
@@ -0,0 +1,148 @@
...
+	public function prepare_notice( $notice ) {
+		$notices   = get_option( 'block_lab_notices', array() );
+		$notices[] = $notice;
+		update_option( 'block_lab_notices', $notices );
+	}
+
+	/**
+	 * Show any admin notices after saving the settings.
+	 */
+	public function show_notices() {
```
kienstra: This is an interesting way of queuing the notices to display 😄

kienstra (commented):
**Looks Good**
*Some Minor Points*

Hi @lukecarbis,
This PR for the pro license looks really good. 

It's WIP and there are some small points, so I won't approve it yet if that's alright.

>@kienstra I have a question about how the plugin is structured in this PR. I've stored additional classes (`class-admin.php`, `class-import.php`, `class-settings.php`, `class-settings.php` and `class-license.php`) in the `php` directory, but I'd like your opinion on whether there should be a (namespaced?) child directory within `php` where these files are kept.

How about a `php/admin/` directory for those files? The AMP plugin has something similar in [includes/admin/](https://github.com/ampproject/amp-wp/tree/286aef5d61e838d4b4b5797acd583a5e0e947f90/includes/admin). 

Of course, many more files relate to `/wp-admin`, but these seem specific to admin menus.

Like you mentioned, it'd be good to namespace the files if they go in a directory like `php/admin/`, to follow the convention of the `php/blocks/` files having a namespace of `Block_Lab\Blocks`, etc...

So maybe the namespace could be something like `Block_Lab\Admin`.

>Similarly, are you happy with `php/views/` for front-endy template files?

Yes, good idea to put them in `php/views/`. If you go with something like a `php/admin/` directory, maybe they could be in `php/admin/views/`. But I think they're fine either way.

php/post-types/class-block-post.php
lukecarbis: Thanks!

php/views/license.php
lukecarbis: Great pickup.

php/class-settings.php
lukecarbis: Whoops! Thanks for noticing.

php/class-license.php
lukecarbis: Oh I totally forgot about that! Yes, definitely.

php/class-settings.php
lukecarbis: Yes, this happens this way for a reason. I am using WordPress' Settings API to save all the options, which means I don't have to write any code which checks for $_POST variables or any code to `update_option` s.

The side effect is that the form goes off to `options.php` to save, and then redirects back to the settings page. After the redirect, I've got no $_POST variables to speak of, so I don't know  if the form was saved, or if it was saved successfully.

Two ways I see around this:
- Hijack the options saving process and redirect back to the settings page with an additional $_GET parameter indicating the result (then using that result to display a notice)
- Save the notice in the database, and check if it exists when the page loads (and show if it does)

I opted for the second method, since I feel it's cleaner, and potentially reusable in other situations.

lukecarbis:
@kienstra Ready for you to re-review. No longer a [WIP]!

I've moved Admin, License, Settings, and Upgrade to `php/admin/` and given them the `Admin` namespace.

Only problem now is that we have a class called `Admin\Admin`, which is a bit repetitive. 

Can you think of a better name for the `Admin\Admin` class, or the namespace we're using? Or do you think this classname is okay?

php/class-settings.php
kienstra: Sounds good, thanks for explaining this 😄

kienstra:
>Can you think of a better name for the `Admin\Admin` class, or the namespace we're using? Or do you think this classname is okay?

Maybe `Admin\Setup`.

Even `Admin\Admin` might be an alright name, though. There's a [Blocks\Block](https://github.com/getblocklab/block-lab/blob/8c8d31d260ca143ea91970f2c207c741d69d1fe3/php/blocks/class-block.php#L15) class,

kienstra:
>I've moved Admin, License, Settings, and Upgrade to `php/admin/` and given them the `Admin` namespace.

Nice 😄

kienstra:
**Pro Page**

@lukecarbis,
It looks like the error message isn't displaying correctly on submitting a wrong license key:

error-message

lukecarbis:
@kienstra That should be fixed now.

kienstra:
**Approved**

Hi @lukecarbis,
This looks really good. Thanks for addressing the points here.

The notice looks good, too:
notice-appears-expected

kienstra (approved):

PR #241, opened 2019-03-03
Create class-postcategory.php

lewist01:
Add a drop-down list of post category.
I added this to allow posts to be displayed in a page by Category Name or all Catagories.

### DIFF
--- a/class-postcategory.php
+++ b/class-postcategory.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * postcategory control.
+ *
+ * @package   Block_Lab
+ * @copyright Copyright(c) 2018, Block Lab
+ * @license http://opensource.org/licenses/GPL-2.0 GNU General Public License, version 2 (GPL-2.0)
+ */
+
+namespace Block_Lab\Blocks\Controls;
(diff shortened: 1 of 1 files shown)


lukecarbis:
Thanks so much for the PR @lewist01! Really appreciate you taking the time to put it together.

Once potential problem I noticed is that it could cause performance problems on sites with many categories. The browser is going to really struggle to show a dropdown list containing 1,000 items, and I've seen sites with many more categories than that.

The good news is that this PR (#233) may seem unrelated, but it implements a reusable API search field. It's going to be really easy to extend that to Posts and Taxonomies like Categories or Tags.

So, I'll close this PR for now, but you can expect the feature you've written to be in either the next release, or the one after.

lewist01:
Hi

Thank you for looking and I am pleased that you found it useful.

I realised for a wider application there may  be issues with sites that have a large amount categories, but needed this to add posts to pages based on a category or multiple categories.
Will be looking into Taxonomies for custom posts as use these on a few of my sites.

 

Thank you again for taking the time to look.

 

Tony

PR #257, opened 2019-03-21
Add a 'Post' control

### LINKED ISSUE #19: "Post" control type
A searchable list of posts, from a given post type.

- Post type
- Post
- Taxonomy filter

kienstra:
* Right now, it's only available in the `/wp-admin` > Block Lab UI.
* The React control is not yet added, so it's not available in the block editor UI.

post-control-added

Closes #19


kienstra:
**Storing Title In Attributes, Not Fetching It**

The [latest commit](https://github.com/getblocklab/block-lab/pull/257#issuecomment-482444369) stores the title [in the block's attributes](https://github.com/getblocklab/block-lab/pull/257/commits/529801ceb81b67ce05d356ad5cec7a810ad9d242#diff-445f4466528bc0064cf52d8c4c5f25beR38). It's not an ideal solution.

But it looks like fetching the title every time the components mounts might have too much of a delay.

This Post control needs to display the post's title in the `<input>`, but save the post's ID as the control value. The title is mutable, and the ID is not.

Before, I didn't see a clear way to store the post's title. So I used `apiFetch()` to get it when the component mounted:

delayed-title

This is probably too long of a delay. Especially as this is in a local environment. It would probably be longer on a normal host.

This is still WIP, but it looks like storing the title [in the block's attributes](https://github.com/getblocklab/block-lab/pull/257/commits/529801ceb81b67ce05d356ad5cec7a810ad9d242#diff-445f4466528bc0064cf52d8c4c5f25beR38) gives a better user experience. It appears without the delay shown above.

kienstra:
This is almost ready for review. The main use case is working, I just need to test it and fix any possible bugs.

lukecarbis:
@kienstra Looking great! What do you think about storing an _array_ instead of just the post title? It could include the post title, post type, and post ID.

php/blocks/controls/class-post.php
```diff
@@ -0,0 +1,80 @@
...
+		);
+		$this->settings[] = new Control_Setting(
+			array(
+				'name'     => 'placeholder',
+				'label'    => __( 'Placeholder Text', 'block-lab' ),
+				'type'     => 'text',
+				'default'  => '',
+				'sanitize' => 'sanitize_text_field',
+			)
+		);
```
lukecarbis, line 62: It would be really cool to include a dropdown list of different post types here. It wouldn't have to be dynamic, as I think it's reasonable to assume the list of post types active on a given site isn't very long.

kienstra: Good idea, @lukecarbis. Would this be a dropdown of all registered post types, like from [get_post_types()](https://codex.wordpress.org/Function_Reference/get_post_types) ?

kienstra:
**Storing Array**

Hi @lukecarbis,
>@kienstra Looking great! What do you think about storing an array instead of just the post title? It could include the post title, post type, and post ID.

Great idea. I'll work on implementing that.

php/blocks/controls/class-post.php
lukecarbis: Exactly, yes.

kienstra: Sounds good, @lukecarbis. And it would be in the `/wp-admin` > Block Lab UI, not the block editor, right?

dropdown-post-types

lukecarbis: Yes – exactly. Then in the block editor the posts would be filtered by the post type you selected here.

kienstra:
**Maybe This Shouldn't Include Attachments**

This now allows selecting all public post types. That includes media (attachments). 

But maybe this shouldn't include media, as there's an Image control for that. 

post-control

php/blocks/controls/class-post.php
kienstra: Good idea. This is implemented now.

kienstra:
**Request For Review**

Hi @lukecarbis,
Could you please review this? It has the `<select>` element to query by post type. For example, if you select 'Pages', it will only query for pages:

page-post-type

php/blocks/controls/class-control-abstract.php
```diff
@@ -388,4 +418,27 @@ function( $option ) use ( &$options ) {
...
+	public function sanitize_post_type( $value ) {
+		$public_post_types = get_post_types( array( 'public' => true ) );
+		foreach ( $public_post_types as $post_type ) {
+			$post_type_object = get_post_type_object( $post_type );
+			if (
+				( $post_type_object && ( $value === $post_type_object->rest_base ) )
+				||
+				$post_type === $value
+			) {
+				return $value;
```
lukecarbis: This looks like it will sort-of fail to sanitise (or at least fail to return a value) if the `$value` passed in isn't a registered post type, _or_ if there are no registered post types. This could be refactored so that we return `null` or `false` or `''` or something if there's a bad result.
e.g.
```
	public function sanitize_post_type( $value ) {
		$value = '';

		$public_post_types = get_post_types( array( 'public' => true ) );
		foreach ( $public_post_types as $post_type ) {
			$post_type_object = get_post_type_object( $post_type );

			if ( $post_type_object && ( $value === $post_type_object->rest_base ) ) {
				$value = $post_type_object->rest_base;
				break;
			}

			if ( $value === $post_type ) {
				$value = $post_type;
				break;
			}
		}

		return $value;
	}
```

But maybe there's a better way?

lukecarbis (changes requested):
Very nicely done! Love the solutioning and architecture behind this.

I noticed that the post type isn't saved in the JSON, only the post title and ID. I can't think of a good reason to store the post type as well, since it's pretty accessible via the post ID, but we had talked about including it. Happy to leave it out if that's what you think is best.

```
<!-- wp:block-lab/post-block
{
    "text-only": "Test",
    "choose-a-page": {
        "id": 2,
        "title": "Sample Page"
    },
    "more-text": "More text",
    "inspector-post": {
        "id": 1,
        "title": "Hello world!"
    },
    "user": "lukecarbis"
}
/-->
```

If it's an easy fix, notice that the user input only saves the username, not the ID? I think it would be better if it matched the Post control. Would you prefer to fix that here or in a separate PR?

Another inconsistency with the User control is that if you type in _invalid_ username, and then blur the field, the field displays an error. The Post control clears the text (in addition to a red border to indicate that there was a problem). I'm not sure if this was intentional, but I prefer the latter (Post style) way of handling invalid text. Could we make this consistent?

Lastly, I agree that we should filter out the Attachment post type. This shouldn't be an option.

RobStino:
Great work fellas!
Testing now. Spotted a couple of things and possibly a bug:
1. If a page/post has no title it can't be selected, even if it's published. Edge-case issue.
2. The filtered select experience feels a little wrong when you want to re-select. Not a bug, bust something to think about in general
3. I was missing a page (original Sample page) in my select options. This might be a bug.

These 3 points are covered in this screencap.

https://d.pr/v/wTI3HT

kienstra:
Hi @lukecarbis,
>I noticed that the post type isn't saved in the JSON, only the post title and ID. I can't think of a good reason to store the post type as well, since it's pretty accessible via the post ID, but we had talked about including it. Happy to leave it out if that's what you think is best.

Yeah, it probably wouldn't hurt to store the `post_type` separately. But `block_value( 'foo-field' )` will return a `WP_Post` object, so they could access it with the `->post_type` property, like you mentioned.

php/blocks/controls/class-control-abstract.php
kienstra: Hi @lukecarbis,
Good point that it doesn't return anything here. I wasn't quite sure what was best.

But it looks like having the validation method return `null` (or not return anything) is necessary to fall back to [the default value](https://github.com/getblocklab/block-lab/blob/ccabbed990b5541e5ac3ff59b1352c6d8bbe3dad/php/blocks/controls/class-control-setting.php#L114). So if the validation fails somehow, this will query for the post type of `post` as expected.

I could see how it's strange that the function sometimes doesn't return anything. But it [looks like](https://www.php.net/manual/en/functions.returning-values.php) not returning anything means that it returns `null`:
>If the return is omitted the value NULL will be returned.

Not a strongly-held opinion, though.

kienstra: On second thought, I think it is better to have a return statement at the end of the function in case it hasn't returned. I'll make a commit with that.

kienstra: https://github.com/getblocklab/block-lab/pull/257/commits/95cc6d2705959fee5f5b72b7646068657ffb146c#diff-dc4ef2232132f132dbd59b26a1f8db80R454 returns `null` at the end of the function. I also refactored it so there's only one method that finds all of the eligible post types.

kienstra:
**Sample Page**

Hi @RobStino,
Thanks a lot for testing this and for your thoughts.

It looks like the "Sample Page" is published in the [screencast](https://d.pr/v/wTI3HT). So I'm not sure why it doesn't appear.

It looks to appear in my local:
sample-page-displays

kienstra:
Hi @lukecarbis,

>If it's an easy fix, notice that the user input only saves the username, not the ID? I think it would be better if it matched the Post control. Would you prefer to fix that here or in a separate PR?

Sure, would you mind if I looked at this in another PR? It would have to migrate the values from the previous `string` to an `array()`. Maybe that won't be too complex, though.

>Another inconsistency with the User control is that if you type in invalid username, and then blur the field, the field displays an error. The Post control clears the text (in addition to a red border to indicate that there was a problem). I'm not sure if this was intentional, but I prefer the latter (Post style) way of handling invalid text. Could we make this consistent?

If it's alright, I'll work on this tomorrow.

>Lastly, I agree that we should filter out the Attachment post type. This shouldn't be an option.

Sounds good, https://github.com/getblocklab/block-lab/pull/257/commits/95cc6d2705959fee5f5b72b7646068657ffb146c#diff-dc4ef2232132f132dbd59b26a1f8db80R292 removes the Attachment post type.

kienstra:
Hi @RobStino,
>1. If a page/post has no title it can't be selected, even if it's published. Edge-case issue.

Good catch. I'll work on this tomorrow if that's alright.

>2. The filtered select experience feels a little wrong when you want to re-select. Not a bug, bust something to think about in general

Yeah, it is a little strange. It probably needs more thought.

kienstra:
**If There's No Post Title**

Hi @RobStino,
Thanks for bringing up that you can't select a post if it has no title.

How about not even fetching posts without titles, so the user can't even select them from the Popover?

This control outputs the title for `block_field( 'foo-post' )`. So maybe it's not meaningful to be able to select a post without a title. Though `block_value( 'foo-post' )` exposes the `WP_Post`.

kienstra:
Hi @lukecarbis,
>Another inconsistency with the User control is that if you type in invalid username, and then blur the field, the field displays an error. The Post control clears the text (in addition to a red border to indicate that there was a problem). I'm not sure if this was intentional, but I prefer the latter (Post style) way of handling invalid text. Could we make this consistent?

Good catch, [these commits](https://github.com/getblocklab/block-lab/compare/95cc6d2705959fee5f5b72b7646068657ffb146c...f18216388a036db41ea24cbcb72be12ef0ab01bb) clear the User control when the `<input>` is invalid:

user-control

kienstra:
https://github.com/getblocklab/block-lab/pull/257/commits/aef4922bcdf7d59c066595473e14a05de8f9a5b1 prevents any result from displaying if there's no title. This is open for discussion, please see https://github.com/getblocklab/block-lab/pull/257#issuecomment-484757502

kienstra:
**Request For Another Round Of Review**

Hi @lukecarbis,
Thanks for waiting. Could you please review this again when you can?

It address all of the points except:

>If it's an easy fix, notice that the user input only saves the username, not the ID? I think it would be better if it matched the Post control. Would you prefer to fix that here or in a separate PR?

This requires a little thought about backwards-compatibility, so it'd be great if I could address this in another PR.

>The filtered select experience feels a little wrong when you want to re-select. Not a bug, bust something to think about in general

Maybe we could also consider this separately.

lukecarbis:
See #288.

php/blocks/controls/class-control-abstract.php
```diff
@@ -281,6 +275,31 @@ public function render_settings_post_type_rest_slug( $setting, $name, $id ) {
...
+	 * Gets the REST slugs of public post types, other than 'attachment'.
+	 *
+	 * @return array {
+	 *     An associative array of the post type REST slugs.
+	 *
+	 *     @type string $rest_slug The REST slug of the post type.
+	 *     @type string $name The name of the post type.n
+	 * }
+	 */
+	public function get_post_type_rest_slugs() {
```
lukecarbis, line 288: Great idea separating this out into its own method.

php/blocks/controls/class-control-abstract.php
```diff
@@ -429,16 +448,9 @@ function( $option ) use ( &$options ) {
...
-				( $post_type_object && ( $value === $post_type_object->rest_base ) )
-				||
-				$post_type === $value
-			) {
-				return $value;
-			}
+		if ( array_key_exists( $value, $this->get_post_type_rest_slugs() ) ) {
+			return $value;
 		}
+		return null;
```
lukecarbis, line 457: 👌

lukecarbis (dismissed):
@kienstra Code looks good. For anyone else testing, I had to resave the block I created for testing earlier in this PR, making sure to reselect the correct Post Type.

I want to make one change before we merge this.

lukecarbis:
Actually, never mind! Merge away.

I was hoping to add an "Any Post Type" as the default for the Post Type option, but only just now realised that there's no REST API endpoint for querying all post types at once.

kienstra:
**Made one last commit**

Hi @lukecarbis, sorry to do this last-minute. But I changed `'title'` to `'name'` in the Post control's attributes. 

I've been working on the Taxonomy control, and it will probably share most of the logic with this Post controls.

And I thought `'name'` would be more appropriate than `'title'`, as taxonomies only have a `'name'`.

php/blocks/controls/class-control-abstract.php
kienstra: Thanks, Luke!

lukecarbis (approved):
👍

PR #260, opened 2019-03-21
In Image control, return the image's ID from block_value() when $echo is not true

### LINKED ISSUE #247: Is it possible to use standard image presets when outputting images?
It would be great use the built-in image presets to display images.

For example an option in the inspector to select an image preset, or to print images in a template using a preset.

Something like ```<?php block_field( 'image_field','image_preset' ); ?>```

### LINKED ISSUE #256: Get attachment ID of image
Hey, seems like it's not possible to get the ID of an image?! I would propose something like this:
```
block_value('image') => www.wordpress.com/image.jpg
block_value('image.id') => 12
```
Alternatively we can of course just return the whole WP_Post object.
```
block_value('image')->ID => 12
```

kienstra:
* The behavior shouldn't change for `block_field( 'foo-image-field' )` or `block_value( 'foo-image-field, false )`
* But for `block_value()`, this should make available the attachment ID
* It looks like many functions for attachments take the ID as a parameter, not the `WP_Post`
* Like `wp_get_attachment_image()` and `wp_get_attachment_image_sizes()`

Closes #256. Closes #247.


php/blocks/controls/class-image.php
```diff
@@ -57,4 +57,68 @@ public function register_settings() {
...
+	 * @return int|false The attachment ID, or false if none was found.
+	 */
+	public function get_attachment_id_from_url( $url ) {
+		$image_file = basename( $url );
+		$query      = new \WP_Query(
+			array(
+				'post_type'   => 'attachment',
+				'post_status' => 'inherit',
+				'fields'      => 'ids',
+				'meta_query'  => array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
```
kienstra: Eventually, we might consider storing the ID of the image in the block, in addition to the URL. Right now, it only stores the URL.

This might help avoid running a meta query to get the ID.

kienstra:
This is almost ready for review, pending a little testing on my part.

php/blocks/controls/class-image.php
kienstra: @todo: consider using `attachment_url_to_postid()` instead of this helper method.

kienstra:
**Expected Results For Image Control**

| Function call | Expected result  |  Notes  | 
|---|---|---|
| `block_value( 'foo-valid-url' )`       | returns attachment `ID`   | Different from before, used to return the URL  |
| `block_value( 'foo-invalid-url' )`       | returns `0` | Also different from before. Returns `false` if the URL isn't on the same site, or it doesn't match an existing attachment (image). |
| `block_field( 'foo-valid-or-invalid-url' )`       | echoes the string URL | No validation added in this PR,  simply echoes the string (no change from previous behavior) |   

Updated: `block_value( 'foo-invalid-url' )`  now returns `0`

kienstra:
**Testing Steps**

1. Create a block with an Image control, or use an existing one:
block-image-control

2. Create a new post with that block, and add an image from the same site. External images, like from http://lorempixel.com, will still work with `block_field()`. But `block_value()` won't get the image ID, as they're not from this site.
3. Add a template like `blocks/block-test-image.php`:

```php
<?php
/**
 * Template for test-image block to test the output of the Image control
 */
$field_name = 'test-image';

?>
<p>There should be an image below:</p>
<img src="<?php block_field( $field_name ); ?>">

<p>Here is the result of block_value(), which should be an int:</p>
<?php echo block_value( $field_name ); ?>

<p>And here is the result of calling wp_get_attachment_image() with the result of block_value(). This should be an image:</p>
<?php echo wp_get_attachment_image( block_value( $field_name ) ); ?>

<p>Here is the result of block_value with an invalid field, this should be false an nothing should display below:</p>
<?php wp_get_attachment_image( block_value( 'random-invalid-field-name' ) ); ?>

```

Here's the expected front-end display of that template:

expected-block-lab

kienstra:
**Request For Review**

Hi @lukecarbis and @RobStino,
Could you please review this when you have a chance? There are test steps and expected behavior above.

Thanks!

lukecarbis:
Closes #256.

php/blocks/controls/class-image.php
```diff
@@ -57,4 +57,20 @@ public function register_settings() {
...
+	 *
+	 * @param string $value The value to either make available as a variable or echoed on the front-end template.
+	 * @param bool   $echo Whether this value will be echoed.
+	 * @return string|int|false $value The value to be made available or echoed on the front-end template.
+	 */
+	public function validate( $value, $echo ) {
+		if ( $echo ) {
+			return $value;
+		} else {
+			$id = attachment_url_to_postid( $value );
```
lukecarbis: What an awesome function!

lukecarbis:
This PR works for me as is, and I'd be happy to merge it if everyone else is comfortable, but there's something nagging at me with this approach.

Are we allowing the editor to paste in any old image URL? If the image is not in the media library, this could have unexpected results in the template, as a `false` response would be returned even for valid images.

I propose that we suggest block creators use a "URL" field if they want to support off-site images, or a copy / paste workflow.

For the image field, we should _force_ the media library to be used, to provide block creators as much consistency as possible. In this case we would:

- Removing the text input containing the URL when editing a block
- Provide a preview image when editing a block instead
- Store the Attachment Post ID in the database, instead of the URL
- Fetch the URL _from the ID_ when using `block_field`

This approach will also be more consistent with the User, Post, and Taxonomy fields storing IDs in the database.

php/blocks/controls/class-image.php
kienstra: Haha, I just found out about it

RobStino:
Shall I hold off the review until the above is decided upon?

kienstra:
**ID or URL?**

Hi @lukecarbis,
Good points.

>Are we allowing the editor to paste in any old image URL?

Yes, there's no validation now if the URL is hosted on the site, when using `block_field( 'foo-image-field' )`. Though there is when using `block_value( 'foo-image-field' )`, as it then tries to get the image's `ID`.

Right, I think the URL control would be more appropriate for external images not hosted on the site. I think all 4 of your points above are right.

I could work on converting the existing URL strings to attachment IDs. Of course, there would be a backward-compatibility break if users have external URLs in the Image control.

kienstra:
Hi @RobStino,
>Shall I hold off the review until the above is decided upon?

Yes, this PR might change a lot.

lukecarbis:
@kienstra I think go for the backwards compatibility break – we'll make sure that we communicate that to our users before the next release.

kienstra:
@lukecarbis, sounds good.

js/blocks/controls/image.js
```diff
@@ -1,90 +1,111 @@
...
-			return
-		}
-		if ( 'blob' === image.url.substr( 0, 4 ) ) {
-			// Still uploading…
-			return
-		}
+	} )( ( ownProps) => {
+		const { imageAlt, imageSrc, setAttributes } = ownProps;
+		const attr = { ...ownProps.attributes };
+		const isUploading = 'undefined' !== typeof attr[ field.name ] && 'string' === typeof attr[ field.name ] && 'Uploading' === attr[ field.name ].substr( 0, 9 );
```
lukecarbis: I tried for ages to find a way of setting `isUploading` is a property of the BlockLabImageControl, so it could be set dynamically instead of just relying on the value of the field. I think there's a better way here, but I don't know it. If you don't know it either, happy to commit this change.

php/blocks/controls/class-image.php
```diff
@@ -46,15 +53,26 @@ public function register_settings() {
...
+	 * Validates the value to be made available to the front-end template.
+	 *
+	 * @param string $value The value to either make available as a variable or echoed on the front-end template.
+	 * @param bool   $echo Whether this value will be echoed.
+	 * @return string|int|false $value The value to be made available or echoed on the front-end template.
+	 */
+	public function validate( $value, $echo ) {
+		// Backwards compatibility, as the value used to be the image's URL instead of its post ID.
+		if ( empty( intval( $value ) ) ) {
+			$value = attachment_url_to_postid( $value );
```
lukecarbis: NICE! This means there's no back-compat break, right?

kienstra: Yeah, this should minimize the back-compat break. If someone had entered an external URL in the `<input>`, this wouldn't help. But it'll usually find the ID for a URL that's from the Media Library.

And as you know, the output of `block_value( 'foo-image-field' )` is changing to output an ID instead of the URL string.

js/blocks/controls/image.js
kienstra: Sure, I'll look at this a little more. It looked pretty complex setting an `isUploading` property 😄

kienstra: https://github.com/getblocklab/block-lab/pull/260/commits/dd5ad114c17144f630a1fa4b2640f266d8d982e4 introduces a state of `isUploading`. It seems to work so far.

uploading-image-control

RobStino:
Done some testing.

- Created a new block
- Added an image field for the editor and one for the inspector
- Added the below to my template file

````
<img src="<?php block_field( 'image-editor' ); ?>" alt="">
<img src="<?php block_field( 'image-inspector' ); ?>" alt="">
<?php $image_editor = block_value( 'image-editor' ); ?>
<?php $image_inspector = block_value( 'image-inspector' ); ?>
<?php echo $image_editor; ?>
<?php echo $image_inspector; ?>
````

I then observed the following in the editor preview
image

And saw the following on the front end with 2017 theme.
image


Noted
- block_value() gives me the ID. This is now expected correct?
- block_field() returns the thumbnail URL rather than the original. Is this expected?

**Comment by kienstra** [comment]
Hi @RobStino,
Thanks for testing this.

> block_value() gives me the ID. This is now expected correct?

Yes

> block_field() returns the thumbnail URL rather than the original. Is this expected?

Good point, https://github.com/getblocklab/block-lab/pull/260/commits/bc589e5d42c0d5563c679488fb41da9c48de7054 outputs the full image size.

**Comment by kienstra** [comment]
**Backwards Compatibility**

The expected behavior for this is to have full backward compatibility for `block_field( 'foo' )`, for a URL that was saved in the previous `<input>`. Even with external URLs.

`block_value( 'foo-image' )` is the main difference in this PR. If the user had already saved a URL with the previous `<input>`, this will look for an ID to return for it. In my testing, that's worked well so far.

But for external images, it won't find an ID, and will return `0`.

Also, the behavior for when it doesn't find an ID is changed. It now returns `0` instead of `false`.

File: php/blocks/controls/class-image.php (hunk 4 above)
Comment (kienstra) [reply php/blocks/controls/class-image.php]: [Below](https://github.com/getblocklab/block-lab/pull/260#issuecomment-479475773) are more details on backwards-compatibility.

**Comment by kienstra** [comment]
**Request For Review**

Hi @lukecarbis,
Could you please review this when you can?

`block_field()` should be backwards-compatible. Though `block_value()` now outputs the ID, and `0` for external images. 

So far, it looks like `block_value()` has found the `ID` for Media Library images that were added before this PR.

Also, I added a 'Remove' button if that's alright. Before, you could simply delete the URL from the `<input>`

remove-button

**Review by lukecarbis — APPROVED** [review]

**Review by lukecarbis — CHANGES_REQUESTED** [review]
@kienstra Changes look great. I made a few minor style changes as well.

I noticed a bug, though. Steps to reproduce: 

1. Add a block with a media control
2. Add an image to the block
3. Blur focus from the block to preview (everything looks good)
4. Click the block to edit it again
5. Remove the image
6. Blur focus from the block to preview

I get this error:
Screen Shot 2019-04-04 at 8 05 43 am

The Remove button is a great idea, by the way! Code looks great too.

**Comment by kienstra** [comment]
Hi @lukecarbis, thanks for testing this and your detailed steps to reproduce. I'm working on this now.

**Comment by kienstra** [comment]
**Addressed Bug**

Hi @lukecarbis,
Thanks for finding this bug. 

https://github.com/getblocklab/block-lab/pull/260/commits/82985b33a8f5b3f07f549c2d1f17a4d69c607371 addresses it by setting the attribute to `0` on removing the image. The back-end enforces that it's an int, so setting it to `null` didn't work.

The REST response to the `<ServerSideRender>` call had:

```
{"code":"rest_invalid_param","message":"Invalid parameter(s): attributes","data":{"status":400,"params":{"attributes":"attributes[test-image] is not of type integer."}}}
```

File: php/blocks/controls/class-image.php [hunk 5]
```diff
@@ -46,15 +53,30 @@ public function register_settings() {
 				'sanitize' => 'sanitize_text_field',
 			)
 		);
-		$this->settings[] = new Control_Setting(
-			array(
-				'name'     => 'default',
```
kienstra, line 51: You probably saw this, but this PR removes the default image URL setting:

image-field

This is because this control now stores the integer image ID, not the URL string.

The change below is a backward-compatibility break. It won't honor default image URLs that were entered before.

RobStino:
Hey fellas.
Just noticed 2 issues:
quick screen recording
https://d.pr/v/3RHw8R

kienstra:
**Help Text**

Hi @RobStino,

The help text definitely isn't appearing in the [video](https://d.pr/v/3RHw8R) above.

But it seems to appear alright for me locally:

help-text

kienstra:
**When there's no template**

The [screencast](https://d.pr/v/3RHw8R) also shows that the block displays `0 0` when no template is found. I wish I could reproduce that.

But it looks as expected locally:

template-not-found

kienstra:
I just merged in `develop`, maybe that'll help.

kienstra:
Maybe a fresh `npm install && npm run dev` would help if you haven't done that already.

lukecarbis:
Wiki updated.

lukecarbis (dismissed):

lukecarbis:
@RobStino Can you please test again for help text and weird response when there's no template? The previous help text should now be replaced by a "placeholder", and help text will appear below the buttons (the position isn't easily changed).

js/blocks/loader/editor.scss
```diff
@@ -104,8 +104,7 @@ div[class^="wp-block-block-lab-"],
       .bl-image__remove {
         display: inline-block;
         vertical-align: top;
-        margin-right: 0.6rem;
-        margin-top: 0.6rem;
+        margin: 1rem 0.6rem 0.6rem 0;
```
kienstra: `1rem` of margin-top might be a little too much in the Inspector. Though even the `0.6rem` that I added before was probably too much.

space-above

kienstra:
**Looks Good, One Minor Comment**

Hi @lukecarbis,
The latest 2 commits look good. There's one minor comment above.

The placeholder looks really good, and prevents the block from getting shorter on removing an image:
placeholder

RobStino:
The placeholder square definitely makes me think "drag and drop"...

kienstra:
**Dropzone Added**

Hi @RobStino and @lukecarbis,
https://github.com/getblocklab/block-lab/pull/260/commits/dc529cedb6a157b7c8aceacd3bfeb07a9f2371f6 allows dragging and dropping an image into the placeholder:

drop-zone

And https://github.com/getblocklab/block-lab/pull/260/commits/7da2e98d55d9acf82960462721526e8e60789bcf makes a slight styling change:
less-space

js/blocks/controls/image.js
```diff
@@ -68,15 +71,35 @@ const BlockLabImageControl = ( props, field, block ) => {
 			setAttributes( attr );
 		}
 
+		const uploadFiles = ( files ) => {
```
kienstra, line 74: This is copied verbatim [from below](https://github.com/getblocklab/block-lab/pull/260/commits/dc529cedb6a157b7c8aceacd3bfeb07a9f2371f6#diff-7bbb33042f7b94c527dc6a5bdb42c258L94).

RobStino:
@kienstra Great functionality on the drag-n-drop.
It just needs some some styling .
Let me demo...
image

I just changed the border to `border: 3px dashed #e2e4e7;`

The icon will just need to be better positioned.

kienstra:
Sure, https://github.com/getblocklab/block-lab/pull/260/commits/8d04908ec24178e518c1b7b9e9765cd5e783206f applies your styling. I'm working on centering the placeholder.

kienstra:
image-control

kienstra:
@RobStino, thanks for testing this. The last 2 commits apply your suggestions.

RobStino:
Awesome!
I did spot something that looks like it's come over from the native Gutenberg drag-drop component. Some white text appears when you drag the image in to the drop zone.
See this video for an example.
https://d.pr/v/q5j5iJ

kienstra:
**Label Text**

Hi @RobStino,
Was this the text (circled in green below)?
>Drop files to upload

label-here

I could remove that if needed by passing `label=" "` to `<DropZone>`

But how about changing it to:
>Drop file to upload

This would change 'files' to 'file', as you can only upload one file.

lukecarbis:
@kienstra Nice work on the drop zone! I'll try to take a pass at this today and see if I can find something that looks good visually.

kienstra:
Thanks, @lukecarbis!

lukecarbis (dismissed):

kienstra:
**Looks Good**

Hi @lukecarbis,
The [last 2 commits](https://github.com/getblocklab/block-lab/pull/260/files/e8d87fb7ec3cfd7f0a54b1f1f80a5df2a7f3c5de..60cb8c0c369fb0a28b004687e9724d4bf1c2d4f7) look good, and they worked locally. I like the help text as it is with your commits, above the image:

help-text-could-clutter

RobStino (dismissed):
Tested and looks great to me!

lukecarbis:
Great! I'll wait for #276 to be finished and merged, then I'll merge develop, then I'll merge this PR (but only if Travis is happy).

kienstra:
Haha, sounds good.

lukecarbis:
@kienstra Could you please have a quick look at 7cda2c8 to make sure it makes sense?

php/blocks/controls/class-image.php
```diff
@@ -63,8 +63,10 @@ public function register_settings() {
 	 * @return string|int $value The value to be made available or echoed on the front-end template, possibly 0 if none found.
 	 */
 	public function validate( $value, $echo ) {
+		$image_id = intval( $value );
```
kienstra, line 66: Hi @lukecarbis, this looks good. I haven't tested it, but if you're happy, I'm happy.

kienstra: Good idea to set up Travis, it's good that it caught this.

lukecarbis (approved):

kienstra:
**Backwards Compatibility**

This PR has a backwards-compatibility break with `block_value( 'foo-url' )`, but not `block_field( 'foo-url' )`

`block_value( 'foo-url' )` now returns an `integer` attachment ID or `0`, as shown in the top 2 rows. 

| Function call | Expected result  |  Notes  | 
|---|---|---|
| `block_value( 'foo-valid-url' )`       | returns attachment `ID`   | Different from before, used to return the URL  |
| `block_value( 'foo-invalid-url' )`       | returns `0` | Also different from before. Returns `0` if the URL isn't on the same site, or it doesn't match an existing attachment (image). |
| `block_field( 'foo-valid-or-invalid-url' )`       | echoes the string URL | No validation added in this PR,  simply echoes the string (no change from previous behavior) |   

But `block_field( 'foo-url' )` should have the same result as before, even if it used to have an external URL.

And if the URL was previously for an external image, not an image hosted in the Media Library, `block_value( 'foo-url' )` will now return `0`.

AndrewSheetMetal:
Hey guys, just a quick question:
We need this feature for a project. When can we expect this feature to be distributed in a release, so we get it when we update the plugin via wordpress?
Can I otherwise download the plugin here from github?
Thanks!

kienstra:
**When are we releasing this?**

Hi @AndrewSheetMetal,
Sure, we're planning on releasing it this coming week. and you could download it from the [WordPress.org repo](https://wordpress.org/plugins/block-lab/). If you could, it'd be good to wait until then to download it. If it's for non-production use, you could checkout the `develop` branch from this GitHub repo, and follow the [Local Setup](https://github.com/getblocklab/block-lab/blob/develop/CONTRIBUTING.md#local-setup) steps.

AndrewSheetMetal:
Hey @kienstra,
thanks for the fast reply! I'm going to wait for the release then.

PR #268, opened 2019-03-27
Design adjustments to the Add / Edit Block interface.

lukecarbis:
This may require you to delete the `meta-box-order_block_lab` user meta, so that the sidebars are placed in their default position (instead of their saved position). Or else, just create a new user for testing this.


lukecarbis:
@kienstra @RobStino Ready for you to test.

Look for:
- Block properties in sidebar
- Fixed height when adding lots of fields

Please test all screen sizes (including mobile). Keen for feedback on usability – consider this an experiment that we can keep or change or just drop parts of.

php/post-types/class-block-post.php
```diff
@@ -447,26 +439,28 @@ public function render_fields_meta_box() {
...
-						<td colspan="5" class="block-fields-rows">
-							<p class="block-no-fields">
+						<td colspan="5">
+							<div class="block-fields-rows">
+								<p class="block-no-fields">
+									<?php
+									echo wp_kses_post(
+										sprintf(
+											// Translators: Placeholders are for <strong> HTML tags.
+											__( 'Click the %1$s+ Add Field%2$s button below to add your first field.' ),
```
kienstra, line 466: This code was here before this PR. But maybe this would be easier for translators:
```php
__( 'Click the <strong>+ Add Field</strong> button below to add your first field.' )
```
Sometimes `__()` calls in Core [have markup](https://github.com/WordPress/wordpress-develop/blob/c833087131091fc52265dd5de22c96bc6a53b798/src/wp-admin/admin-footer.php#L35) in them.

php/post-types/class-block-post.php
```diff
@@ -252,8 +252,8 @@ public function add_meta_boxes() {
 			__( 'Block Properties', 'block-lab' ),
 			array( $this, 'render_properties_meta_box' ),
 			$this->slug,
-			'normal',
-			'high'
+			'side',
+			'default'
```
kienstra, line 268: What do you think about changing `default` to `low`?

Right now, the "Template" meta box isn't visible "above the fold" on a Macbook 15 inch display:

template-meta-box

Changing this to `low` moves the "Template" meta box above the "Block Properties" box:

moved-above

kienstra:
The icon button looks a lot better. It takes up less space than before:

icon-picker

kienstra:
**Scrolling**

It's nice how the field scrolls to the top on selecting it:

field-scrolls-to-top

kienstra (dismissed):
**Approved**

Hi @lukecarbis,
Nice work, this has some really nice improvements in the block post type UI.

kienstra:
**Testing On Mobile And Tablet**

# Mobile (Galaxy S7)
Looks good
galaxy-s7

# Tablet (iPad 6th Generation)
Also looks good
ipad-6

lukecarbis:
I agree with moving the template block – I'll experiment with putting it in the main area instead of the sidebar.

lukecarbis:
Here's my attempt at rehashing the template meta box. It now only displays if a template is _needed_.



Otherwise, we get this (styled to match the "Permalink" in other post types):

Screen Shot 2019-03-28 at 11 15 32 pm

Keen for feedback on this! I like it because the "New Block" screen is much more balanced.

kienstra:
**Looks Great**

Hi @lukecarbis,
The display for when the template exists looks really good. It's cleaner:

when-template-exists

The display for when there's no template also looks better. It's nice how you can edit the template path.
looks-good

What do you think about not allowing a user to edit the `.php` extension?

Also, it looks like edits to the path aren't saved on updating the page. But it seems like this commit didn't intend to validate and save that value yet.

RobStino:
Just reviewed the changes here.
In general, I really like this!
2 things I think need attention:
- The template file _copy paste_ text-input makes it feel like you can edit the file path/slug. Which isn't accurate.
- The icon size or spacing just needs some pixel-nudging.
Walkthrough of the above -> https://d.pr/v/SpiC21

lukecarbis:
@kienstra @RobStino Great thoughts, thanks. I've updated the PR with a better click-to-copy implementation, and some pixel nudging on the icons. Please review carefully, as these commits may have occurred while under the influence.

Screen Shot 2019-03-30 at 9 06 54 pm

lukecarbis:
Also, just wanted to show my thinking on the icon pixel nudging, given different scrollbar settings:

php/post-types/class-block-post.php
```diff
@@ -213,24 +214,34 @@ public function enqueue_scripts() {
...
 			wp_localize_script(
 				'block-post',
 				'blockLab',
 				array(
 					'fieldSettingsNonce' => wp_create_nonce( 'block_lab_field_settings_nonce' ),
+					'copySuccessMessage' => __( 'Copied to clipboard.', 'block-lab' ),
+					'copyFailMessage'    => sprintf(
+						// translators: Placeholder is a shortcut key combination.
+						__( '%1$s to copy.', 'block-lab' ),
+						strpos( getenv( 'HTTP_USER_AGENT' ), 'Mac' ) ? 'Cmd+C' : 'Ctrl+C'
```
kienstra, line 238: Nice. good idea to show `'Cmd+C'`or `'Ctrl+C'` based on the user agent.

php/post-types/class-block-post.php
```diff
@@ -653,72 +672,90 @@ class="regular-text"
...
+				</span>
+			</p>
+			<p>
+				<strong><?php esc_html_e( 'Learn more:', 'block-lab' ); ?></strong>
+				<?php
+				echo wp_kses_post(
+					sprintf(
+						'<a href="%1$s" target="_blank">%2$s</a> | ',
+						'https://github.com/getblocklab/block-lab/wiki/3.-Displaying-custom-blocks',
+						esc_html__( 'Block Templates', 'block-lab' )
```
kienstra, line 701: Good idea to escape this.

kienstra:
**Copying UI**

Copying the template works really well:
copied-to-clipboard

Nice tooltip, I hadn't thought of doing something like that before.

php/post-types/class-block-post.php
```diff
@@ -653,72 +672,90 @@ class="regular-text"
 	 */
 	public function render_template_meta_box() {
 		global $post;
+		?>
+		<div class="template-notice">
+			<h3><span class="dashicons dashicons-yes"></span><?php esc_html_e( 'Next step: Create a block template.', 'block-lab' ); ?></h3>
```
kienstra, line 675: copy-template

php/post-types/class-block-post.php
```diff
@@ -653,72 +672,90 @@ class="regular-text"
...
+			</p>
+		</div>
+		<?php
+	}
+
+	/**
+	 * Display the template location below the title.
+	 */
+	public function template_location() {
+		global $post;
```
kienstra: This is a minor point, but maybe this could be:

```php
$post = get_post();
```

It works in my local 😄

js/admin.block-post.js
```diff
@@ -34,6 +34,32 @@
 			$( '#block-properties-icon-current' ).html( svg );
 		});
 
+		$( '#block_template .template-location a.filename' ).on( 'click', function( event ) {
+			event.preventDefault();
+
+			let copy  = $( '#block_template .template-location .click-to-copy' ),
```
kienstra, line 40: No need to change this, but I've lately been using `const` instead of `let` whenever the variable won't be reassigned, like here.

kienstra:
**Approved**

Hi @lukecarbis,
This looks good!

Not sure if you were looking for a final review, but it looks good from my perspective.

kienstra (dismissed):

php/post-types/class-block-post.php
lukecarbis: I'm open to any suggestions on the copy, too. cc. @RobStino

js/admin.block-post.js
lukecarbis: I'll leave it as is in this instance, for consistency with the rest of the script.

php/post-types/class-block-post.php
lukecarbis: Done in e463246.

lukecarbis:
@kienstra Can I please request another code review on e463246 and feabbd1?

RobStino:
@lukecarbis great work on the click-to-copy. Super clear now.
2 more thoughts
**1. Copy update**
_Used to determine the location of the template file._
change to
_Used to determine the name of the template file._
**2. Live slug -> filename update**
Is it possible to have the filename in the click-to-copy area updated live if the slug is changed? Or, because the slug is only changed if *Published*, it makes more sense to updated on reload. What do you think?

lukecarbis:
@RobStino Copy update made. I tried out changing the filename live as the slug is changed, but the JS is messy, and it doesn't quite feel right since the changes haven't yet been saved.

php/post-types/class-block-post.php
```diff
@@ -198,8 +198,7 @@ public function add_caps() {
 	 * @return void
 	 */
 	public function enqueue_scripts() {
-		global $post;
-
+		$post   = get_post();
```
kienstra, line 201: Nice, thanks

tests/php/post-types/test-class-block-post.php
```diff
@@ -114,6 +114,82 @@ public function test_get_field_value() {
...
+	 */
+	public function test_render_properties_meta_box() {
+		$this->load_dummy_block();
+
+		ob_start();
+		$this->instance->render_properties_meta_box();
+		$properties_meta_box = ob_get_clean();
+
+		$this->assertNotEmpty( $properties_meta_box );
+		$this->assertGreaterThan( 0, strpos( $properties_meta_box, 'block-properties-slug' ) );
```
kienstra, line 151: Nice, I didn't know about `assertGreaterThan()`

tests/php/post-types/test-class-block-post.php
```diff
@@ -126,4 +202,21 @@ public function set_valid_license() {
 			)
 		);
 	}
+
+	/**
+	 * Initialises a dummy block.
+	 */
+	public function load_dummy_block() {
```
kienstra, line 209: Good idea to have a helper function that loads the block.

php/post-types/class-block-post.php
kienstra: Thanks, this looks good.

kienstra (approved):
**Approved**
 
Hi @lukecarbis,
The [last 3 commits](https://github.com/getblocklab/block-lab/compare/53ce0c0c7a86d22dd32716e39035519b4e8e84c5...3de67dee3955c8f402a3c4e9d8b50f204171539c) look good, and they worked as expected when I tested them briefly.

PR #269, opened 2019-03-27
Add onboarding class which creates some dummy data

lukecarbis:
- [x] Dummy data
- [x] "Get Started" message on plugin activation
- [x] Notice above the first "dummy" block, with CTA to edit the block
- [x] Notice in the "Fields" metabox of the first dummy block, with CTA to add new fields
- [x] Notice after publishing the dummy block, with CTA to download / add a template to your theme
- [x] Notice after adding the template for the dummy block, with CTA to use the block in a post
- [x] Add link to / information about template functions in the "Template" metabox

- ~~Highlight the dummy block in the block selector the first time it's used~~ (too complicated)


lukecarbis:
Merged the `feature/repeater` branch in here.

lukecarbis:
@kienstra @RobStino Would love some feedback on the "Welcome" onboarding message.

Screen Shot 2019-07-28 at 4 35 19 pm

lukecarbis:
Ready for review!

Please pay careful attention to:
- Language / messaging / copy / tone. Took a couple of goes at most messages, but happy for feedback, changes, and refinements.
- Technical implementation (of course)

Video preview: https://imgur.com/a/FHs41Sc

RobStino:
Overall, this is awesome! Love the vibe and tone!

Couple of thoughts and a (I think) couple of bugs.

**Thoughts**
- Each of the prompts throughout the journey is styled slightly differently. It might make it easier to follow the onboarding journey if they were all styled the same.
- Perhaps adding something like "Step 1" etc to each prompt would help as well.
- In the "add a template file" prompt, could we adjust the copy here to be specific for onboarding? I.e. and instruction to come back to this screen and refresh once the template file has been added?

**Bugs**
- I only saw the pointing hand once. I loved it, but is it meant to be used throughout?
- After I added the template file, I didn't see any more prompts. I was meant to get one about adding the new block in the editor right?

lukecarbis:
@RobStino 

- The first prompt is different, yes, but the rest should be basically the same. Could you point out specifically where they could be brought into better alignment? I think keeping the primary one different to the rest is okay.

- I'm worried adding "Steps" will make things seem to formal and structured. I purposefully left them out, so that learning to use the plugin would seem more "organic" and contextual.

- Yes, change the "add template prompt" during onboarding is a great idea! Will do.

- The pointing hand only appears twice in the journey, and one of those could be skipped. If you choose not to action the primary Welcome dialog, and instead go to All Blocks (which could happen at any point)… if there's only one block and it's the example block, you'll see another onboarding message with a hand – you might have missed this one?

- You should see one more prompt after adding a template file – yes. Couldn't reproduce on my end, could you do a screen recording?

php/admin/class-onboarding.php
```diff
@@ -0,0 +1,392 @@
...
+		$screen = get_current_screen();
+		$slug   = $this->plugin->block_post->slug;
+
+		if ( ! is_object( $screen ) ) {
+			return;
+		}
+
+		$post_id = filter_input( INPUT_GET, 'post', FILTER_SANITIZE_NUMBER_INT );
+
+		/**
```
kienstra: Maybe the `/**` could instead be `/*` 

I think the `/**` at the top of a comment is only for DocBlocks.

php/admin/class-onboarding.php
```diff
@@ -0,0 +1,392 @@
...
+		</div>
+		<?php
+	}
+
+	/**
+	 * Render the Add Fields message.
+	 */
+	public function show_publish_notice() {
+		$post  = get_post();
+		$block = new Block( $post->ID );
```
kienstra: There's nothing wrong with this, but maybe using [get_the_ID()](https://developer.wordpress.org/reference/functions/get_the_ID/) here would be simpler:

```php
$block = new Block( get_the_ID() );
```

php/admin/class-onboarding.php
```diff
@@ -0,0 +1,392 @@
...
+		?>
+		<div class="block-lab-publish block-lab-notice">
+			<h2>🧪 <?php esc_html_e( 'Time to experiment!', 'block-lab' ); ?></h2>
+			<p class="intro"><?php esc_html_e( 'Choose an icon, change the category, and investigate a few different field types.', 'block-lab' ); ?></p>
+			<p>
+				<?php
+				echo wp_kses_post(
+					sprintf(
+						// translators: Placeholders are <strong> html tags.
+						__( 'When you\'re ready, save your block by pressing %1$sPublish%2$s.', 'block-lab' ),
```
kienstra, line 247: If you'd like, I think it's [fine to pass](https://github.com/WordPress/wordpress-develop/blob/d19f4d7ff4695b1cec812000c14fba18aaa4864b/src/wp-includes/comment.php#L1216) `<strong>` in the string, like:

```php
__( 'When you\'re ready, save your block by pressing <strong>Publish</strong>.', 'block-lab' ),
```

Likewise in `show_edit_block_notice()`.

php/admin/class-onboarding.php
```diff
@@ -0,0 +1,392 @@
...
+
+	/**
+	 * Create a dummy starter block when the plugin is activated for the first time.
+	 */
+	public function add_dummy_data() {
+		// Note: wp_count_posts() does not work here.
+		$blocks = get_posts(
+			array(
+				'post_type'   => block_lab()->block_post->slug,
+				'numberposts' => '-1',
```
kienstra: Would it be possible to set some kind of limit?

php/admin/class-settings.php
```diff
@@ -57,7 +57,7 @@ public function enqueue_scripts() {
 	 */
 	public function add_submenu_pages() {
 		add_submenu_page(
-			'edit.php?post_type=block_lab',
+			'edit.php?post_type=' . block_lab()->block_post->slug,
```
kienstra: Great idea to use this instead of hard-coding it.

php/views/upgrade.php
```diff
@@ -9,7 +9,7 @@
 
 ?>
 <section class="container">
-	<div class="dashboard_welcome tile" style="background-image: url('https://getblocklab.com/wp-content/uploads/2019/02/Block-Lab-Pro-Hero-Background-1.svg');">
+	<div class="dashboard_welcome tile">
```
kienstra, line 12: It's nice how the `background-image` is now a CSS file, instead of inline.

php/class-plugin.php
```diff
@@ -20,6 +20,39 @@ class Plugin extends Plugin_Abstract {
...
+
+	/**
+	 * Execute this as early as possible.
+	 */
+	public function init() {
+		$this->block_post = new Post_Types\Block_Post();
+		$this->register_component( $this->block_post );
+
+		$this->loader = new Blocks\Loader();
+		$this->register_component( $this->loader );
```
kienstra: Great idea to move this into `init()`.

php/admin/class-onboarding.php
```diff
@@ -0,0 +1,392 @@
...
+
+		if ( ! $post || ! isset( $post->post_name ) || empty( $post->post_name ) ) {
+			return;
+		}
+
+		if ( 'publish' !== $post->post_status ) {
+			return;
+		}
+
+		$template = block_lab_locate_template( 'blocks/block-' . $post->post_name . '.php', '', true );
```
kienstra: This is fine as is, but I might have used:

```php
"blocks/block-"{$post->post_name}.php"
```

...instead of:

```php
'blocks/block-' . $post->post_name . '.php'
```

It's just a personal preference, though.

kienstra (commented):
**Quick Review**

Hi @lukecarbis,
Sorry I haven't been very active lately. Here's a quick code review of this, though I haven't tested it. I might have more comments after testing it.

php/admin/class-onboarding.php
lukecarbis: TIL!

lukecarbis: Fix in e5a6585.

php/admin/class-onboarding.php
lukecarbis: 1 aught to work!

php/admin/class-onboarding.php
lukecarbis: My preference is to not include HTML in strings. If you need to change the tags (e.g. add a class or ID), then the string becomes fuzzy and needs a translator to fix again.

php/admin/class-onboarding.php
lukecarbis: Fixed in 2705c6c.

php/admin/class-onboarding.php
lukecarbis: Fixed in 86fbdb3.

lukecarbis:
@kienstra Thanks for the review! Made some changes – could you please have another look?

php/admin/class-onboarding.php
kienstra: Sure, that's fine. That makes sense.

kienstra: Good point that changing the tag requires a new translation, I haven't thought of that.

kienstra (approved):
**Approved**

Hi @lukecarbis,
Thanks for addressing my minor points here. This looks good from a code perspective, though I haven't tested it. If you'd like, I could test it much later today, maybe in 4-8 hours.

RobStino:
@lukecarbis Looks good!
I got the prompt to create a new post/add block this time.

php/admin/class-onboarding.php
kienstra: Nice!

php/class-plugin.php
```diff
@@ -21,19 +21,45 @@ class Plugin extends Plugin_Abstract {
...
-	 * @return bool
+	 * @var Blocks\Loader
+	 */
+	public $loader;
+
+	/**
+	 * Execute this as early as possible.
+	 */
+	public function init() {
+		$this->set_util();
```
lukecarbis: @kienstra – how do you feel about implementing this here? Is it better to do it in `block-lab.php` instead? What setting the `Util` in a parent `init` method?

kienstra: Hi @lukecarbis,

To go along with this `init()` method, maybe it'd be good to simply remove `set_util()` and put the function's contents here:

```php
$this->util = new Util();
$this->register_component( $this->util );
```

kienstra: PR #384, targeting this PR's branch, implements this. It also changes the way the post type slug is stored. 

This is your PR, so it's up to you whether to use that approach, or part of it. I didn't want to push straight to this branch.

kienstra:
Sorry, I know the Repeater PR (#298) is blocking this.

kienstra:
Sorry for the big merge conflict from the repeater branch. It's a big refactoring.

PR #275, opened 2019-04-04
Version bump

lukecarbis:


RobStino (approved):
Tested. All good

kienstra (approved):
**Approved**

Hi @lukecarbis,
This looks good. I don't see any references to the previous version of `1.2.1`, except where we would expect, like in the Changelog.

PR #276, opened 2019-04-09
Add a .travis.yml file

kienstra:
* Mainly as generated by `wp plugin scaffold <plugin name>`
* Makes some adjustments, like changing the lowest PHP version from `5.3` to `5.4`
* This might need some more changes, like changing the branch from master to something like develop.


.travis.yml
```diff
@@ -0,0 +1,63 @@
...
+  email:
+    on_success: never
+    on_failure: change
+
+cache:
+  directories:
+    - $HOME/.composer/cache
+
+matrix:
+  include:
```
lukecarbis, line 16: We probably only need to test `latest` and `trunk`,

kienstra: OK

kienstra: Sure, https://github.com/getblocklab/block-lab/pull/276/commits/f41885acabb849c45be645209a353c92c104492b removes WordPres 4.5 from the build.

phpcs.xml
```diff
@@ -6,6 +6,10 @@
 
     <config name="testVersion" value="5.4-"/>
 
+    <arg value="s"/>
```
kienstra, line 9: I renamed this so that `phpcs` in Travis will pick this up, without having to pass it as an argument, like `phpcs --standard=phpcs.ruleset.xml`

kienstra: If this name is important, we could keep the name and just pass it as an argument in `.travis.yml`

lukecarbis (approved):

Issue #270, opened 2019-03-28
Error loading block: Cookie nonce is invalid


MattKevan:
I've just pushed my site from local to staging, and now most of the Block Lab blocks I created are displaying blank with the error ```Error loading block: Cookie nonce is invalid```. Blocks display fine on the front end, it's just the editor preview that's broken.

Looking in the browser web inspector I get a 403 forbidden error for each block that's not displaying. 

`http://co-create.flywheelsites.com/wp-json/wp/v2/block-renderer/block-lab/cc-hero-banner?context=edit&attributes% ... post_id=5&_locale=user
Failed to load resource: the server responded with a status of 403 (Forbidden)`

There seems to be an issue with Gutenberg if the site URL and Wordpress URLs are different, but in my case they're the same.

https://github.com/WordPress/gutenberg/issues/1761

Any thoughts on this would be appreciated.

lukecarbis:
@MattKevan I encountered this same error once while demoing Block Lab live to an audience, and I couldn't figure out what was going wrong. It ended up fixing itself, which made me wonder if there could be a cache or transient issue at work here.

Thanks for the instructions on how to reproduce the problem, I'll try to replicate it when I next get a chance.

MattKevan:
Looking into it a bit deeper, only blocks with images have the problem. Delete the image and the error goes away, return the image and so does the error. 

I found this applied to both newly-created blocks and ones that had been pushed to stage with the migration. I also have caching switched off, so I'd have expected the problem to have gone away by now if it was a cache issue.

lukecarbis:
@MattKevan If #285 doesn't work, could you try disabling WP Super Cache (or any other caching plugin that you've got installed), and resaving your permalinks?

If that still doesn't help, try following some of the steps outlined here:
https://wordpress.org/support/topic/jetpack-4-3-error-cookie-nonce-is-invalid/

lukecarbis:
For what it's worth, I received this error when I was connected to a wireless network that was trying to redirect traffic to its network login screen. I have a strong feeling this is happening because WordPress is not receiving the URLs it expects.

kienstra:
Hi @MattKevan, 
Thanks for letting us know about this.

Have you seen problems like this with other parts of the block editor, not just Block Lab?

lukecarbis:
[...]

PR #285, opened 2019-04-14
Workaround for users with different Site Address and WordPress URL settings

lukecarbis:
@mattkevan I couldn't manage to replicate #270 exactly, but I did replicate the error you referenced. Can you check whether this fix resolves the problem you reported?

### DIFF
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -79,7 +79,6 @@ class Loader extends Component_Abstract {
 		}
 	}
 
-
 	/**
 	 * Launch the blocks inside Gutenberg.
 	 */
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -63,6 +63,20 @@ class Loader extends Component_Abstract {
 		 * PHP block loading.
 		 */
 		add_action( 'plugins_loaded', array( $this, 'dynamic_block_loader' ) );
+
+		/**
+		 * Workaround for users with different Site Address and WordPress URL settings.
+		 *
+		 * @see https://github.com/WordPress/gutenberg/issues/1761
+		 */
+		if ( is_admin() && home_url() !== site_url() ) {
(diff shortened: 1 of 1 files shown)


lukecarbis:
@kienstra Even if this doesn't resolve #270, it does resolve a problem for users with a different Site URL and Home URL. See https://github.com/WordPress/gutenberg/issues/1761.

kienstra:
**Looks Good**
*Question About Where This Should Be Fixed*

Hi @lukecarbis,
The code looks good, but do you think Block Lab is the right place to fix this?

You understand this issue better than I do. But in that issue you linked to, it looks like this affects the block editor and the REST API, not just Block Lab: https://github.com/WordPress/gutenberg/issues/1761#issuecomment-387912777

>Given the impact of this issue was a deliberate design decision for the REST API...

It'd be interesting to see how this works for @mattKevan, or if he had this issue with the block editor in general, or only this plugin.

kienstra:
I couldn't reproduce the issue when I set the WordPress Address (home) to a different URL than the Site Address (siteurl):

different-from-wp-address

With the `develop` branch checked out, adding blocks worked fine:

diff-site-url

Maybe there's a better way to reproduce this issue, though.

kienstra (commented):
I couldn't reproduce the issue when I set the WordPress Address (home) to a different URL than the Site Address (siteurl):

different-from-wp-address

With the `develop` branch checked out, adding blocks worked fine:

diff-site-url

Maybe there's a better way to reproduce this issue, though.

lukecarbis:
@kienstra I agree that block lab might not be the right place for this fix. 

Try reproducing by changing the domain in the URL (not just the path).

lukecarbis:
Given that this is an issue that affects all dynamic blocks (in fact, any block that makes a REST API call), I'm going to suggest we allow the fix to come from the Gutenberg team.

PR #289, opened 2019-04-22
Add a Taxonomy control

### LINKED ISSUE #277: "Taxonomy" control type
It would be great if we could select a particular taxonomy to choose from as a control setting, and then see a searchable dropdown list of terms in that taxonomy in the post editor.

kienstra:
* Similar to the Post control, there's a `<select>` to choose the taxonomy type

Closes #277

### DIFF
--- a/js/blocks/controls/content.js
+++ b/js/blocks/controls/content.js
@@ -1,0 +1,48 @@
+import FetchInput from './fetch-input';
+
+/**
+ * Gets a content control, eg. a Post or Taxonomy control.
+ *
+ * @return {Function} A component for a control.
+ */
+export default ( props, field, getNameFromAPI ) => {
+	const { setAttributes } = props;
+	const attr = { ...props.attributes };
+	const DEFAULT_ID = 0;
+	const DEFAULT_NAME = '';
+
+	/**
+	 * Gets the ID from an API response.
+	 *
+	 * @param {Object} value The value in which to look for the ID.
+	 * @return {Number} The ID from the value, or 0.
+	 */
+	const getIdfromAPI = apiResponse => ( apiResponse && apiResponse.id ) ? parseInt( apiResponse.id ) : DEFAULT_ID;
+
+	attr[ field.name ] = Object.assign( { id: DEFAULT_ID, title: DEFAULT_NAME }, attr[ field.name ] );
+	const valueAttribute = attr[ field.name ];
+
+	return (
+		<FetchInput
+			field={field}
+			apiSlug={field.rest_slug}
+			value={valueAttribute['id']}
+			displayValue={valueAttribute['title']}
+			getValueFromAPI={getIdfromAPI}
+			getDisplayValueFromAPI={getNameFromAPI}
+			onChange={value => {
+				if ( 'string' === typeof value ) {
+					// The value is probably from the user typing into the <input>.
+					valueAttribute['title'] = value;
+					valueAttribute['id'] = DEFAULT_ID;
+				} else {
+					// The value is probably an Object, from the user selecting a link in the Popover.
+					valueAttribute['title'] = getNameFromAPI( value );
+					valueAttribute['id'] = getIdfromAPI( value );
+				}
+				attr[ field.name ] = valueAttribute;
+				setAttributes( attr );
+			}}
+		/>
+	);
+}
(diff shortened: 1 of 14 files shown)


js/blocks/controls/content.js
```diff
@@ -0,0 +1,48 @@
+import FetchInput from './fetch-input';
```
kienstra: The name of this file, `content.js`, probably isn't the best. 

It's a function that content controls use: Post and Taxonomy.

kienstra:
# Block Lab > Edit Block 

taxonomy-control

# Block UI

taxonomy-control-block

kienstra:
**Function Results For Taxonomy Control**

|   | `block_value()`  | `block_field()`  |
|---|---|---|
| term exists for ID | `WP_Term` object  | string term name  |
| term doesn't exist for ID  | `null`  |  `'No'` |

kienstra:
**Testing Template**

In case this might help, here's a testing template I've been using:

```php
<?php
/**
 * Template to test the output of the Taxonomy control.
 */

$field_name = 'test-taxonomy'; // Change this as needed.
$term       = block_value( $field_name );

?>
<p>There should be a taxonomy name below:</p>
<span><?php block_field( $field_name ); ?></span>

<p>Here is the result of block_value()->term_id, which should be an int:</p>
<?php echo $term->term_id; ?>

<p>Here is the class name of the block_value() return value, this should be WP_Term:</p>
<?php echo get_class( $term ); ?>

<p>And here is the result of block_value()->name</p>
<?php echo $term->name; ?>

<p>Here is the result of calling block_field(), this should be a string term name: </p>
<?php block_field( $field_name ); ?>

<p>Here is the result of calling block_field() with an invalid field argument, 'No' should display below:</p>
<?php block_field( 'bipb-random-some-invalid-field-name' ); ?>

```

js/blocks/controls/post.js
```diff
@@ -1,63 +1,15 @@
-import FetchInput from './fetch-input';
```
kienstra, line 1: This shouldn't change the Post control, only refactor most of it into `content.js`.

kienstra:
**Request For Review**

Hi @lukecarbis and @RobStino,
Could you please review this? It's similar to the Post control, in that it allows [selecting the taxonomy type](https://github.com/getblocklab/block-lab/pull/289#issuecomment-486955922). 

Thanks!

php/blocks/controls/class-control-abstract.php
```diff
@@ -261,18 +261,8 @@ public function render_settings_textarea_array( $setting, $name, $id ) {
...
-			foreach ( $rest_slugs as $rest_slug => $post_type_name ) :
-				?>
-				<option value="<?php echo esc_attr( $rest_slug ); ?>" <?php selected( $rest_slug, $setting->get_value() ); ?>>
-					<?php echo esc_html( $post_type_name ); ?>
-				</option>
-			<?php endforeach; ?>
-		</select>
-		<?php
+		$values = $this->get_post_type_rest_slugs();
+		$this->render_select( compact( 'setting', 'name', 'id', 'values' ) );
```
lukecarbis: I'm happy to run with this if you are, just thought I'd point out that many consider `compact()` bad practice (as I'm sure you know).

php/blocks/controls/class-user.php
```diff
@@ -3,7 +3,7 @@
  * User control.
  *
  * @package   Block_Lab
- * @copyright Copyright(c) 2018, Block Lab
+ * @copyright Copyright(c) 2019, Block Lab
```
lukecarbis, line 6: Nice pickup.

kienstra: Thanks!

php/blocks/controls/class-control-abstract.php
kienstra: Yeah, good point. I'll look at this now.

kienstra: https://github.com/getblocklab/block-lab/pull/289/commits/bb7ca6d2ef34babe61e3c7627a0de251ed391ce2 replaces the `compact()` calls with arrays.

lukecarbis (approved):
Looks good!

kienstra:
Thanks, @lukecarbis!

PR #292, opened 2019-04-23
Prevent gulp from packaging the bin directory

lukecarbis:


kienstra (approved):
**Approved**

Hi @lukecarbis,
This looks good, good catch.

PR #296, opened 2019-04-26
Fix an issue for the Post control where block_field() does not output anything

kienstra:
* Before, `post.js` saved the post's title in the attributes [as name](https://github.com/getblocklab/block-lab/blob/739872aab54b1e1edd91b1c120bd60855c6625ff/js/blocks/controls/post.js#L49), but the `validate()` method [looks for title](https://github.com/getblocklab/block-lab/blob/739872aab54b1e1edd91b1c120bd60855c6625ff/php/blocks/controls/class-post.php#L74)
* I made a [last-minute commit](https://github.com/getblocklab/block-lab/commit/0d6928fce7ea59b9e3423f94e6878d5d0199fb35) before the last `v1.2.3` release to make it save the `post_title` as `name` in the attributes instead of `title`
* This PR fixes this issue, but also gets the `post_title` from the stored post ID. So if the post title later changes, it'll output the new title.

### DIFF
--- a/php/blocks/controls/class-post.php
+++ b/php/blocks/controls/class-post.php
@@ -70,11 +70,11 @@ class Post extends Control_Abstract {
 	 * @return string|WP_Post|null $value The value to be made available or echoed on the front-end template.
 	 */
 	public function validate( $value, $echo ) {
+		$post = isset( $value['id'] ) ? get_post( $value['id'] ) : null;
 		if ( $echo ) {
-			return isset( $value['title'] ) ? $value['title'] : '';
+			return $post ? $post->post_title : '';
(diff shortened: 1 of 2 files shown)


kienstra:
**Request For Review**

Hi @lukecarbis and @RobStino, 
Could you please review this? Now,`block_field()` outputs the post title. And if the user later changes the title in the post's editor, this displays the new title.

Also, this shouldn't affect backwards-compatibility.

lukecarbis (approved):
Works! Great pickup.

kienstra:
Thanks, @lukecarbis!

PR #298, opened 2019-04-28
Repeater Control

### LINKED ISSUE #36: "Repeater" control type
😮

### LINKED ISSUE #391: Update "Learn More" links in New Block alert
The "Learn More" links should be updated for the new docs.

- Block Templates: https://getblocklab.com/docs/get-started/add-a-block-lab-block-to-your-website-content/
- Template Functions: https://getblocklab.com/docs/functions/

learn-more-links

lukecarbis:
Closes #36.
Closes #391.


lukecarbis:
This is still _VERY_ early – it only includes 1 thing: An admin UI for adding / managing Repeater fields.

The "Add Sub-Field" button doesn't even work yet.

But I wanted to get an early review before heading too far down this path to make sure we're happy with the UX. What are your thoughts?

first-pass-at-repeater

RobStino:
Hey @lukecarbis,
So excited to see this branch kicked off!!!!!
Here are the designs I mocked up for how I think this could look.
I'm not 100% committed to particular design bits and pieces, but I think it could be important to have the parent field persistently open and "wrapping" the sub fields. I recall we talked about borrowing from the (non-Customizer) menu builder for submenu items, but the above makes the sub-fields feel a bit disconnected from the parent field.
What do you think?
Block_lab_Repeater_Field

RobStino:
Here is a mock of what the frontend could look like as well: https://user-images.githubusercontent.com/1578148/56863483-05b50280-69fa-11e9-9e28-cad86f719af2.png

lukecarbis:
I actually disagree. I think the backend gets far too crowded like that. Since the headers are all the same, isn't it better to keep all the fields visible without clicking into the repeater?

lukecarbis:
Added a commit which enables new Sub Fields to be added. There are still a few known issues (e.g. background colours when dragging, dragging sub-rows doesn't work, the auto-scroll doesn't go to the right place), but I just wanted to give you a little more of an idea on how this could work.

kienstra:
**Looks Good**

Hi @lukecarbis,
Nice, this looks good:

adding-sub-field

RobStino:
Short screencap of initial thoughts on the WIP.
As discussed in today's team call.
https://d.pr/v/8N9Nrx

lukecarbis:
@kienstra I merged #221 into this PR, since I know that it's got some changes that affect the repeater field.

Progress so far: The UI for adding a repeater field _should_ now be done. I still haven't gotten around to saving it, and I need to do a general tidy up of the CSS. But those should be the last things to do for the block editor side of things (still lots to do in the editor editor).

@RobStino – could you please review the design / UX so far?

pcfreak30:
I think UI wise you may want to consider the possibility of a lot of repeater nesting. So I think if you add something, it may need to be a modal? Trying to add in 3-4 levels of repeaters would make the UI very cramped.

@lukecarbis

pcfreak30:
I think commit 5775c6590c78b2319ed4522841ab8646d928be2c should actually be removed TBH. That flexibility should be allowed I think.

lukecarbis:
Thanks for your thoughts @pcfreak30. Although I agree that having multiple depths of nesting should be the goal – I think that for this release we should stick to just one. Otherwise we might never ship this massive feature.

RobStino:
Hi @lukecarbis 
Great work on this so far.
I recall discussing moving the `location` field/column from the sub-fields table as it's not required.
Was this still the plan?

lukecarbis:
Nice pickup @Rob. Fixed in https://github.com/getblocklab/block-lab/pull/298/commits/b95782cb8fe7aaa69ab022e87d3930e943d3d0e3.

RobStino:
@lukecarbis I was just looking at this and noticed that the "Location" column has been removed from the parent table all together. It's still available as a field option, but isn't visible in the table view. Was this intentional?

lukecarbis:
Yes, this is intentional. It's due to the fact that some fields don't have a location setting (e.g. Rich Text, Repeater, likely more advanced fields in the future).

I think we need a UI that addresses this in a different way.

kienstra:
**Question About Template Function**

Hi @lukecarbis and @RobStino,
Hope you had a great weekend.

How about a template function `block_sub_field( $parent, $child, $echo )` to output sub-field values in the template? This would essentially be a wrapper for [block_field()](https://github.com/getblocklab/block-lab/blob/a48846506c74342454150c0bb7c04d5f700f4f2d/php/helpers.php#L20), but for sub-fields. 

For example, in a block template like `blocks/block-home-carousel.php`:

```php
<img src="<?php block_sub_field( 'home-carousel', 'image-1' ); ?>">
```

ACF [has iteration functions](https://www.advancedcustomfields.com/resources/get_sub_field/) like `have_rows()` and `the_row()`. 

We won't have multiple nested repeater fields, just a parent repeater and sub-fields. So I don't see much need for those, but I'm open to ideas.

Thanks!

kienstra:
Ah, maybe multiple nesting is planned, per https://github.com/getblocklab/block-lab/pull/298#issuecomment-497277092

>Although I agree that having multiple depths of nesting should be the goal – I think that for this release we should stick to just one. Otherwise we might never ship this massive feature.

kienstra:
If there will eventually be multiple nesting, maybe the template function could accept an `array()` of the nested repeaters:

```php
get_sub_field( array( 'home-carousel', 'image-1', 'alt-1' ) )
```

That's getting complex, though. 

I'd like to avoid [iterator functions](https://www.advancedcustomfields.com/resources/get_sub_field/) like `the_row()` and `have_rows()` that use and set the global state.

lukecarbis:
@kienstra I'm actually all for keeping things as simple as possible. I have trouble imagining use cases that rely on complicated systems with multiple nestings of repeaters. I say we focus on doing a single level of repeater nesting, and doing it as simply as possible.

kienstra:
Thanks, @lukecarbis! That sounds great to me.

RobStino:
Simple sounds good to me.
My only thoughts re this is that whether a decision to keep it simple locks us away from being able to expand it later on, if we chose to.
We may never do that, but keep it simple for now is a different decision to keeping it simple for ever. As this is an untried feature for our users, I don't know if we have the insights to make a forever decision.
If we can do simple now and keep it open to expanding in the future, that's ideal I think.

WebGardenGeek:
Just read through this thread and I really like what you guys came up with. It's great that the repeater is essentially a container for whatever subfields you want to add.

I have used a plugin called WCK (WordPress Creation Kit Pro) for custom post type and options page development and they do something very similar for metabox fields, which when it comes right down to it are fairly similar to blocks, except they're not drag and drop and they are assigned to a post type instead of just being a universal block that one can add to a post, page, or admin options page etc.

You simply indicate whether you want a repeater or not and then you add variables to the repeater. Then when you go to actually edit that metabox on the actual post/page it allows you to fill out the variables and add them to the array and you can do that as many times as you want (the repeater part).

Seems like things are coming along - let me know if you want a beta tester ;-)

kienstra:
**Repeater Field In Editor**

This will probably need better styling and more thought about the layout. But here's the working Repeater field in the editor.

If you had created a block with a Repeater before, you'll probably have to go back to `/wp-admin` > Block Lab > Edit Block and click 'Update.'

repeater-field

Still, this doesn't yet have template functions to output the values on the front-end, like `block_field()`.

js/blocks/controls/repeater.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+const { BaseControl } = wp.components;
+
+/**
+ * Internal dependendies
+ */
+import { getRenderedFields } from '../loader/edit';
+
+const BlockLabRepeaterControl = ( props, field, block ) => {
+	const subFields = field.sub_fields ? getRenderedFields( field.sub_fields, props, block ) : null;
```
kienstra: If it's alright, I changed `sub-fields` to `sub_fields`, to enable accessing it like `field.sub_fields`.

kienstra:
**Template function now works**

This is definitely open for discussion.

But https://github.com/getblocklab/block-lab/pull/298/commits/3c6b624117c1f7b0f30000e1b86b621022839d8d gives a way for `block_field()` to output repeater fields.

For example:
```php
block_field( 'home-carousel', 'image-1' )
```

The function still works like before.

Update: I'll probably have to reconsider this.

kienstra:
I still have quite a bit of work in this PR, but Repeater fields are now working.

kienstra:
**Should the Repeater be available in the Inspector?**

Hi @lukecarbis,
Do you think the Repeater should be available in the Inspector? It might be hard to have a 'nested' effect, [like here](https://github.com/getblocklab/block-lab/pull/298#issuecomment-511086383). 

There isn't much room in the Inspector.

Thanks, Luke! Hope you're doing great.

php/blocks/class-loader.php
```diff
@@ -168,23 +168,43 @@ public function get_block_attributes( $block ) {
...
+
+	/**
+	 * Sets the field values in the attributes, enabling them to appear in the block.
+	 *
+	 * @param array  $attributes The attributes in which to store the field value.
+	 * @param string $field_name The name of the field, like 'home-hero'.
+	 * @param Field  $field      The Field to set the attributes from.
+	 * @return array $attributes The attributes, with the new field value set.
+	 */
+	public function set_attributes_from_field( $attributes, $field_name, $field ) {
```
kienstra: This is only abstracted from `get_block_attributes()`, and not changed much.

kienstra:
**Request For Help With Block Editor Display**

Hi @lukecarbis and @RobStino,
Update: This probably doesn't need feedback until I implement the repeater rows in the block editor.

Could you please help with the layout and design of the Repeater field in the block editor?

Here's the current display:
current-repeater

Feel free to commit to this PR, like to [repeater.js](https://github.com/getblocklab/block-lab/pull/298/files#diff-4c5cb88a2e8c8ccdc511d166dd025c8eR15). Or to share thoughts or screenshots.

Thanks!

kienstra:
**Current Display**

repeater-here

* The repeater field now allows adding and removing rows.
* But it doesn't save its values in the block editor. I need to work on that.
* It probably needs styling and layout work.

kienstra:
**Question About Maximum Rows**

What do you think about having a field for 'Maximum Rows':

maximum-rows

This would be the maximum amount of rows that a user could create in the Repeater field.

Ideally, this wouldn't be needed. But it looks like we'll have to declare the repeater row fields as block attributes, even before they're added.

The Repeater control allows dynamically adding rows ([shown above](https://github.com/getblocklab/block-lab/pull/298#issuecomment-513629775)). Each of them has fields, which become attributes of the block.

But it looks like blocks don't allow dynamically adding attributes. The attributes must be declared once in [registerBlockType](https://github.com/getblocklab/block-lab/blob/a3ee65d05e46f80552bd49198743a299dee624ef/js/blocks/loader/index.js#L37).

So if there were 2 sub-fields in the repeater:

```js
example-image
example-text
```

...there would need to be attributes declared in [registerBlockType](https://github.com/getblocklab/block-lab/blob/a3ee65d05e46f80552bd49198743a299dee624ef/js/blocks/loader/index.js#L37), even if the user hasn't added any row yet:

```js
{
        example-image[0]: { type: 'string' },  // For the first row.
        example-text[0]: { type: 'string' },
        example-image[1]: { type: 'string' }, // For the second row.
        example-text[1]: { type: 'string' },
        example-image[2]: { type: 'string' }, // For the third row.
        example-text[2]: { type: 'string' },
        // etc...
}
```

There probably needs to be some limit to this, it probably shouldn't declare hundreds of rows.

**That's why maybe there should be a 'Maximum Rows' field.**

I'm looking at other workarounds for this.

RobStino:
Layout idea as covered in today's product call.
Concept for moving away from stacking rows.
image

lukecarbis:
Note: I found a bug where including a space in the sub-field name causes the repeater to break. e.g., try creating a repeater with the sub-field "Button Text". You can even just set the slug to `button`.

Edit: No, I think the bug is just that saving a block without editing the repeater causes it to break.

lukecarbis:
@kienstra Back over to you mate. In terms of next steps, could we move the "Next" and "Previous" buttons outside of the individual rows, and alongside the "Add new" button?

Screen Shot 2019-08-12 at 12 18 18 pm

js/blocks/controls/post.js
```diff
@@ -1,15 +1,7 @@
...
-	/**
-	 * Gets the post title from an API response.
-	 *
-	 * @param {Object} apiResponse The API response in which to look for the post title.
-	 * @return {String} The post title from the response, or the default.
-	 */
-	const getTitleFromAPI = apiResponse => ( apiResponse && apiResponse.title && apiResponse.title.rendered ) ? apiResponse.title.rendered : '';
-
-	return ContentControl( props, field, getTitleFromAPI );
+const BlockLabPostControl = ( props ) => {
```
kienstra, line 3: It looks like `getTitleFromAPI()` wasn't actually used in the `ContentControl`, or anywhere else. I'm not sure why I added it 😄

js/blocks/loader/edit.js
```diff
@@ -1,67 +0,0 @@
...
-	} )
-
-	return (
-		<div key={ block.name + "-fields" }>
-			{fields}
-		</div>
-	)
-}
-
-const editComponent = ( props, block ) => {
```
kienstra, line 37: This file is mainly refactored. For example, `editComponent` is moved to `components/edit.js`

js/blocks/components/advanced-controls.js
```diff
@@ -15,15 +24,15 @@ const inspectorControls = ( props, block ) => {
...
 
 	return (
-		<InspectorAdvancedControls key={"inspector-advanced-controls-" + block.name}>
+		<InspectorAdvancedControls key={ `inspector-advanced-controls-${ block.name }` }>
 			<DotTip tipId="block-lab/additional-css-class">
-				<p className="bl-dot-tip" dangerouslySetInnerHTML={{__html: tip}}></p>
+				<p className="bl-dot-tip" dangerouslySetInnerHTML={ { __html: tip } }></p>
 				<p className="bl-dot-tip read-more">
-					<a href="https://github.com/getblocklab/block-lab/wiki/7.-FAQ" target="_blank">{__( 'Read more', 'block-lab' )}</a>
+					<a href="https://github.com/getblocklab/block-lab/wiki/7.-FAQ" target="_blank">{ __( 'Read more', 'block-lab' ) }</a>
```
kienstra: @todo: update this link

js/blocks/components/fields.js
```diff
@@ -0,0 +1,82 @@
...
+		 * This function is passed to the control so that the control can save the value,
+		 * depending on whether the control is in a repeater row or not.
+		 *
+		 * @param {mixed} newValue The new control value.
+		 */
+		const onChange = ( newValue ) => {
+			const attr = { ...parentBlockProps.attributes };
+			const { setAttributes } = parentBlockProps;
+
+			if ( undefined !== rowIndex ) { // If this is in a repeater row.
```
kienstra: @todo: change order of if/else

js/blocks/components/fields.js
```diff
@@ -0,0 +1,82 @@
...
+		 * If this is in a repeater row, the value is appropriate for that.
+		 *
+		 * @param {Object} props The properties of the Control function.
+		 */
+		const getValue = ( props ) => {
+			const { field, parentBlockProps, rowIndex } = props;
+			const attr = { ...parentBlockProps.attributes };
+
+			// If the field has a parent, it's probably in a repeater row.
+			return ( field.parent && attr[ field.parent ][ rowIndex ] ) ? attr[ field.parent ][ rowIndex ][ field.name ] : attr[ field.name ];
```
kienstra: @todo: Make this a proper if statement.

kienstra (approved):

js/blocks/components/fields.js
kienstra: This is applied in https://github.com/getblocklab/block-lab/pull/298/commits/a0a0ac2f98801cbfb8be4d1ee80bd2270c8858be.

js/blocks/components/fields.js
kienstra: This is applied in https://github.com/getblocklab/block-lab/pull/298/commits/5abc5644bf97c1123e5864b30a7b6993ca332902

PR #306, opened 2019-05-19
Display a notice if a field isn't available anymore

### LINKED ISSUE #242: License Expiration
We should probably handle pro license expiration better. Right now, there's a PHP notice if a 'pro' control like 'user' isn't available:
53790835-799cc800-3eed-11e9-90bc-c65f9247839d

kienstra:
Update: This description is mainly outdated. Please see the comments below.
* This PR is the beginning of the conversation. The copy can change 😄 
* Right now, this says that maybe the pro license expired.
* But there should probably be a check that this is a pro
field.
new-notice

Right now, the link goes to:
link-goes-to

Fixes #242

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -543,7 +543,30 @@ class Block_Post extends Component_Abstract {
 				<code id="block-fields-name-code_<?php echo esc_attr( $uid ); ?>"><?php echo esc_html( $field->name ); ?></code>
 			</div>
 			<div class="block-fields-control" id="block-fields-control_<?php echo esc_attr( $uid ); ?>">
-				<?php echo esc_html( $this->controls[ $field->control ]->label ); ?>
+				<?php
+				if ( isset( $this->controls[ $field->control ] ) ) :
+					echo esc_html( $this->controls[ $field->control ]->label );
+				else :
+					?>
+					<div class="notice notice-info inline notice-alt">
+						<?php
+						/* translators: %1$s is the field type, %2$s is the URL for the Pro license */
+						printf(
+							wp_kses_post( 'This <code>%1$s</code> field type is not available. Maybe your <a href="%2$s">pro license</a> expired.', 'block-lab' ),
+							esc_html( $field->control ),
+							esc_url(
+								add_query_arg(
+									array(
+										'post_type' => 'block_lab',
+										'page'      => 'block-lab-pro',
+									),
+									admin_url( 'edit.php' )
+								)
+							)
+						);
+						?>
+					</div>
+				<?php endif; ?>
 			</div>
 			<div class="block-fields-location" id="block-fields-location_<?php echo esc_attr( $uid ); ?>">
 				<?php
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -34,6 +34,17 @@ class Block_Post extends Component_Abstract {
 	public $controls = array();
 
 	/**
+	 * The pro controls.
+	 *
+	 * @var array
+	 */
+	public $pro_controls = array(
+		'post',
+		'taxonomy',
+		'user',
+	);
+
+	/**
 	 * Register any hooks that this component needs.
 	 *
 	 * @return void
@@ -87,17 +98,22 @@ class Block_Post extends Component_Abstract {
 		);
 
 		if ( block_lab()->is_pro() ) {
-			$controls = array_merge(
-				$controls,
-				array(
-					'post'     => new Controls\Post(),
-					'taxonomy' => new Controls\Taxonomy(),
-					'user'     => new Controls\User(),
-				)
-			);
+			foreach ( $this->pro_controls as $pro_control ) {
+				$controls[ $pro_control ] = $this->instantiate_pro_control( $pro_control );
+			}
 		}
 
 		$this->controls = apply_filters( 'block_lab_controls', $controls );
+	}
+
+	/**
+	 * Gets an instantiated pro control.
+	 *
+	 * @return object The instantiated pro control.
+	 */
+	public function instantiate_pro_control( $control_name ) {
+		$control_class = 'Block_Lab\\Blocks\\Controls\\' . ucwords( $control_name );
+		return new $control_class();
 	}
 
 	/**
@@ -520,6 +536,8 @@ class Block_Post extends Component_Abstract {
 		if ( ! $uid ) {
 			$uid = '{{ data.uid }}';
 		}
+		$is_field_disabled = ( ! isset( $this->controls[ $field->control ] ) && in_array( $field->control, $this->pro_controls, true ) );
+
 		?>
 		<div class="block-fields-row" data-uid="<?php echo esc_attr( $uid ); ?>">
 			<div class="block-fields-sort">
@@ -544,7 +562,7 @@ class Block_Post extends Component_Abstract {
 			</div>
 			<div class="block-fields-control" id="block-fields-control_<?php echo esc_attr( $uid ); ?>">
 				<?php
-				if ( isset( $this->controls[ $field->control ] ) ) :
+				if ( ! $is_field_disabled ) :
 					echo esc_html( $this->controls[ $field->control ]->label );
 				else :
 					?>
@@ -602,7 +620,15 @@ class Block_Post extends Component_Abstract {
 								id="block-fields-edit-label-input_<?php echo esc_attr( $uid ); ?>"
 								class="regular-text"
 								value="<?php echo esc_attr( $field->label ); ?>"
-								data-sync="block-fields-label_<?php echo esc_attr( $uid ); ?>" />
+								data-sync="block-fields-label_<?php echo esc_attr( $uid ); ?>"
+								<?php echo $is_field_disabled ? 'readonly="readonly"' : ''; ?> />
+							<?php if ( $is_field_disabled ) : ?>
+								<input
+									name="block-is-disabled-pro-field[<?php echo esc_attr( $uid ); ?>]"
+									type="hidden"
+									value="true"
+								/>
+							<?php endif; ?>
 						</td>
 					</tr>
 					<tr class="block-fields-edit-name">
@@ -622,7 +648,8 @@ class Block_Post extends Component_Abstract {
 								id="block-fields-edit-name-input_<?php echo esc_attr( $uid ); ?>"
 								class="regular-text"
 								value="<?php echo esc_attr( $field->name ); ?>"
-								data-sync="block-fields-name-code_<?php echo esc_attr( $uid ); ?>" />
+								data-sync="block-fields-name-code_<?php echo esc_attr( $uid ); ?>"
+								<?php echo $is_field_disabled ? 'readonly="readonly"' : ''; ?> />
 						</td>
 					</tr>
 					<tr class="block-fields-edit-control">
@@ -636,8 +663,16 @@ class Block_Post extends Component_Abstract {
 							<select
 								name="block-fields-control[<?php echo esc_attr( $uid ); ?>]"
 								id="block-fields-edit-control-input_<?php echo esc_attr( $uid ); ?>"
-								data-sync="block-fields-control_<?php echo esc_attr( $uid ); ?>" >
-								<?php foreach ( $this->controls as $control ) : ?>
+								data-sync="block-fields-control_<?php echo esc_attr( $uid ); ?>"
+								<?php disabled( $is_field_disabled ); ?> >
+								<?php
+								$fields_for_select = $this->controls;
+								// If this field is disabled, it was probably added when there was a valid pro license, so still display it.
+								if ( $is_field_disabled && in_array( $field->control, $this->pro_controls, true ) ) {
+									$fields_for_select[ $field->control ] = $this->instantiate_pro_control( $field->control );
+								}
+								foreach ( $fields_for_select as $control ) :
+									?>
 									<option
 										value="<?php echo esc_attr( $control->name ); ?>"
 										<?php selected( $field->control, $control->name ); ?>>
@@ -658,7 +693,8 @@ class Block_Post extends Component_Abstract {
 							<select
 								name="block-fields-location[<?php echo esc_attr( $uid ); ?>]"
 								id="block-fields-edit-location-input_<?php echo esc_attr( $uid ); ?>"
-								data-sync="block-fields-location_<?php echo esc_attr( $uid ); ?>" >
+								data-sync="block-fields-location_<?php echo esc_attr( $uid ); ?>"
+								<?php disabled( $is_field_disabled ); ?> >
 									<option
 										value="editor"
 										<?php selected( $field->location, 'editor' ); ?>>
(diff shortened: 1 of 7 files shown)


kienstra:
This still allows editing a Pro field that's not available, but the `<select>` will default to a "Text" type:
prof-field


Maybe this is confusing. But it allows the user to change the field. For example, maybe they would change a Post field to a Text field with the post ID.

kienstra:
**Question About Copy And Link**

Hi @lukecarbis and @RobStino,
What do you think about the copy and the link location here?

>This post field type is not available. Maybe your pro license expired.

copy-and-link

The link goes to the "Go Pro" page:
link-goes-to

Thanks!

lukecarbis (changes requested):
@kienstra So glad you're onto this! Great PR.

Here's my suggestion on how the notice should look (using the warning dashicon). I'm not really happy with forcing a notice in that small space:
Screen Shot 2019-05-20 at 8 44 37 am

(@RobStino feel free to chime in on this)

Also note my copy change suggestion. This is a bit shorter and to the point:
> This {field type} field requires an active pro license.

I really like linking to the Pro page. Good call.

The other nice-to-have with this PR would be to _keep_ the settings for the disabled control, but have them disabled. This is so that the values won't be lost if you save the Block, and then re-activate your pro license. (Sure, it would be trivial to re-enable the fields using the Inspector and save changes to them that way, but the point is that the field won't show in the editor or the frontend)

RobStino:
UI and copy is great. 
Some thoughts. On the move so sorry about the brevity. 😄
With an inactive license, what _should_ we disable. Access to the Pro field in all instances, or just the ability to edit the field in the block editor?
What if a block built with pro sits on a site for many years. Is the site owner required to have an active pro license ongoing? Even if they never build a pro block again? 
I guess it comes down to what pro is. The ability to “build blocks with pro fields” or the ability to “build and use blocks with pro fields “
I don’t know if I have a strong opinion, so super interested in your thoughts.

kienstra:
Hi @RobStino,
Thanks, good questions.

>What if a block built with pro sits on a site for many years. Is the site owner required to have an active pro license ongoing? Even if they never build a pro block again?

Good point. The front-end output of Pro fields will be different when the license expires. I hadn't thought of that before.

This is because the pro fields usually use an `array()` for their value. For example, the Post control saves the value as an array, and [has a validate() method](https://github.com/getblocklab/block-lab/blob/eda671f134819f5dd87af271c4d6822c7364e160/php/blocks/controls/class-post.php#L130) that parses the `array()` for `block_value()` and `block_field()`. 

But when the pro version license expires, that `validate()` method doesn't run, and the output is unexpected:

block-field-array

kienstra:
@lukecarbis, good points. I'm working on applying them.

kienstra:
https://github.com/getblocklab/block-lab/pull/306/commits/2ca1e2571caa0e6a8042afc7ac5a4d02bc10f44b applies Luke's suggestion for the copy:

new-copy

lukecarbis:
It would be great to apply a little styling so that the icon sits left of the text, as per the screenshot I posted earlier.

kienstra:
>The other nice-to-have with this PR would be to keep the settings for the disabled control, but have them disabled. This is so that the values won't be lost if you save the Block, and then re-activate your pro license. (Sure, it would be trivial to re-enable the fields using the Inspector and save changes to them that way, but the point is that the field won't show in the editor or the frontend)

This is a good idea. I'm working on it now, though I haven't figured out a simple way to do this.

kienstra:
https://github.com/getblocklab/block-lab/pull/306/commits/e3d26a6a6d379c1f2335731f32a548236ba878af adds a margin to separate the dashicon and the message:

margin-here

kienstra:
>The other nice-to-have with this PR would be to keep the settings for the disabled control, but have them disabled. This is so that the values won't be lost if you save the Block, and then re-activate your pro license. (Sure, it would be trivial to re-enable the fields using the Inspector and save changes to them that way, but the point is that the field won't show in the editor or the frontend)

Hi @lukecarbis,
Good idea. The commits above make it possible to get the same values for pro fields when you reactivate an expired pro license:

Screenshot 1) For example, here's a Taxonomy field with the pro license activated:
taxonomy-field

Screenshot 2) When the pro license expires, here's what it'll look like:
taxonmy-here

And if the pro license is reactivated, it will looks like Screenshot 1 again.

Is this what you had in mind?

php/post-types/class-block-post.php
```diff
@@ -578,7 +620,15 @@ public function render_fields_meta_box_row( $field, $uid = false ) {
 								id="block-fields-edit-label-input_<?php echo esc_attr( $uid ); ?>"
 								class="regular-text"
 								value="<?php echo esc_attr( $field->label ); ?>"
-								data-sync="block-fields-label_<?php echo esc_attr( $uid ); ?>" />
+								data-sync="block-fields-label_<?php echo esc_attr( $uid ); ?>"
+								<?php echo $is_field_disabled ? 'readonly="readonly"' : ''; ?> />
+							<?php if ( $is_field_disabled ) : ?>
+								<input
+									name="block-is-disabled-pro-field[<?php echo esc_attr( $uid ); ?>]"
+									type="hidden"
```
kienstra, line 652: Having a hidden `<input>`to show that the pro field (control) is disabled isn't ideal.
 
But because I set the `<select>` for the control type as `disabled`, its value won't be present in the POST request. So I couldn't figure out how to find if it's a pro control, without adding this hidden `<input>`.

php/post-types/class-block-post.php
```diff
@@ -598,7 +648,8 @@ class="regular-text"
 								id="block-fields-edit-name-input_<?php echo esc_attr( $uid ); ?>"
 								class="regular-text"
 								value="<?php echo esc_attr( $field->name ); ?>"
-								data-sync="block-fields-name-code_<?php echo esc_attr( $uid ); ?>" />
+								data-sync="block-fields-name-code_<?php echo esc_attr( $uid ); ?>"
+								<?php echo $is_field_disabled ? 'readonly="readonly"' : ''; ?> />
```
kienstra, line 676: You're probably aware of this, but using `readonly` instead of `disabled` means that it will be in the `POST` request. Having the name of the control in the request is important. Otherwise, it's not saved.

RobStino:
Hey,
Testing this now. Looking good, but I did notice something odd.
Firstly, I did see that you can still add and interact as normal with a Pro field within a block when the Pro license has been deactivated. This is cool! It looks like this was already thought through and my comments further up weren't needed. :D
The odd thing I noticed was with the User field.
While the Pro license was active, block_field returned a string of the Username. So the output in the editor and on the frontend was just `Rob`.
With the license deactived it output `1,Rob` in both the editor and frontend. I'm guessing this is an array of the User ID and User Display Name.

*License Active*
image

*License Inactive*
image

kienstra:
Hi @RobStino,

Good catch. It's definitely not ideal to have that different front-end output when the User (or another pro field) is disabled.

I [see that also](https://github.com/getblocklab/block-lab/pull/306#issuecomment-493802778). It doesn't look to be caused by this PR. 

But I'll look at how we might fix it. If there isn't time before the release today, maybe we could consider that in a different PR.

kienstra:
There's a fix for this that's working locally. I'll push it today.

kienstra:
With https://github.com/getblocklab/block-lab/pull/306/commits/1c26b8ebf6ddccd8179614cdaeb68803e56f9bda, the front-end display of a pro field with an expired pro license will still be the same:

tax-field

kienstra:
As it is right now, there's no change to existing pro fields when a pro license expires:
* You can edit the field in the block editor. For example, you could change a Taxonomy field value from `foo` to `bar`, and it will save just like before.
* It [still displays](https://github.com/getblocklab/block-lab/pull/306#issuecomment-494129721) on the front-end the same.

But you can't change the block in the `/wp-admin` > Block Lab UI:
taxonmy-here

php/post-types/class-block-post.php
```diff
@@ -87,19 +98,25 @@ public function register_controls() {
...
-			$controls = array_merge(
-				$controls,
-				array(
-					'post'     => new Controls\Post(),
-					'taxonomy' => new Controls\Taxonomy(),
-					'user'     => new Controls\User(),
-				)
-			);
+			foreach ( $this->pro_controls as $pro_control ) {
+				$controls[ $pro_control ] = $this->instantiate_pro_control( $pro_control );
```
lukecarbis: Instead of having a (clever) method to `instantiate_pro_control()`, why not just set assign the `pro_controls` to an empty array when they're being defined, and the here do something like:

```
$this->pro_controls = array(
    'post' => new Controls\Post(),
    'taxonomy' => new Controls\Taxonomy(),
    'user' => new Controls\User(),
);
if ( block_lab()->is_pro() ) {
    $controls = array_merge( $controls, $this->pro_controls );
}
```

lukecarbis: Surely this classes `init` method is early enough?

php/post-types/class-block-post.php
```diff
@@ -578,7 +626,15 @@ public function render_fields_meta_box_row( $field, $uid = false ) {
 								id="block-fields-edit-label-input_<?php echo esc_attr( $uid ); ?>"
 								class="regular-text"
 								value="<?php echo esc_attr( $field->label ); ?>"
-								data-sync="block-fields-label_<?php echo esc_attr( $uid ); ?>" />
+								data-sync="block-fields-label_<?php echo esc_attr( $uid ); ?>"
+								<?php echo $is_field_disabled ? 'readonly="readonly"' : ''; ?> />
+							<?php if ( $is_field_disabled ) : ?>
```
lukecarbis, line 649: The indentation here is a little funky…

php/post-types/class-block-post.php
lukecarbis: Have you tested this out with checkboxes?

php/post-types/class-block-post.php
lukecarbis: Makes sense. Maybe we could add some CSS along the lines of:
```
.block-fields-edit input[readonly="readonly"],
.block-fields-edit textarea[readonly="readonly"] {
    color: #999;
}
```

lukecarbis (changes requested):
Looks great Ryan!

RobStino:
Tested and confirmed that values persist in the editor and the frontend whether Pro is active or not. Thanks Ryan.

php/post-types/class-block-post.php
kienstra: Yeah, that's an interesting idea. 

It would make the codebase a little cleaner. And it wouldn't add much overhead. 

But I thought that because most plugin users don't have the pro license, it'd be best to not instantiate the pro controls when they're not going to have access to them. 

`instantiate_pro_control()` is used in a few places to instantiate a pro control just-in-time.

kienstra:
Hi @lukecarbis, 
>Makes sense. Maybe we could add some CSS along the lines of...

Great idea, https://github.com/getblocklab/block-lab/pull/306/commits/11ff4711a7f1e154092900b69c54fc805e65e198 uses your CSS:

input-readonly

kienstra:
Hi @lukecarbis,
>Have you tested this out with checkboxes?

Like changing the `<input type="hidden">` [here](https://github.com/getblocklab/block-lab/pull/306#discussion_r285797750) into a checkbox?

php/post-types/class-block-post.php
kienstra: Sure, does [this look better](https://github.com/getblocklab/block-lab/pull/306/commits/11ff4711a7f1e154092900b69c54fc805e65e198#diff-018b57db3c29f17ff59f382617e39f14R631)?

lukecarbis:
@kienstra No, I mean checking if checkboxes submit the correct value if they are disabled (like selects). But I just realised that we're not loading all the additional options anyway, so it doesn't matter. Please ignore.

php/post-types/class-block-post.php
lukecarbis: Yes!

kienstra: Nice, thanks

php/post-types/class-block-post.php
kienstra: Still, your idea would make the codebase cleaner, a pretty strong consideration. My opinion above isn't strongly-held.

lukecarbis: Great thinking. I think the part I don't like is the part where we have to pass the class name as an argument (e.g. `instantiate_control( 'Taxonomy' )`. Feels like a hack to me, especially with the `ucwords`.

What if we:

- Change the method name to `instantiate_control()` since it's not really pro specific anyway
- Even simpler, `get_control()`?
- Check to see that the class _can_ be instantiated before we instantiate it?
- Check to see if the class _is already_ instantiated before we instantiate it?

I'm not sure how all that would fit together – do you think there's some improvements we can pull out from there?

kienstra: Sure, I'm working on this now.

php/post-types/class-block-post.php
```diff
@@ -81,40 +81,57 @@ public function register_hooks() {
...
-			'select'      => new Controls\Select(),
-			'multiselect' => new Controls\Multiselect(),
-			'toggle'      => new Controls\Toggle(),
-			'range'       => new Controls\Range(),
-			'checkbox'    => new Controls\Checkbox(),
-			'radio'       => new Controls\Radio(),
+		$control_names = array(
+			'text',
+			'textarea',
+			'url',
```
kienstra, line 87: To make this work with `ucwords()`, this control class [is renamed](https://github.com/getblocklab/block-lab/pull/306/commits/addc1afeefcb5c777638bd484b5a1e476faa0242#diff-032d8fb82d07a7fc45dc4fd064a36ca8R3) from `Url` to `URL`.

php/post-types/class-block-post.php
```diff
@@ -81,40 +81,57 @@ public function register_hooks() {
...
-			'number'      => new Controls\Number(),
-			'color'       => new Controls\Color(),
-			'image'       => new Controls\Image(),
-			'select'      => new Controls\Select(),
-			'multiselect' => new Controls\Multiselect(),
-			'toggle'      => new Controls\Toggle(),
-			'range'       => new Controls\Range(),
-			'checkbox'    => new Controls\Checkbox(),
-			'radio'       => new Controls\Radio(),
+		$control_names = array(
```
kienstra, line 84: This could have been a property, but I don't think it's needed outside this function. Also, it might add to confusion, as there are `$controls` and `$pro_controls` properties.

php/post-types/class-block-post.php
kienstra: @lukecarbis, how does https://github.com/getblocklab/block-lab/commit/addc1afeefcb5c777638bd484b5a1e476faa0242 look? It'll require some regression testing to make sure the URL control still works, as this renames the class from `URL` to `Url`.

kienstra:
**Basic Regression Testing**

The Url control looks to still work the same, before and after this PR:

# Before

before-pr-306

# After
after-pr-306

php/post-types/class-block-post.php
```diff
@@ -81,40 +81,57 @@ public function register_hooks() {
...
 	/**
-	 * Gets an instantiated pro control.
+	 * Gets an instantiated control.
 	 *
 	 * @param string $control_name The name of the pro control.
-	 * @return object The instantiated pro control.
+	 * @return object|null The instantiated pro control.
 	 */
-	public function instantiate_pro_control( $control_name ) {
+	public function get_control( $control_name ) {
```
lukecarbis, line 126: I like this method a lot better now.

lukecarbis (approved):
Great work!

php/post-types/class-block-post.php
kienstra: Thanks!

PR #309, opened 2019-05-21
Fix className returning false instead of an empty string

### LINKED ISSUE #307: block_field( 'className' ) outputs `No` when Custom Class is empty
Folks,
Just testing out the `className` solution. Sorry I didn’t get to it the other day.
I did notice something.
If no custom class is set, `block_field` outputs `No`. I guess there is a small chance that someone would have an actual class called `No`.

lukecarbis:
Closes #307.

### DIFF
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -37,6 +37,11 @@ function block_field( $name, $echo = true ) {
 		$value = $block_lab_attributes[ $name ];
 	}
 
+	// Cast default Editor attributes appropriately.
+	if ( 'className' === $name ) {
+		$value = strval( $value );
+	}
+
 	// Cast block value as correct type.
 	if ( isset( $block_lab_config['fields'][ $name ]['type'] ) ) {
 		switch ( $block_lab_config['fields'][ $name ]['type'] ) {


RobStino (approved):
Tested and looks good.
Return is now an empty string if no Custom Class is set.
Good work.

RobStino:
@kienstra You ok to code review and merge?

kienstra:
Hi @RobStino, I'm looking at this now, but will get back to you soon.

kienstra:
**OK To Merge**

Hi @lukecarbis and @RobStino,
Sounds good, feel free to merge when you'd like.

I couldn't reproduce the issue, but it looks like you could.

Locally, on the `develop` branch, when the `className` was empty:

classname-empty

...calling `block_field( 'className' )` outputs an empty string, and `$value` was also an empty string in that function:

block-field-call

But maybe I missed a step.

kienstra (approved):

kienstra:
Ah, maybe `block_field( 'className' )` output `'No'` before if no value had ever been set for that. Like if it was `undefined`.

kienstra:
I just reproduced the issue in #307. It appears if the `className` has never had a value:

classname-empty

PR #310, opened 2019-05-21
Ensure that the $value is a string before passing it to array_key_exists()

kienstra:
Steps to reproduce:

1. `git checkout develop && git checkout 75fefc3  # checkout the last version, 1.3.0`

2. Create a block with a Textarea field. It probably doesn't matter if you populate any of the inputs, like 'Placeholder Text.' Click 'Publish':

textarea-here

3. `git checkout 304e3834f2aedbfb8b7276567ce31d82dd3f520f # Checkout the latest from develop`

4. Go back to the UI from step 2 and click 'Update.' You probably don't have to change anything, just click 'Update.'

5. Expected: The Textarea field appears

6. Actual: There's a PHP warning:

```
Warning: array_key_exists(): The first argument should be either a string or an integer in /srv/www/env/public_html/wp-content/plugins/block-lab/php/blocks/controls/class-textarea.php on line 143
```
textarea-warning

### DIFF
--- a/php/blocks/controls/class-textarea.php
+++ b/php/blocks/controls/class-textarea.php
@@ -140,7 +140,7 @@ class Textarea extends Control_Abstract {
 	 * @return string|null The sanitized rest_base of the post type, or null.
 	 */
 	public function sanitize_new_line_format( $value ) {
-		if ( array_key_exists( $value, $this->get_new_line_formats() ) ) {
+		if ( is_string( $value ) && array_key_exists( $value, $this->get_new_line_formats() ) ) {
 			return $value;
 		}
 		return null;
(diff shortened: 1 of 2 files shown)


kienstra:
**Request For Review**

Hi @lukecarbis and @RobStino,
Could you please review this PR for an edge case?

lukecarbis (approved):
Nice pickup. I tested this and could reproduce the error. Using `is_string()` is a great solution, since it means that existing content won't automatically get paragraph tags attached to it (which could be unexpected) until the Block has been updated.

kienstra:
Thanks, @lukecarbis!

PR #321, opened 2019-05-28
Messaging & Images Update

RobStino:
Update the copy and images for Block Lab across the various .org and WP Admin screens. This includes logo and visual brand updates as well as improvements to messaging.


README.md
```diff
@@ -9,18 +9,18 @@ Stable tag: trunk
 License: GPLv2 or later
 License URI: http://www.gnu.org/licenses/gpl
 
-A WordPress Admin interface and a simple templating system for building custom Gutenberg blocks.
+The easy way to build custom blocks for Gutenberg.
```
lukecarbis, line 12: Happy to keep the word "Gutenberg" here. Just thought I'd start the conversation, though. We've not really used the word "Gutenberg" anywhere else in the plugin. I guess the reason is that the average WordPress user might not know what that means.

On the other hand, maybe it's reasonable to expect our target audience to know what Gutenberg is. We've been referring to it as "the WordPress Editor" elsewhere.

lukecarbis: Same in the description. We can always update the description at any time, but it's worth keeping in mind that Gutenberg may not be "the new WordPress editor" for everyone, and increasingly will be the only WordPress editor people remember.

RobStino: I know what you mean. I've been using the term "The WordPress Editor" more and more and then using Gutenberg as a term of clarification.
I did use Gutenberg here though as I feel it's still a strong search term. For people searching etc, they'll probably be using it.
Those are assumptions though, but my gut feeling is keep GB for now, and then review in 6 months or so.

lukecarbis (dismissed):

lukecarbis (approved):

PR #327, opened 2019-06-05
Move the upgrade to settings redirect out of the upgrade class

### LINKED ISSUE #245: No access error on "Go Pro" page once license is activated
I noticed today that when you activate your Pro license via the "Go Pro" page in the WP Admin, once it's activated you get a "You can't access this page" style error. I think this is because the page disappears once the license is activated, but it doesn't redirect elsewhere and instead tries to reload the page.
A simple solution for this would be to redirect to the Settings page after license activation.

lukecarbis:
Closes #245.

### DIFF
--- a/php/admin/class-admin.php
+++ b/php/admin/class-admin.php
@@ -59,10 +59,7 @@ class Admin extends Component_Abstract {
 			$this->upgrade = new Upgrade();
 			block_lab()->register_component( $this->upgrade );
 		} else {
-			$page = filter_input( INPUT_GET, 'page', FILTER_SANITIZE_STRING );
-			if ( 'block-lab-pro' === $page ) {
-				$this->settings_redirect();
-			}
+			$this->maybe_settings_redirect();
 		}
 
 		if ( block_lab()->is_pro() ) {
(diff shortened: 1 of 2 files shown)


php/admin/class-admin.php
```diff
@@ -58,6 +58,11 @@ public function init() {
 		if ( $show_pro_nag && ! block_lab()->is_pro() ) {
 			$this->upgrade = new Upgrade();
 			block_lab()->register_component( $this->upgrade );
+		} else {
+			$page = filter_input( INPUT_GET, 'page', FILTER_SANITIZE_STRING );
+			if ( 'block-lab-pro' === $page ) {
+				$this->settings_redirect();
+			}
```
kienstra: Maybe all of this `else` block could be inside `settings_redirect`, as it looks like that method is only used once. If you agree, maybe another name would be good, like `maybe_settings_redirect()`.

```php
} else {
        $this->maybe_settings_redirect();
}
```

kienstra (dismissed):
**Approved**

Hi @lukecarbis,
This looks good.

I couldn't reproduce this without this PR. But it seems to work fine with or without this PR.

mid-tier-mobile

php/admin/class-admin.php
lukecarbis: Great idea. Done in https://github.com/getblocklab/block-lab/pull/327/commits/383bfe4a299e91cb53cbef85c654ff7e6e2ace93.

lukecarbis:
@kienstra I could reliably reproduce the error by removing all block lab transients before pressing the Activate button.

kienstra (approved):
Thanks, @lukecarbis! The latest commit looks good. 

I'll work on reproducing this, but in the meantime, I think it's good to merge.

kienstra:
Hm, I couldn't reproduce this after running `wp transient delete block_lab_license`. But that's alright 😄

PR #328, opened 2019-06-05
Correct the wording for license details

### LINKED ISSUE #308: License information show as renewing even when not
Currently Block Lab shows a message indicating that licenses are set to auto-renew, even when this is not the case (which it currently isn't for any Early Access Pass customer).

Screen Shot 2019-05-21 at 8 45 34 am

lukecarbis:
Closes #308.

Currently licenses do not auto-renew.

In fact, there's no way to tell whether or not the license _will_ auto-renew. See the [EDD Software Licensing API](https://docs.easydigitaldownloads.com/article/384-software-licensing-api).

So I simply changed the copy to "Will expire" instead of "Will auto-renew".

### DIFF
--- a/php/views/license.php
+++ b/php/views/license.php
@@ -59,7 +59,7 @@
 							'<p>%1$s %2$s</p>',
 							sprintf(
 								// translators: A date.
-								__( 'Your license will be automatically renewed on %1$s.', 'block-lab' ),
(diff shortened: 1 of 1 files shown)


kienstra (approved):
**Approved**

Hi @lukecarbis,
This looks good! 

license-expires

PR #329, opened 2019-06-05
Show category in list table

### LINKED ISSUE #324: Show 'category' in Content Block manager admin screen
Would be helpful to see the Category in the Content Block manager screen.  Even better if this field was sortable.  

Very crude mock-up attached.
mock-up

lukecarbis:
Closes #324.

Screen Shot 2019-06-05 at 10 14 41 pm

This PR has includes minor refactoring of how we are populating the category select in the block edit screen. It won't look or work any differently, but please test that too.

Screen Shot 2019-06-05 at 10 14 04 pm

### DIFF
--- a/js/admin.block-post.js
+++ b/js/admin.block-post.js
@@ -12,7 +12,6 @@
 
 	$(function() {
 		blockTitleInit();
-		blockCategoryInit();
 		blockIconInit();
 		blockFieldInit();
 
@@ -162,29 +161,6 @@
 					autoSlug = false;
 				}
 			});
-		}
-	};
-
-	let blockCategoryInit = function() {
-		let categories       = wp.blocks.getCategories(),
-			categoriesLength = categories.length,
-			category         = $( '#block-properties-category-saved' );
-
-		for (let i = 0; i < categoriesLength; i++) {
-			if ( 'reusable' === categories[i].slug ) {
-				continue;
-			}
-			$( '<option/>', {
-				value: categories[i].slug,
-				text: categories[i].title,
-			} ).appendTo( '#block-properties-category' );
-		}
-
-		if ( category.val() !== '' ) {
-			let option = $( '#block-properties-category option[value="' + category.val() + '"]' );
-			if ( option.length > 0 ) {
-				$( '#block-properties-category' ).prop( 'selectedIndex', option.index() );
-			}
 		}
 	};
 
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -1153,10 +1153,8 @@ class Block_Post extends Component_Abstract {
 			echo esc_html( implode( ', ', $block->keywords ) );
 		}
 		if ( 'category' === $column ) {
-			global $post;
-
 			$block      = new Block( $post_id );
-			$categories = get_block_categories( $post );
+			$categories = get_block_categories( get_post() );
 			$categories = wp_list_pluck( $categories, 'title', 'slug' );
 
 			echo esc_html( $categories[ $block->category ] );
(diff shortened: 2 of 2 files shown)


js/admin.block-post.js
```diff
@@ -165,29 +164,6 @@
...
-		}
-
-		if ( category.val() !== '' ) {
-			let option = $( '#block-properties-category option[value="' + category.val() + '"]' );
-			if ( option.length > 0 ) {
-				$( '#block-properties-category' ).prop( 'selectedIndex', option.index() );
-			}
-		}
-	};
-
```
kienstra, line 190: Good idea to change this from JS to PHP.

php/post-types/class-block-post.php
```diff
@@ -1142,9 +1152,14 @@ public function list_table_content( $column, $post_id ) {
...
 			echo esc_html( implode( ', ', $block->keywords ) );
 		}
-		if ( 'fields' === $column ) {
-			$block = new Block( $post_id );
-			echo esc_html( count( $block->fields ) );
+		if ( 'category' === $column ) {
+			global $post;
+
+			$block      = new Block( $post_id );
+			$categories = get_block_categories( $post );
```
kienstra: Would it be possible to pass `$post_id`  or `get_post( $post_id )` instead of `$post` to `get_block_categories()`? This seems to work the same locally.
 
This might prevent using a `global $post`

kienstra (dismissed):
**Approved**

Hi @lukecarbis,
Nice change, this looks good.

There's just a minor question below, but feel free to merge when you're ready.

categories-in-list-table

php/post-types/class-block-post.php
lukecarbis: Yes, it is possible. In fact, passing `false` works also. In many ways, this is better, as it means that the list of block categories won't be filtered by third-parties.

`get_block_categories()` does expect a `WP_Post` object according to its docblock, but that parameter is only used as an additional argument in the `block_categories` filter. See [wp-admin/includes/post.php:2173](https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/post.php#L2173).

I very nearly just passed `false` instead of the global, but I decided not to because I felt it was a bit hacky. What do you think? Personally, I'd prefer to pass `false` if lean that way.

The thing that swayed my decision in the end was core's implementation here: [wp-admin/edit-form-blocks.php:90](https://github.com/WordPress/WordPress/blob/master/wp-admin/edit-form-blocks.php#L90)

kienstra: Hi @lukecarbis,
Good point, like you mentioned, [that Core implementation](https://github.com/WordPress/WordPress/blob/master/wp-admin/edit-form-blocks.php#L90) passes the global `$post` to `get_block_categories()`.

The way this PR uses the global `$post` isn't wrong, but I've usually seen globals declared at the top of the function. 

Maybe passing `get_post()` to `get_block_categories()` would be a little cleaner, but it'd probably work the same.

kienstra: >I very nearly just passed false instead of the global, but I decided not to because I felt it was a bit hacky. What do you think?

Yeah, it will probably work the same. 

But that probably wouldn't be my vote, as it violates the `@param WP_Post $post` tag for [get_block_categories](https://github.com/WordPress/WordPress/blob/c77e771c849a85b63cef1b9457ad1a501a331afa/wp-admin/includes/post.php#L2131), and the `@param WP_Post $post` tag for the [block categories filter](https://github.com/WordPress/WordPress/blob/c77e771c849a85b63cef1b9457ad1a501a331afa/wp-admin/includes/post.php#L2173).

lukecarbis: Great idea. Done in https://github.com/getblocklab/block-lab/pull/329/commits/be4a5b712215a29ddd63ce6850d31656f1b18b33.

I'd still be open to passing `false` or even `null` as an argument, if you think that would be better.

kienstra (approved):

php/post-types/class-block-post.php
kienstra: Haha, our comments crossed paths.

Thanks, that commit looks good.

PR #337, opened 2019-06-14
Don't auto-slug fields that already have a slug defined

### LINKED ISSUE #335: Auto-generate the slug only when the title field is filled out for the first time
I think this is a bug, as the problem and solution are described in this issue, which is marked as resolved: https://github.com/getblocklab/block-lab/issues/134

Steps to reproduce:

1. Add a new Field to a block
2. Edit the Field Title, and the Field Slug is automatically generated
3. Modify the Field Slug
4. Edit the Field Title again

Expected functionality: Field Title should update without changing the Field Slug again
Issue: Field Title updates, but overwrites the Field Slug

See video attached.
recording

lukecarbis:
Closes #335.

### DIFF
--- a/js/admin.block-post.js
+++ b/js/admin.block-post.js
@@ -124,12 +124,21 @@
 				fetchFieldSettings( fieldRow, $( this ).val() );
 			})
 			.on( 'change keyup', '.block-fields-edit-label input', function() {
-				let slug = slugify( $( this ).val() );
+				let slug = $( this )
+					.closest( '.block-fields-edit' )
+					.find( '.block-fields-edit-name input' );
+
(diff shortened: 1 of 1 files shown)


kienstra:
**Approved**

Hi @lukecarbis,
Nice work! This worked as expected:

1. On creating a new field, editing the 'Field Label' still updates (auto-slugs) the 'Field Name' 
2. Once the 'Field Label' is blurred, like by clicking somewhere else, it no longer 'auto-slugs'
3. You can still edit the 'Field Name' (slug) manually

auto-slug

kienstra (approved):

PR #340, opened 2019-06-16
Add a hook in the render_callback to enable enqueuing scripts

kienstra:
As suggested by Luke and Joachim, this will enable enqueuing custom JS for a block.

To see this in action:
1. Create a block, like with a slug of `'test-image'`. It probably doesn't matter if it has any fields or a template.
2. Add this to another plugin, and make sure that the 5th argument to `wp_enqueue_script()`, `$in_footer`, is `true`:

```php
/**
 * Enqueues a script in the test-image block.
 *
 * @param array $block The block that is rendered.
 */
add_action( 'block_lab_render_block_template', function( $block ) {
	if ( 'test-image' === $block['name'] ) {
		wp_enqueue_script( 'example-script', 'https://example.org/fake-script.js', array(), '0.1', true );
	}
} );
```

3. Add this block to a post, and inspect the source of the front-end
4. Expected: the script should appear:

script-on-the-front-en

5. On any URL without that block, it shouldn't appear


kienstra:
**Request For Testing**

Hi @lukecarbis,
Hope you're doing great. Could you please review this?

This is based on your idea to add a hook for scripts.

php/blocks/class-loader.php
```diff
@@ -250,6 +250,18 @@ public function render_block_template( $block, $attributes ) {
...
+		 * Runs in the 'render_callback' of the block.
+		 *
+		 * For example, if a block depends on a JavaScript file,
+		 * this is a good place to call wp_enqueue_script().
+		 * In that case, pass true as the 5th argument ($in_footer) to wp_enqueue_script().
+		 *
+		 * @param array $block The block that is rendered.
+		 * @param array $attributes The block attributes.
+		 */
+		do_action( 'block_lab_render_block_template', $block, $attributes );
```
lukecarbis: Do you think we should instantiate a `Block` instance here? Maybe at the top of the `render_block_template()` method? Would it be a cleaner way of passing the Block around, instead of an associative array?

It might be outside the scope of this PR, but it would be a backward compatibility break if we decided to do that _after_ releasing this action hook.

lukecarbis: Should we also create a second action specifically for the current block? e.g.

```
do_action( "block_lab_render_block_template_{$block['name']}", $block, $attributes );
```

lukecarbis (dismissed):
Tested, and everything works as expected – the JS is only loaded on pages that include the block.

I've left a couple of questions, but if you don't think they warrant changes, you can merge.

I actually have one more question, though:

I noticed that JS is also being loaded in the editor (even if the block isn't used, since we load all blocks in the editor). Should we provide a way of only loading JS on the frontend? Or should we leave that to the hooked function? e.g.:

```
add_action( 'block_lab_render_block_template', function( $block ) {
	if (
            'simple-text-block' === $block['name'] &&
            ! is_admin() &&
            ! defined('REST_REQUEST') &&
            ! wp_doing_ajax()
        ) {
		wp_enqueue_script( 'example-script', 'https://example.org/fake-script.js', array(), '0.1', true );
	}
} );
```

php/blocks/class-loader.php
kienstra: Good question. It might depend on if scripts will need to be enqueued for several blocks. 

If a script is only needed for one block, that dynamic action above would definitely be simpler.

kienstra: Yeah, it could help to have a `Block` instance. I don't see a good way to get the block's `ID` that's needed in the [Block constructor](https://github.com/getblocklab/block-lab/blob/c7bce5f2e29e786d4e832192c7cde8e265fd9a0d/php/blocks/class-block.php#L66), though.

This action definitely needs better documentation on how to get the block name, though. Otherwise, they'd have to look through the code a lot.

What do you think about https://github.com/getblocklab/block-lab/pull/340/commits/ea98b4bf08c52db6cf9397776d8746d238551f54?

kienstra:
>I noticed that JS is also being loaded in the editor (even if the block isn't used, since we load all blocks in the editor). Should we provide a way of only loading JS on the frontend? Or should we leave that to the hooked function? e.g.:

Good question. For most cases, the JS probably will only be needed on the front-end. But some blocks block might need JS in the editor, like a map block.

The approach on this PR so far will cover the most use cases, but will usually require a lot of code.

Like you mentioned, it'll usually require wrapping `wp_enqueue_script()` in the conditional with `is_admin() &&  ! defined('REST_REQUEST')` etc...

php/blocks/class-loader.php
lukecarbis: I think I'm suggesting we do both.

kienstra: Ah, OK.

kienstra:
**Maybe not add a hook at all?**

Hi @lukecarbis,
I should have checked earlier, but it looks like Core already has filters in the `render_callback`, like:

https://github.com/WordPress/wordpress-develop/blob/22d530e2bb3fb1b28929a454173050f820724880/src/wp-includes/blocks.php#L267-L275

These have access to most of the important data that the hook in this PR does, like the block name:

block-name-is-available

Of course, it'd be best if this were an action, not a filter. It's a little hacky using a filter when it's not filtering anything.

But this PR doesn't add much, other than being an action instead of a filter.

lukecarbis:
@kienstra I think it could be worth adding our own action hooks, especially if we can make them "front end only" hooks. Simply due to the fact that it would be easier for Block Lab users to use that way. That said, I'm a little on the fence.

If we had a clearly documented way of adding JS to a block that only used core filters, and you thought that was a better route, then I'd be happy to close this PR and write an FAQ on the topic instead.

kienstra:
Hi @lukecarbis,
>@kienstra I think it could be worth adding our own action hooks, especially if we can make them "front end only" hooks. Simply due to the fact that it would be easier for Block Lab users to use that way. That said, I'm a little on the fence.

Good point. The fact that there's also a [Core filter](https://github.com/WordPress/wordpress-develop/blob/22d530e2bb3fb1b28929a454173050f820724880/src/wp-includes/blocks.php#L267-L275) for the `render_callback` means we can make ours a little more specific, and only for the front-end.

How about https://github.com/getblocklab/block-lab/pull/340/commits/76bf375addd6e81d51de6d872bb9ba86627645e2?

I still need to address your point about a dynamic hook, based on the `$block['name'])`.

php/blocks/class-loader.php
```diff
@@ -250,6 +250,22 @@ public function render_block_template( $block, $attributes ) {
 		$block_lab_attributes = $attributes;
 		$block_lab_config     = $block;
 
+		if ( ! is_admin() && ! defined( 'REST_REQUEST' ) && ! wp_doing_ajax() ) {
```
lukecarbis: If `REST_REQUEST` was set, but `false`, this conditional would return an unexpected result. Unfortunately, I think we need something more like:

```
if ( ! is_admin() && ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST ) && ! wp_doing_ajax() )
```

kienstra: Ah, good point.

lukecarbis:
https://github.com/getblocklab/block-lab/commit/76bf375addd6e81d51de6d872bb9ba86627645e2 looks good – left one minor comment there.

php/blocks/class-loader.php
kienstra: Sounds good, applied in https://github.com/getblocklab/block-lab/pull/340/commits/730a67cf7d00f34d3bf9d093ffab057284bd6824

kienstra:
Hi @lukecarbis, thanks for your help here, I think this PR is a lot better with your feedback.

What do you think about the latest changes? I'll probably be offline for almost a day with the flight, so feel free to push to this and/or merge it.

lukecarbis:
Looks great! What do you think about turning the $block array into a Block object? It would require a new method in the Block class (e.g. `from_array()`), but would be slightly cleaner. It would have a backwards compatibility if we chose to do this later, since we're providing the array to the action hook.

lukecarbis:
I'd be happy to do the dev work here if you thought it was a good idea.

lukecarbis:
Also, it would match nicely with the work done in the repeater branch: https://github.com/getblocklab/block-lab/pull/298/files#diff-313c59455c007aac47ea3e7de92b6809

kienstra:
**Block Instance**

Hi @lukecarbis,
That's interesting creating a `Block` instance for the hook. Were you thinking it'd be easier for users to access values, like `$block->name` instead of `$block['name']`?

It looks like the existing `$block` in the hook and an instance of `Block` would have mainly the same data, though a `Block` instance would have `Field` instances instead of field arrays.

So I'm not sure if it'd be worth the overhead of instantiating a `Block`. 

Though like you mentioned, maybe it would go better with the [Repeater control](https://github.com/getblocklab/block-lab/pull/298/files#diff-313c59455c007aac47ea3e7de92b6809). Sorry, I haven't looked at that enough to say.

lukecarbis:
@kienstra Exactly. There would be no need to do things like `if ( isset( $block['name'] )`, and it would also harden our code against malformed post data. Same goes for `get_block_attributes`. I might try to put something together, borrowing from the repeater PR, and see how that looks in a separate PR.

If you like that idea, we could merge that one first, and then merge this.

lukecarbis:
@kienstra What do you think of #342?

kienstra:
>@kienstra What do you think of #342?

Sorry for the delay. I'll review #342.

lukecarbis (approved):
@kienstra Merged in our work on #342 for you. I'm happy with this PR, if you are. Please review https://github.com/getblocklab/block-lab/pull/340/commits/6892f6f312c48cfece84ee22a9b2e679ac8d663a first, though.

kienstra:
Thanks, @lukecarbis! I'll review this soon.

kienstra:
**Looks Good**

Hi @lukecarbis,
Thanks for https://github.com/getblocklab/block-lab/pull/340/commits/6892f6f312c48cfece84ee22a9b2e679ac8d663a, that looks good.

block-is-block-instance

kienstra:
**Merging**

Hi @lukecarbis,
If it's alright, I'm going to merge this. Thanks for your work on this PR and #342.

lukecarbis:
@kienstra just noticed in your screenshot that the `@param` tag in the docblocks introduced here still uses array instead of Block. Think you could update that in a new PR?

kienstra:
Sure, @lukecarbis!

PR #341, opened 2019-06-17
🐛Make sure zero can be selected in Range Control due to boolean bug.

### LINKED ISSUE #333: Range field type does not allow "0" as a value
This seems to be a bug.  I have a Range field type, which should allow a range of 0 to 100.  However editors are unable to enter the value "0".

When I allow the range of -1 to 100, editors are still unable to use the value "0" (although can select "-1").

ronalfy:
Fixes https://github.com/getblocklab/block-lab/issues/333

Boolean issue causes range control to never be set at zero.

### DIFF
--- a/js/blocks/controls/range.js
+++ b/js/blocks/controls/range.js
@@ -9,10 +9,12 @@ const BlockLabRangeControl = ( props, field, block ) => {
 			afterIcon={field.afterIcon}
 			label={field.label}
 			help={field.help}
-			value={attr[ field.name ] || field.default}
+			value={ ( attr[ field.name ] || '0' == attr[ field.name ] ) ? attr[ field.name ] : field.default }
 			onChange={rangeControl => {
-				attr[ field.name ] = rangeControl
-				setAttributes( attr )
+				if ( undefined !== rangeControl ) {
+					attr[ field.name ] = rangeControl
+					setAttributes( attr )
+				}
 			}}
 			min={field.min}
 			max={field.max}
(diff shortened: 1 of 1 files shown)


js/blocks/controls/range.js
```diff
@@ -9,10 +9,12 @@ const BlockLabRangeControl = ( props, field, block ) => {
 			afterIcon={field.afterIcon}
 			label={field.label}
 			help={field.help}
-			value={attr[ field.name ] || field.default}
+			value={ ( attr[ field.name ] || '0' == attr[ field.name ] ) ? attr[ field.name ] : field.default }
 			onChange={rangeControl => {
-				attr[ field.name ] = rangeControl
-				setAttributes( attr )
+				if ( undefined !== rangeControl ) {
```
kienstra: What do you think about removing this conditional?
```js
if ( undefined !== rangeControl ) {
```
With that conditional, it looks like it's not possible to:
1. highlight and delete the value
2. backspace when there's only 1 digit

values-here-and-here

If that conditional is removed and the user sets the value to `undefined`, it looks like it still displays the default value on the front-end, when calling `block_field( 'example-range' );`

kienstra:
**Looks Good**
*One Minor Point*

Hi @ronaldhuereca,
Thanks, this looks good. And it fixes the issue of not being able to select `0` in the Range control:

here-is-0

There's just one question below.

kienstra (changes requested):

js/blocks/controls/range.js
ronalfy: This should be addressed in https://github.com/getblocklab/block-lab/pull/341/commits/d4baa80afd17569bd11f1a5d93def8714718e309 

Thanks for the feedback.

kienstra: Nice, https://github.com/getblocklab/block-lab/commit/d4baa80afd17569bd11f1a5d93def8714718e309 looks good.

kienstra (approved):

PR #342, opened 2019-06-17
Use Block objects instead of config arrays

lukecarbis:


php/helpers.php
```diff
@@ -132,7 +131,7 @@ function block_value( $name ) {
  */
 function block_config() {
 	global $block_lab_config;
-	return $block_lab_config;
+	return (array) $block_lab_config;
```
lukecarbis, line 134: Returning as an array for backwards compatibility.

php/blocks/class-loader.php
```diff
@@ -191,22 +186,6 @@ public function get_block_attributes( $block ) {
...
-				$attributes[ $field_name ]['meta'] = $field['meta'];
-			}
-
-			if ( ! empty( $field['selector'] ) ) {
-				$attributes[ $field_name ]['selector'] = $field['selector'];
-			}
-
-			if ( ! empty( $field['query'] ) ) {
-				$attributes[ $field_name ]['query'] = $field['query'];
-			}
```
lukecarbis, line 209: These fields are literally never set, and don't actually do anything. Not sure how they ended up here.

kienstra: Yeah, interesting.

kienstra:
Hi @lukecarbis,
This looks good enough to proceed. I haven't reviewed it very closely, but the overall approach looks good.

lukecarbis:
@kienstra Ready for your full review.

kienstra:
Thanks, @lukecarbis!

php/blocks/class-block.php
```diff
@@ -111,19 +126,14 @@ public function from_json( $json ) {
...
 		if ( isset( $config['fields'] ) ) {
-			foreach ( $config['fields'] as $field ) {
-				$field_defaults = array( 'name', 'label', 'control', 'type', 'order' );
-				$field_settings = array_diff( array_keys( $field ), $field_defaults );
-				foreach ( $field_settings as $setting ) {
-					$field['settings'][ $setting ] = $field[ $setting ];
-				}
-				$this->fields[] = new Field( $field );
+			foreach ( $config['fields'] as $key => $field ) {
+				$this->fields[ $key ] = new Field( $field );
```
kienstra, line 130: Nice change here! It's nice how it moves the Field-related logic into `Field::from_array()`

kienstra (approved):
**Approved**

Hi @lukecarbis,
This looks good.

The Field logic is moved to the `Field` class, and the Location column doesn't appear anymore:

location-column-removed-here

PR #343, opened 2019-06-17
Create Custom Block Categories

### LINKED ISSUE #215: Custom block categories
Support for, and ability to define, custom block categories.

https://wordpress.org/gutenberg/handbook/designers-developers/developers/filters/block-filters/#managing-block-categories

ronalfy:
Resolves https://github.com/getblocklab/block-lab/issues/215

1. Added Block Category Meta below categories on post edit screen
2. Categories are saved as a site-wide option, which will be useful in Multisite setups.
3. Added a settings panel to remove existing categories.

See the demo video below on how it works:

https://www.dropbox.com/s/v6l42od8wipnhvs/github-pr.mp4?dl=0


js/admin.block-post.js
```diff
@@ -13,6 +13,7 @@
 	$(function() {
 		blockTitleInit();
 		blockIconInit();
+		blockCustomCategoryIconInit();
```
lukecarbis: I think we can entirely remove icons from our custom categories.

js/admin.block-post.js
```diff
@@ -54,6 +63,64 @@
...
+			let categoryVal = $.trim( $( '#block-properties-category-name' ).val() );
+			if ( '' === categoryVal ) {
+
+				// Show error message.
+				$( '#block-properties-category-create-status' ).removeClass().addClass( 'error' ).html( blockLab.emptyCategory );
+				$( '#block-properties-category-create-wrapper' ).find( 'input[type="text"]:first').focus();
+				return;
+			}
+			$( this ).html( blockLab.saving ).attr( 'disabled', 'disabled' );
+			wp.ajax.send( 'save_custom_category', {
```
lukecarbis: If we just save the category as post meta, and register it when the blocks are being registered, then there's no need to actually save the category independently at any point – and therefore no need for this ajax call.

php/admin/class-settings.php
```diff
@@ -84,6 +84,9 @@ public function render_page() {
 		$this->render_page_header();
 
 		switch ( $tab ) {
+			case 'blockcategories':
```
lukecarbis: I'd prefer not to introduce an new settings section if possible.

php/post-types/class-block-post.php
```diff
@@ -369,6 +396,54 @@ public function save_draft_button() {
...
 
+	/**
+	 * Returns icon fields based on passed classname and icon.
+	 *
+	 * @param string $classname The classname to use for the icon classes.
+	 * @param string $icon      The icon to use as a default.
+	 *
+	 * @return void
+	 */
+	private function get_icons( $classname = 'icon', $icon = '' ) {
```
lukecarbis: I'm not sure if it's worth spinning this out into its own method. The name confused me – maybe something like `render_icon_input()` would work better?

lukecarbis (changes requested):
@ronalfy Great start. Thanks for your work on this.

I'm a bit worried that it introduces too much complexity and heaps of new code. My main concern is around how can we distill this down into only its bare essentials?

lukecarbis:
Left some thoughts here: https://www.fromsmash.com/blocklab-343

Do you have PHPCS setup to autolint in your IDE?

php/admin/class-settings.php
```diff
@@ -101,9 +104,12 @@ public function render_page_header() {
 		?>
 		<h2><?php echo esc_html( get_admin_page_title() ); ?></h2>
 		<h2 class="nav-tab-wrapper">
-			<a href="<?php echo esc_url( add_query_arg( 'tab', 'license' ) ); ?>" title="<?php esc_attr_e( 'License', 'block-lab' ); ?>" class="nav-tab nav-tab-active dashicons-before dashicons-nametag">
+			<a href="<?php echo esc_url( add_query_arg( 'tab', 'license' ) ); ?>" title="<?php esc_attr_e( 'License', 'block-lab' ); ?>" class="nav-tab <?php echo ( ! isset( $_GET['tab'] ) || 'license' === $_GET['tab'] ) ? 'nav-tab-active' : ''; //phpcs:ignore ?> dashicons-before dashicons-nametag">
```
lukecarbis: You can use `filter_input` here. See how we do it:
https://github.com/getblocklab/block-lab/blob/d9f5a827cd51652766c343ed2768f0490fd79cb9/php/admin/class-settings.php#L42

js/admin.block-post.js
ronalfy: Roger that.

js/admin.block-post.js
ronalfy: My main concern is category re-use across the site and in a multisite setup. If we store categories as an independent option, there's re-use rather than storing as post meta for a per-site setup. What are your thoughts?

php/admin/class-settings.php
ronalfy: Sounds good.

php/post-types/class-block-post.php
ronalfy: Yeah, I moved it to its own method when I was trying to re-use icons for the block categories. Evidently, you can't create SVGs manually but at the block level. I'll move it back to its own context.

php/admin/class-settings.php
ronalfy: Removed settings screen, so this no longer applies.

php/post-types/class-block-post.php
ronalfy: I moved this function back to its own context.

ronalfy:
> Left some thoughts here: https://www.fromsmash.com/blocklab-343
> 
> Do you have PHPCS setup to autolint in your IDE?

By autolint, do you mean to show real-time errors? If so, yes.

js/admin.block-post.js
ronalfy: @lukecarbis I've posted a new video with the changes and my justification to keep Ajax saving and a site-wide option for categories for MS installs.

https://www.dropbox.com/s/r8urivzxmckp1uk/block-lab-custom-categories.mp4?dl=0

lukecarbis: @ronaldhuereca Given that blocks themselves don't work across sites, I don't think it's worth introducing a new site wide option to keep categories working across sites. I think it's reasonable to expect that categories won't follow you between sites. It might even be preferable.

I don't actually think it should be post meta, per se, but just saved in exactly the same way as the current category gets saved (which is inside a JSON blob).

lukecarbis:
@ronalfy Please see the implementation in #357.

PR #345, opened 2019-06-18
Remove duplicate id from fields

mukeshpanchal27:
**Steps to replicate issue**

1. Admin Panel > Block Lab > Add New
2. Add new field type as per your need
3. Add another new field type as per your need

**Issue**

Each new field generate new Field Label and Field Name field with duplicate id of "block-fields-edit-label-description" and "block-fields-edit-name-description"

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -642,7 +642,7 @@ class Block_Post extends Component_Abstract {
 							<label for="block-fields-edit-label-input_<?php echo esc_attr( $uid ); ?>">
 								<?php esc_html_e( 'Field Label', 'block-lab' ); ?>
 							</label>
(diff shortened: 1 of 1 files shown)


lukecarbis (changes requested):
Great pickup @mukeshpanchal27 .

I don't think these ID's are actually used by anything – maybe it would be better to simply remove the ID attribute altogether?

There's also `block-properties-keywords-description` on line 402, and `block-properties-keywords-description` on line 473. Obviously the first one is wrong, but I think we can remove the IDs from both.

mukeshpanchal27:
@lukecarbis updated PR with remove ids attributes which will duplicated when we add new fields from admin panel

lukecarbis (approved):

lukecarbis:
Thanks heaps @mukeshpanchal27 !

PR #346, opened 2019-06-19
Fix: Undefined index for category column

mukeshpanchal27:
**Steps to replicate issue**

Admin Panel > Block Lab > Add New
Add Title then leave page
Post will auto save
Go to Admin Panel > Block Lab there it shows "Notice: Undefined index: in /var/www/html/wordpress/wp-content/plugins/block-lab/php/post-types/class-block-post.php on line 1160"

**Issue**

When we add new field without saving it means when field save by auto save then plugin generate undefined notice

**Screenshot**
Error

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -1114,8 +1114,7 @@ class Block_Post extends Component_Abstract {
 			echo esc_html( implode( ', ', $block->keywords ) );
 		}
 		if ( 'category' === $column ) {
-			$block      = new Block( $post_id );
-			
+			$block = new Block( $post_id );
 			if ( ! empty( $block->category ) ) {
 				$categories = get_block_categories( get_post() );
 				$categories = wp_list_pluck( $categories, 'title', 'slug' );
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -1117,8 +1117,9 @@ class Block_Post extends Component_Abstract {
 			$block      = new Block( $post_id );
 			$categories = get_block_categories( get_post() );
 			$categories = wp_list_pluck( $categories, 'title', 'slug' );
-
-			echo esc_html( $categories[ $block->category ] );
+			if ( isset( $categories[ $block->category ] ) ) {
+				echo esc_html( $categories[ $block->category ] );
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


php/post-types/class-block-post.php
```diff
@@ -1117,8 +1117,9 @@ public function list_table_content( $column, $post_id ) {
 			$block      = new Block( $post_id );
 			$categories = get_block_categories( get_post() );
```
lukecarbis: Should we also include a check to see if the block has a category?
`if ( ! empty( $block->category ) )`

lukecarbis (changes requested):
Great! Thanks for finding this. One small question before merging.

php/post-types/class-block-post.php
mukeshpanchal27: Yes it's working fine if we add above conditional logic. But isset is already used in some conditional logic so i have added isset instead of not empty condition.

`if ( isset( $this->controls[ $control_name ] ) ) {`
`if ( isset( $icons[ $block->icon ] ) ) {`

lukecarbis: The problem is occurring because `$block->category` is equal to an empty string, and `$categories['']` is not set.

Checking if `$categories['']` is set does help, so we should absolutely keep that, but I'm just suggesting that it would also nice to check right after we instantiate the `Block` whether the category is empty. If it is, we can skip running `get_block_categories`, `wp_list_pluck`, etc.

lukecarbis: Like this: 

```php
$block = new Block( $post_id );

if ( ! empty( $block->category ) ) {
    $categories = get_block_categories( get_post() );
    $categories = wp_list_pluck( $categories, 'title', 'slug' );

    if ( isset( $categories[ $block->category ] ) ) {
        echo esc_html( $categories[ $block->category ] );
    }
}
```

mukeshpanchal27:
@lukecarbis Apply changes please check it and let us know in case of any issue

lukecarbis:
@mukeshpanchal27 Thanks again! It looks like Travis has picked up some minor code styling problems with the PR – we can't merge it unless it passes all these tests.

- Looks like you've got unnecessary indentation on line 1118.
- Since we've moved the definition of `$categories` inside the conditional, line 1117 only needs a single space before the equals sign

lukecarbis (approved):
Tested, and could replicate the issue. This PR resolved it.

Thank you Mukesh!

PR #348, opened 2019-06-20
Change the @param tag type from array to Block

kienstra:
* A follow-up to #340
* As Luke mentioned, this should be array because the parameter changed to an array.


kienstra:
**Request For Review**

Hi @lukecarbis,
Thanks for bringing this up. Could you please review this?

lukecarbis (approved):

PR #353, opened 2019-06-26
Fix an issue where the block_field( 'className' ) doesn't output anything

kienstra:
@RobStino found this issue, where `block_field( 'className' )` has no output. Similarly, `block_value( 'className' )` didn't return anything.

### DIFF
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -28,7 +28,7 @@ function block_field( $name, $echo = true ) {
 	if (
 		! isset( $block_lab_attributes ) ||
 		! is_array( $block_lab_attributes ) ||
-		! isset( $block_lab_config->fields[ $name ] )
+		( ! isset( $block_lab_config->fields[ $name ] ) && 'className' !== $name )
 	) {
 		return null;
 	}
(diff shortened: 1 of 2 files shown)


php/helpers.php
```diff
@@ -44,46 +44,48 @@ function block_field( $name, $echo = true ) {
...
-		case 'string':
-			$value = strval( $value );
-			break;
-		case 'textarea':
-			$value = strval( $value );
-			if ( isset( $block_lab_config->fields[ $name ]->settings['new_lines'] ) ) {
-				if ( 'autop' === $block_lab_config->fields[ $name ]->settings['new_lines'] ) {
-					$value = wpautop( $value );
+	if ( isset( $block_lab_config->fields[ $name ]->type ) ) {
+		switch ( $block_lab_config->fields[ $name ]->type ) {
```
kienstra, line 48: In a previous version ,[1.3.1](https://github.com/getblocklab/block-lab/blob/0dc89aab9b4e05a7d00ecdb063e0f10da5f1400f/php/helpers.php#L46), the switch statement in `block_field()` was wrapped in an `isset()` conditional:

https://github.com/getblocklab/block-lab/blob/0dc89aab9b4e05a7d00ecdb063e0f10da5f1400f/php/helpers.php#L45-L47

This PR adds that `isset()` conditional back in, though maybe there was another reason for removing it.

php/helpers.php
```diff
@@ -44,46 +44,48 @@ function block_field( $name, $echo = true ) {
...
+			case 'string':
```
kienstra, line 49: The diff inside this `switch` block is whitespace-only, from indenting it inside the new conditional. You might want to check `Hide whitespace changes`

php/helpers.php
```diff
@@ -28,7 +28,7 @@ function block_field( $name, $echo = true ) {
 	if (
 		! isset( $block_lab_attributes ) ||
 		! is_array( $block_lab_attributes ) ||
-		! isset( $block_lab_config->fields[ $name ] )
+		( ! isset( $block_lab_config->fields[ $name ] ) && 'className' !== $name )
```
kienstra, line 31: `className` isn't normally in `$block_lab_config->fields`

php/helpers.php
kienstra: Since `className` isn't in `$block_lab_config['fields']`, this caused an error before.

kienstra:
**Steps To Reproduce**

Thanks to @RobStino for finding this.

1. Create a new post, and add any Block Lab block
2. In the the Inspector, under Advanced > Additional CSS Class(es), add a class like `'foo-class'`
3. In the template's PHP file, like `block-test-class.php`, use `block_field( 'className' )`
4. Expected: it outputs the class, like `'foo-class'`
5. Actual: It doesn't output anything

kienstra:
**Request For Review**

Hi @lukecarbis and @RobStino,
Hope you're having a great week. Could you please test this? 

Though this looks to fix the issue for now, there might need to be more thought about `block_field()` eventually. 

It has a lot of special handling for `className` that might be confusing, or allow future bugs.

Thanks!

tests/php/test-helpers.php
```diff
@@ -21,8 +21,11 @@ public function test_block_field() {
 		global $block_lab_attributes, $block_lab_config;
```
kienstra: I'm making it a priority to add more unit tests to the plugin. They might have prevented an issue like this.

kienstra:
Also, I'm making it a priority to add more unit tests to the plugin. They might have prevented an issue like this.

lukecarbis (approved):
Tested – everything here works great. :)

PR #356, opened 2019-06-28
Add unit tests for classes in the admin/ directory

kienstra:
* So far, adds test classes for the `Admin\License` and `Admin\Upgrade` classes
* Adds more tests for the `Admin\Settings` class


php/admin/class-license.php
```diff
@@ -41,7 +41,7 @@ public function init() {
 	 * Register any hooks that this component needs.
 	 */
 	public function register_hooks() {
-		add_filter( 'pre_update_option_block_lab_license_key', array( $this, 'save_license_key' ), 10, 1 );
+		add_filter( 'pre_update_option_block_lab_license_key', array( $this, 'save_license_key' ) );
```
kienstra, line 44: Removed `10` and `1` as they're the defaults.

php/admin/class-import.php
```diff
@@ -204,7 +204,7 @@ public function validate_upload( $file ) {
 	 *
 	 * @param array $blocks An array of Block Lab content blocks.
 	 */
-	public function import( $blocks ) {
+	public function import_blocks( $blocks ) {
```
kienstra, line 207: There was a warning that this method of `import()` has the same name as the class, `Import`, and could be confused with a constructor.

kienstra:
Here's the current test coverage of the `admin/` directory:

admin-directory-coverage

php/admin/class-import.php
kienstra: This will require regression testing of importing, to make sure it doesn't break anything.

kienstra:
**Request For Review**

Hi @lukecarbis,
Hope your weekend's off to a great start.

Could you please review this? The repeater is my next priority.

Thanks!

lukecarbis:
Can confirm importing still works. :)

lukecarbis (approved):
Looks great @kienstra. Not sure why the code coverage doesn't show green across the board for the admin directory though?

Happy to see the inclusion of the `fixtures`, we can use that in the future to store mock block json.

kienstra:
Hi @lukecarbis,

> Looks great @kienstra. Not sure why the code coverage doesn't show green across the board for the admin directory though?

Good question. I think I can increase the coverage a little, I'll work on that.

>Happy to see the inclusion of the fixtures, we can use that in the future to store mock block json.

Ah, good point. I hadn't thought of that.

lukecarbis:
Didn't mean to ask for more tests! Just based on my review, I thought you'd covered off on everything.

PR #357, opened 2019-06-29
Custom categories

### LINKED ISSUE #215: Custom block categories

lukecarbis:
Closes #343
Closes #215

In this PR, block categories are never saved as custom post meta, and don't require an admin screen to manage. To delete a category, simply don't use it in any blocks.

This moves the category label and slug into the `Block` class, stored as an array ([the same format as WordPress](https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/post.php#L2131)). Unfortunately, since they were previously saved as strings, this requires a couple of small backwards compatibility functions.

Lastly, I've slightly refactored `Loader` to move the block registration portion of loading blocks out of `dynamic_block_loader`, and into its own method (`register_block`).

Screen Shot 2019-06-30 at 8 05 06 am


php/post-types/class-block-post.php
```diff
@@ -931,7 +943,28 @@ public function save_block( $data ) {
...
+					'icon'  => null,
+				);
+			} else {
+				$category_slugs = wp_list_pluck( $categories, 'slug' );
+				$category_key   = array_search( $category_slug, $category_slugs, true );
+				$category       = $categories[ $category_key ];
+			}
+
+			if ( ! $category ) {
+				$category = $categories[0];
```
kienstra: There are probably very few cases where `$categories[0]` won't be set. For that to happen, something would probably have to [filter block_categories](https://github.com/WordPress/wordpress-develop/blob/10855438eaca4da1aa24616fdc5f90f5745a45e4/src/wp-admin/includes/post.php#L2173) and return an empty array.

But what do you think about checking that it's set, to be sure?

Maybe something like:
```php
if ( ! $category ) {
	$category = isset( $categories[0] ) ? $categories[0] : '';
}
```

php/post-types/class-block-post.php
```diff
@@ -1115,14 +1148,7 @@ public function list_table_content( $column, $post_id ) {
...
 			$block = new Block( $post_id );
-			if ( ! empty( $block->category ) ) {
-				$categories = get_block_categories( get_post() );
-				$categories = wp_list_pluck( $categories, 'title', 'slug' );
-
-				if ( isset( $categories[ $block->category ] ) ) {
-					echo esc_html( $categories[ $block->category ] );
-				}
-			}
+			echo esc_html( $block->category['title'] );
```
kienstra, line 1151: Nice, this is a lot cleaner now.

js/blocks/loader/index.js
```diff
@@ -31,7 +31,7 @@ const registerBlocks = () => {
 		// Register the block.
 		registerBlockType( blockName, {
 			title: block.title,
-			category: block.category,
+			category: block.category.slug,
```
kienstra: What do you think about checking the type here?

```diff
diff --git a/js/blocks/loader/index.js b/js/blocks/loader/index.js
index 92d6120..e779c1e 100644
--- a/js/blocks/loader/index.js
+++ b/js/blocks/loader/index.js
@@ -31,7 +31,7 @@ const registerBlocks = () => {
                // Register the block.
                registerBlockType( blockName, {
                        title: block.title,
-                       category: block.category.slug,
+                       category: 'object' === typeof block.category ? block.category.slug : block.category,
                        icon: icon,
                        keywords: block.keywords,
                        attributes: blockAttributes( block ),
```

Without that, there were errors in my console, like:

```
The block "block-lab/range-controls" must have a registered category.
```

That error only appeared with blocks created before this PR. 

It looks like the new logic for saving categories only runs on clicking 'Update' or 'Save' in the Block Lab UI. So when I clicked 'Update' there, the console error didn't appear for the block:

logic-runs-here

kienstra (changes requested):
**Custom Category UI Looks Good**
*One Point About Possibly Checking The Type*

Hi @lukecarbis,
Nice work, the custom category UI looks good:

custom-cat-here

There's just one [point below](https://github.com/getblocklab/block-lab/pull/357#discussion_r298859854) about possibly checking the `typeof block.category`, to help with backwards compatibility.

js/blocks/loader/index.js
kienstra: The need to click 'Update' has also been an issue with a few of my changes, like with how the User control changed from storing a simple integer user ID to an object.

*Eventually*, we might think about having a [plugin update action](https://github.com/ampproject/amp-wp/pull/1487/files#diff-1a37f9f05bb7060179a77d7c2298231bR98) like Weston created. 

Then, its [callbacks](https://github.com/ampproject/amp-wp/pull/1487/files#diff-1a37f9f05bb7060179a77d7c2298231bR133) could apply changes like the one above, and users wouldn't have to click 'Update' for the new change.

Then, all of these changes could possibly be in their own class, like `Backwards_Compatibility`. 

Overall, this would probably add complexity. It'd just move the backwards-compatibility code to one place.

kienstra:
Also, like you mentioned, it's nice how there's no need to delete a custom category. As long as it's not selected in a block, it doesn't appear in the block editor (Gutenberg).

js/blocks/loader/index.js
lukecarbis: The type check is a great idea. I agree that we should have a plugin update action, and a `Backwards_Compatibility` or `Updates` class at some point in the future.

lukecarbis:
@kienstra Ready for you to take another look.

php/post-types/class-block-post.php
lukecarbis: Implemented in https://github.com/getblocklab/block-lab/pull/357/commits/b22cdfd24a0d10d5f78745317e3cee182c339077.

js/blocks/loader/index.js
lukecarbis: Implemented in https://github.com/getblocklab/block-lab/pull/357/commits/b22cdfd24a0d10d5f78745317e3cee182c339077.

kienstra:
**Approved**

Hi @lukecarbis,
Nice work, this looks good.

Adding a 'Custom Category' in the Block Lab UI works well. The custom category appears in the block editor and the 'Content Blocks' page.

Once #364 is merged into `develop`, merging `develop` into this should fix the failed Travis build

kienstra (approved):

PR #358, opened 2019-07-03
Add trailingslashit to the template locater

lukecarbis:
This PR makes it easier to change the template path directory, by switching to the `trailingslashit` function when locating templates, and adding documentation to the `block_lab_template_path` filter.


php/helpers.php
```diff
@@ -165,7 +165,17 @@ function block_field_config( $name ) {
  * @return string|array
  */
 function block_lab_locate_template( $template_names, $path = '', $single = true ) {
-	$path            = apply_filters( 'block_lab_template_path', $path );
+	/**
+	 * Filters the path where block templates are saved.
+	 *
+	 * Note that template names are prefixed with the blocks directory.
+	 * e.g. `blocks/block-template.php`
+	 *
```
kienstra, line 174: Maybe this could also mention:

```
 * The logic below will look for the prefixed template name inside the $path.
```

kienstra:
**Approved**

Hi @lukecarbis,
This looks good. Great idea to use `trailingslashit()`.

This method still got the templates as expected from the:

* Parent theme
* Child theme
* Template directory filtered with `'block_lab_template_path'`

kienstra (dismissed):

php/helpers.php
lukecarbis: Great idea. Added in https://github.com/getblocklab/block-lab/pull/358/commits/4cc09669bd1bb6343cbeed438b37c4fc96b1a52c.

lukecarbis:
@kienstra I just added that extra comment line that you suggested. Could you please re-approve?

kienstra (approved):
**Approved**

Hi @lukecarbis,
Thanks for making that commit. This looks good.

PR #363, opened 2019-07-10
Add a block attributes filter

lukecarbis:
cc. @phpbits


lukecarbis:
@kienstra This is a real quick win, if you could please review?

php/blocks/class-loader.php
```diff
@@ -187,7 +187,7 @@ public function get_block_attributes( $block ) {
 			}
 		}
 
-		return $attributes;
+		return apply_filters( 'block_lab_get_block_attributes', $attributes );
```
kienstra: @lukecarbis, would you mind if I committed a DocBlock for `apply_filters` here, including a `@param` tag for `$attributes`?

lukecarbis: Of course! I should have thought of that. Go for it.

kienstra: Thanks!

kienstra:
**Passing `$block` to filter?**

Hi @lukecarbis, 
what do you think about https://github.com/getblocklab/block-lab/pull/363/commits/3c335e49d477a9bd5e55f163a40a09c8cd3b5217, which also passes `$block` to the filter?

Otherwise, it might be hard to know which block it's for:

attributes-value

kienstra:
Also, I'm working on the failed Travis build in #364.

lukecarbis:
@kienstra Thanks for those two commits. They both look great! Especially 3c335e49d477a9bd5e55f163a40a09c8cd3b5217 – good thinking. Ready for you to re-review.

kienstra (approved):
**Approved**

Thanks, @lukecarbis! This should be ready to merge.

PR #364, opened 2019-07-11
Address an issue where fields from other blocks appeared in the Inspector

### LINKED ISSUE #362: Some fields from other blocks appear in the Inspector
Thanks to Grant for reporting this.

It looks like there are some fields that unexpectedly display in the Inspector when other blocks are selected. I saw this for 'Select' and 'Multi-Select' fields, but not 'Text' fields.

Steps to reproduce
1. Create a new Block Lab Block, like 'Block 1 Testing Inspector'
2. Add 'Select' and 'Multi-Select' fields, and select 'Inspector' instead of 'Editor'

select-multi-select

3. Publish the block

4. Create another block, like 'Block Only Text In Editor'

5. It's probably not very important what fields it has. It could just have a text field in the editor:

block-only-text-in-editor

6. Publish the block

7. Create a new post, and add both blocks

8. Expected: When editing the block 'Block Only Text In Editor,' it should not have the Inspector field from the other block

9. Actual: the Inspector has a field from the block in step 2:

inpector-here

kienstra:
There are steps to reproduce this in https://github.com/getblocklab/block-lab/issues/362#issue-466449829.

This issue didn't always appear. Sometimes there were fields from other blocks in the inspector, and sometimes there weren't. Even the exact same post sometimes had this problem, and sometimes it didn't.

But this does seem to fix the issue as described in https://github.com/getblocklab/block-lab/issues/362#issue-466449829.

The `field` was actually an object, and it was attempting to concatenate it with a string. 

the-inspector

Fixes #362


kienstra:
If it's alright, I'll keep working on the failed Travis builds tomorrow.

kienstra:
**Request For Review**

Hi @lukecarbis,
Thanks for your patience. This is ready for review.

The Travis fix here should work for #357 and #363 as well.

Feel free to merge this if you approve. I'll be on a flight tomorrow.

Then, if you merge the `develop` branch into the other PRs, their Travis builds should pass.

tests/php/admin/test-class-import.php
```diff
@@ -216,11 +228,24 @@ public function test_render_page() {
...
+		Monkey\Functions\expect( 'is_uploaded_file' )
+			->once()
+			->with( $tmp_name )
+			->andReturn( true );
+
+		Monkey\Functions\expect( 'move_uploaded_file' )
+			->once()
+			->andReturn( true );
+
+		Monkey\Functions\expect( 'chmod' )
```
kienstra, line 245: I think the `chmod()` function is the one that caused the failed build.

lukecarbis (approved):
I can't seem to replicate the problem, but the change you've suggested does make sense to me.

PR #366, opened 2019-07-14
Apply loaded controls filter to the inspector controls

lukecarbis:
When trying to create a custom control add-on for Block Lab, I encountered a bug whereby it's not possible to add custom controls to the inspector sidebar.

This PR fixes the problem.

**Before**
Screen Shot 2019-07-14 at 12 24 26 pm

**After**
Screen Shot 2019-07-14 at 12 24 41 pm


kienstra:
**Approved**

Hi @lukecarbis,
This looks good. Like you mentioned, it makes sense that the `controls` for the inspector are also filtered, like the [controls for the editor](https://github.com/getblocklab/block-lab/blob/255a41696923f1319e9bdbf0701aefa902836977/js/blocks/loader/edit.js#L19).

In the future, it'd be good if we could use camelCase for filter names, following the convention in Gutenberg. Though that still looks strange with my PHP background 😄. 

But of course this filter already exists, so it's fine.

kienstra (approved):

PR #370, opened 2019-07-22
Load child theme templates before parent theme

lukecarbis:
Closes #334.

Tested with both child theme and parent theme template files and stylesheets.

Screen Shot 2019-07-22 at 7 20 20 pm
Screen Shot 2019-07-22 at 7 20 10 pm

My parent template looks like this:
```
<h3 class="css-bugfix">parent</h3>
```

My child template looks like this:
```
<h3 class="css-bugfix">child</h3>
```

My parent stylesheet looks like this:
```
h3.css-bugfix {
	color: hotpink;
}
```

My child stylesheet looks like this:
```
h3.css-bugfix {
	color: yellow;
}
```

When I view the block (tested in both the editor and frontend), I see the child template / stylesheet, as expected:
Screen Shot 2019-07-22 at 7 21 16 pm

In `develop` I (incorrectly) see the parent template / stylesheet:
Screen Shot 2019-07-22 at 7 21 38 pm

PR #371, opened 2019-07-22
Global styles

### LINKED ISSUE #351: Global CSS Template File
When building blocks for a site, I often have css styles that are re-useable across blocks, as well as the the larger theme. An example is styling for a `tile` component. This tile component could have styling that includes things like `border`, `padding`, and `background-color`. These styles could easily be relevant to multiple blocks.

The current options for handling these global styles have flaws.
1. Include styles in each block css template file. This is not DRY.
2. Include styles in theme stylesheet. These are not loaded in the editor.
3. Include styles in preview css template file. Also not DRY.
4. Include styles in editor-styles.css and enqueue to the admin. Not DRY.

An idea for a solution is to have a global `blocks.css` file that is saved in the `blocks` folder. This file will be:
- loaded in the editor
- loaded on the frontend unconditionally.

Open to discussion, especially around issues with it, but I think this could prove very helpful.

lukecarbis:
With this PR, Block Lab will try to enqueue a global stylesheet.

It will try to do this under one of two conditions:
1. You are using Gutenberg
2. _Any_ custom block lab block is used on the page

*Note*: These styles won't load if you're not using any Block Lab blocks.

The global stylesheet will be loaded if it exists in one of these places:
- `{theme}/blocks/blocks.css`
- `{theme}/blocks/css/blocks.css`
- `{child theme}/blocks/blocks.css`
- `{child theme}/blocks/css/blocks.css`

If the global stylesheet exists in multiple places, it will only be loaded the first time it's found. So if you've got one in `blocks/blocks.css` _and_ `blocks/css/blocks.css`, then second stylesheet will be ignored.

Note that once #370 is merged, the child theme will be preferenced over the parent theme.

Closes #351 .

### DIFF
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -105,6 +105,8 @@ class Loader extends Component_Abstract {
 			$this->enqueue_block_styles( $block_name, array( 'preview', 'block' ) );
 		}
 
+		$this->enqueue_global_styles();
+
 		// Used to conditionally show notices for blocks belonging to an author.
 		$author_blocks = get_posts(
 			array(
@@ -277,6 +279,12 @@ class Loader extends Component_Abstract {
 			}
 
 			$this->enqueue_block_styles( $block->name, 'block' );
+
+			/**
+			 * The wp_enqueue_style function handles duplicates, so we don't need to worry about multiple blocks
+			 * loading the global styles more than once.
+			 */
+			$this->enqueue_global_styles();
 		}
 
 		$block_lab_attributes = $attributes;
@@ -345,6 +353,30 @@ class Loader extends Component_Abstract {
 		if ( ! empty( $stylesheet_url ) ) {
 			wp_enqueue_style(
 				"block-lab__block-{$name}",
+				$stylesheet_url,
+				array(),
+				wp_get_theme()->get( 'Version' )
+			);
+		}
+	}
+	/**
+	 * Enqueues global block styles.
+	 */
+	public function enqueue_global_styles() {
+		$locations = array(
+			'blocks/css/blocks.css',
+			'blocks/blocks.css',
+		);
+
+		$stylesheet_path = block_lab_locate_template( $locations );
+		$stylesheet_url  = str_replace( untrailingslashit( ABSPATH ), '', $stylesheet_path );
+
+		/**
+		 * Enqueue the stylesheet, if it exists.
+		 */
+		if ( ! empty( $stylesheet_url ) ) {
+			wp_enqueue_style(
+				'block-lab__global-styles',
 				$stylesheet_url,
 				array(),
 				wp_get_theme()->get( 'Version' )
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -359,6 +359,7 @@ class Loader extends Component_Abstract {
 			);
 		}
 	}
+
 	/**
 	 * Enqueues global block styles.
 	 */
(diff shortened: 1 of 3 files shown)


php/blocks/class-loader.php
```diff
@@ -351,6 +359,30 @@ public function enqueue_block_styles( $name, $type = 'block' ) {
 			);
 		}
 	}
+	/**
```
kienstra, line 363: Not a big deal, but this should probably have an empty line above it.

tests/php/blocks/test-class-loader.php
```diff
@@ -47,34 +47,39 @@ public function test_render_block_template() {
...
+				wp_enqueue_script( $slug, $script_url, array(), '0.1', true );
+			}
+		);
 
 		$this->instance->render_block_template( $block, array() );
 		$scripts = wp_scripts();
 		$script  = $scripts->registered[ $slug ];
 
-		$this->assertTrue( in_array( $slug, $scripts->queue ) );
+		$this->assertTrue( in_array( $slug, $scripts->queue, true ) );
```
kienstra, line 82: Ah, good idea to add `true`.

tests/php/blocks/test-class-loader.php
```diff
@@ -131,4 +136,42 @@ public function test_enqueue_block_styles() {
...
+	 * @covers \Block_Lab\Blocks\Loader::enqueue_global_styles()
+	 */
+	public function test_enqueue_global_styles() {
+		$wp_styles       = wp_styles();
+		$enqueue_handle  = 'block-lab__global-styles';
+		$stylesheet_path = get_template_directory();
+
+		if ( ! file_exists( $stylesheet_path . '/blocks/' ) ) {
+			mkdir( $stylesheet_path . '/blocks/' );
+			mkdir( $stylesheet_path . '/blocks/css/' );
```
kienstra, line 152: Good idea to create the directory.

tests/php/blocks/test-class-loader.php
```diff
@@ -131,4 +136,42 @@ public function test_enqueue_block_styles() {
...
+		}
+
+		// In order of reverse priority.
+		$files = array(
+			"{$stylesheet_path}/blocks/blocks.css",
+			"{$stylesheet_path}/blocks/css/blocks.css",
+		);
+
+		// Check that the correct stylesheet is enqueued.
+		foreach ( $files as $key => $file ) {
```
kienstra, line 162: Good idea to run this in a `foreach` loop.

kienstra:
**Approved**

Hi @lukecarbis,
This looks good. The only code review points is very minor.

Like you mentioned, I also saw that the stylesheet is loaded if it's in any of these 4 locations:

>{theme}/blocks/blocks.css
>{theme}/blocks/css/blocks.css
>{child theme}/blocks/blocks.css
>{child theme}/blocks/css/blocks.css

blocks-css-present

It looks like it's *always* enqueued in the block editor, regardless of whether a Block Lab block was added. But maybe that's intentional, as one could be added.

kienstra (dismissed):

lukecarbis:
@kienstra Yes, always enqueuing it in the editor is intentional, so it's there and ready in case a block is added. I'll sort out that extra line now.

php/blocks/class-loader.php
lukecarbis: Fixed in 16befc7.

kienstra:
>@kienstra Yes, always enqueuing it in the editor is intentional, so it's there and ready in case a block is added.

Sounds good.

kienstra (approved):

jcrooke:
I don't think the stylesheet should get included when you're viewing the blocks on the front-end, it should only be included when in the Gutenberg editor - otherwise you have a block stylesheet being loaded alongside your regular stylesheet.

PR #373, opened 2019-07-25
Add `editorskit` on helper attributes checker for plugin integration

phpbits:
Hi,

I've added `editorskit` attributes checker aside from `className` in order for me to create custom helper for devices visibility classes. Thanks!

Cheers,
Jeffrey

### DIFF
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -28,7 +28,7 @@ function block_field( $name, $echo = true ) {
 	if (
 		! isset( $block_lab_attributes ) ||
 		! is_array( $block_lab_attributes ) ||
-		( ! isset( $block_lab_config->fields[ $name ] ) && 'className' !== $name )
+		( ! isset( $block_lab_config->fields[ $name ] ) && 'className' !== $name && 'editorskit' !== $name )
 	) {
 		return null;
 	}
@@ -41,6 +41,11 @@ function block_field( $name, $echo = true ) {
 	// Cast default Editor attributes appropriately.
 	if ( 'className' === $name ) {
 		$value = strval( $value );
+	}
+
+	// Return immediately for EditorsKit attributes
+	if ( 'editorskit' === $name ) {
+		return $value;
 	}
 
 	// Cast block value as correct type.


kienstra:
**Thanks!**
*Question about use case*

Hi @phpbits,
Thanks for opening this, and for being active in Block Lab.

>I've added editorskit attributes checker aside from className in order for me to create custom helper for devices visibility classes. Thanks!

Could you please elaborate on the use case of passing `'editorskit'` to `block_field()`?

Where would the value for that come from?

I'd like to avoid referencing other plugins directly in this plugin. But maybe there's a way to extend this or a filter that would accomplish the same thing.

phpbits:
@kienstra I'm planning to create helper function for EditorsKit classnames that are being used on device visibility since JS filter won't work on PHP-registered blocks. Filter might work but it has to be added before the `if` conditions. Probably move this one before the conditions too then add filter afterwards :

`$value = false; // This is a good default, it allows us to pick up on unchecked checkboxes.`

Let me know your thoughts. Thanks!

kienstra:
Hi @phpbits,
Thanks for your patience.

1. Where would the value from `block_field( 'editorskit' )` come from?

Would there be another `<input>`, similar to the one that sets the `className`?

additional-class

>@kienstra I'm planning to create helper function for EditorsKit classnames that are being used on device visibility since JS filter won't work on PHP-registered blocks. 

2. Could you please elaborate on the helper function for EditorsKit, and where its value(s) comes from?

phpbits:
@kienstra I'm adding separate section with Toggles and then I'm adding the specific classnames via JS filter based on the value. EditorsKit is also free and you can also try it if you want to : https://wordpress.org/plugins/block-options/ . Thanks!

The helper function will bridge the gap between JS and PHP since classnames added via JS filters are not automatically added on `className` attributes. So whenever users create the rendering for BlockLab then can just add :

`<div class=" <?php block_field('className');?> <?php editorskit_block_classname(); ?>">Content</div>`

Let me know your thoughts. Thanks!

kienstra:
Hi @phpbits,
Thanks!

Are the following 2 points correct?

1. It sounds like the helper function might look something like this:

```php
function editorskit_block_classname() {
        return function_exists( 'block_field' ) ? block_field( 'editorskit' ) : null;
}
```

2. The value of `'editorskit'` will be in the block attributes (the global `$block_lab_attributes`):

https://github.com/getblocklab/block-lab/blob/314bf85a825e9c21fa3fce36dbe6eea85212c0a4/php/helpers.php#L26

block-lab-attributes

kienstra:
Would a filter like this work for your case?

```diff
diff --git a/php/helpers.php b/php/helpers.php
index 8066509..7b9eb77 100644
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -25,11 +25,27 @@ function block_field( $name, $echo = true ) {
         */
        global $block_lab_attributes, $block_lab_config;
 
-       if (
-               ! isset( $block_lab_attributes ) ||
-               ! is_array( $block_lab_attributes ) ||
-               ( ! isset( $block_lab_config->fields[ $name ] ) && 'className' !== $name )
-       ) {
+       $is_name_allowed = (
+               isset( $block_lab_attributes )
+               &&
+               is_array( $block_lab_attributes )
+               &&
+               ( isset( $block_lab_config->fields[ $name ] ) || 'className' === $name )
+       );
+
+       /**
+        * Filters whether the $name argument to this function is allowed.
+        *
+        * By default, this only allows a $name that is in the block fields or the 'className'.
+        * This filter can prevent returning null from this function
+        * when looking for another value in the block attributes.
+        *
+        * @param bool   $is_name_allowed Whether the name is allowed.
+        * @param string $name The name of value to get.
+        */
+       $is_block_field_name_allowed = apply_filters( 'is_block_field_name_allowed', $is_name_allowed, $name );
+
+       if ( ! $is_block_field_name_allowed ) {
                return null;
        }
```

phpbits:
@kienstra Can I modify the `return` with that filter too? I need to return `$value` since EditorsKit values are stored in `object`. Thanks a lot!

phpbits:
> 1. It sounds like the helper function might look something like this:
> 
> ```
> function editorskit_block_classname() {
>         return function_exists( 'block_field' ) ? block_field( 'editorskit' ) : null;
> }
> ```

I've tried this one but didn't work. `editorskit` is stored on the same level as `className` and I needed the full `$value` on return. Thanks!

kienstra:
>@kienstra Can I modify the return with that filter too? I need to return $value since EditorsKit values are stored in object. Thanks a lot!

Hi @phpbits,
That [filter above](https://github.com/getblocklab/block-lab/pull/373#issuecomment-516999394) wouldn't modify `$value`, but would this existing filter work for that?

https://github.com/getblocklab/block-lab/blob/314bf85a825e9c21fa3fce36dbe6eea85212c0a4/php/helpers.php#L83-L90

phpbits:
@kienstra Do you have a branch that I can test? I'll try it out on my end first and let you know if the filter works as expected. Thanks!

kienstra:
>@kienstra Do you have a branch that I can test? I'll try it out on my end first and let you know if the filter works as expected. Thanks!

Hi @phpbits,
Sure, the branch [add/allowed-filter](https://github.com/getblocklab/block-lab/tree/add/allowed-filter) has the [filter above](https://github.com/getblocklab/block-lab/pull/373#issuecomment-516999394).

kienstra:
@phpbits, it'd be great if you could share the code that you're using to integrate with this. It could help to show that this use case needs a filter, and maybe if there would be other similar use cases.

phpbits:
@kienstra will surely do! I'll create a tutorial when the integration is available. I'll test the branch. Thanks!

phpbits:
> Hi @phpbits,
> That [filter above](https://github.com/getblocklab/block-lab/pull/373#issuecomment-516999394) wouldn't modify `$value`, but would this existing filter work for that?
> 
> https://github.com/getblocklab/block-lab/blob/314bf85a825e9c21fa3fce36dbe6eea85212c0a4/php/helpers.php#L83-L90

@kienstra Checked and it seems that `$name` parameter is also needed in order to manage the `$value`. It's returning the original `false` value.

phpbits:
@kienstra Sorry! Figured it out! It seems that I don't need the `$name` parameters after all :)

phpbits:
@kienstra The filter works 💯 By adding the helper function to the wrapper element, the devices visibility classnames can be easily generated;

`<div class="<?php echo editorskit_blocklab_classname();?>"><?php block_field( 'new-field' ); ?></div>`

Here's the link of the integration : https://github.com/phpbits/block-options/blob/integration/block-lab/includes/helper.php 👍

kienstra:
Hi @phpbits,
Thanks for the [link](https://github.com/phpbits/block-options/blob/integration/block-lab/includes/helper.php) for the integration. I'll look at it more, and think about this filter a little more.

phpbits:
@kienstra You are very welcome! I've included the helper function the EditorsKit latest version. Let me know when the filters are available and I'll create a tutorial. Thanks!

php/helpers.php
```diff
@@ -43,6 +43,11 @@ function block_field( $name, $echo = true ) {
 		$value = strval( $value );
 	}
 
+	// Return immediately for EditorsKit attributes
```
lukecarbis, line 46: No period at the end of this comment is causing Travis to fail the `phpcs` check. Seems minor, but could you please change it to:
`// Return immediately for EditorsKit attributes.`

lukecarbis (changes requested):

php/helpers.php
kienstra: Hi @lukecarbis, I should have commented on this PR, but the current approach we're considering is a different branch entirely, [add/allowed-filter](https://github.com/getblocklab/block-lab/pull/373#issuecomment-517110834).

kienstra:
If it's alright, I'm closing this in favor of the [add/allowed-filter](https://github.com/getblocklab/block-lab/tree/add/allowed-filter) branch, though I still need to consider that more and open a PR.

PR #374, opened 2019-07-28
Add padding to right side of upgrade page

lukecarbis:
The upgrade page is missing padding on the right hand side. This PR fixes that with a one-line CSS change.

Before:
Screen Shot 2019-07-28 at 2 47 28 pm

After:
Screen Shot 2019-07-28 at 2 46 52 pm


kienstra:
**Approved**

Hi @lukecarbis,
The new `padding-right` looks good:

padding-added-now

kienstra (approved):

PR #377, opened 2019-08-05
Add Utils class with helper methods for listing block template locations

lukecarbis:
Fix for a bug affecting the template location display when the template is in one of the new template locations.

Bug description: https://d.pr/v/iPUk0q


php/blocks/class-loader.php
```diff
@@ -398,27 +394,20 @@ public function block_template( $name, $type = 'block' ) {
 
 		// So lets fix it.
 		if ( empty( $wp_query ) ) {
-			$wp_query = new \WP_Query(); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+			$wp_query = new \WP_Query(); // phpcs: ignore WordPress.WP.GlobalVariablesOverride.Prohibited.
 		}
 
 		$types         = (array) $type;
 		$located       = '';
 		$template_file = '';
```
kienstra: It looks like this assignment of `$template_file` might not be needed anymore.

php/helpers.php
```diff
@@ -227,6 +227,22 @@ function block_lab_locate_template( $template_names, $path = '', $single = true
...
 
+/**
+ * Return an array of possible template locations.
+ *
+ * @param string $name The name of the block (slug as defined in UI).
+ * @param string $type The type of template to load.
+ *
+ * @return array
+ */
+function block_lab_template_locations( $name, $type = 'block' ) {
```
kienstra: With the current architecture, this `helpers.php` file is probably the best place for this function.

But in the future, it might be nice to look for another location for this. It'd be nice if it was in a class, and not in the global scope.

kienstra: The only concern might be if by chance someone sees this function and starts using it in another plugin. Then, we'd have a harder time deprecating or removing it. 

But I'm not sure how someone would use this function.

php/blocks/class-loader.php
```diff
@@ -336,11 +336,7 @@ public function enqueue_block_styles( $name, $type = 'block' ) {
 		foreach ( $types as $type ) {
 			$locations = array_merge(
 				$locations,
-				array(
-					"blocks/{$name}/{$type}.css",
-					"blocks/css/{$type}-{$name}.css",
-					"blocks/{$type}-{$name}.css",
-				)
+				block_lab_template_locations( $name, $type )
```
kienstra: This edit might not be necessary. It looks to be causing the [failed unit test](https://travis-ci.com/getblocklab/block-lab/jobs/222315202).

php/blocks/class-loader.php
```diff
@@ -434,7 +423,7 @@ public function block_template( $name, $type = 'block' ) {
 				'<div class="notice notice-warning">%s</div>',
 				wp_kses_post(
 					// Translators: Placeholder is a file path.
-					sprintf( __( 'Template file %s not found.' ), '<code>' . esc_html( $template_file ) . '</code>' )
+					sprintf( __( 'Template file %s not found.' ), '<code>' . esc_html( $templates[0] ) . '</code>' )
```
kienstra, line 425: `$templates[0]` will probably always be set, unless there's a change to the codebase.

But what do you think about checking that `isset( $templates[0] )`, to future-proof this? Maybe in the conditional above:

```php
if ( ! current_user_can( 'edit_posts' ) || ! isset( $templates[0] ) ) {
	return;
}
```

php/post-types/class-block-post.php
```diff
@@ -325,7 +325,8 @@ public function add_meta_boxes() {
 		);
 
 		if ( isset( $post->post_name ) && ! empty( $post->post_name ) ) {
-			$template = block_lab_locate_template( 'blocks/block-' . $post->post_name . '.php', '', true );
+			$locations = block_lab_template_locations( $post->post_name, 'block' );
```
kienstra: What do you think about not passing the second argument of `'block'`? It looks to be the default argument.

php/blocks/class-loader.php
```diff
@@ -398,27 +394,20 @@ public function block_template( $name, $type = 'block' ) {
 
 		// So lets fix it.
 		if ( empty( $wp_query ) ) {
-			$wp_query = new \WP_Query(); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+			$wp_query = new \WP_Query(); // phpcs: ignore WordPress.WP.GlobalVariablesOverride.Prohibited.
```
kienstra: Later, it'd be nice to add a unit test for `block_template()`, as that's a really important method. You already added a really nice test for `test_enqueue_block_styles()` 😄 

I don't think this has to happen in this PR, though.

kienstra (changes requested):
**Looks Good**
*A Few Small Points*


Hi @lukecarbis,
This looks good, and it fixes the issue in my local.

There are just a few small points here. Also, if it's alright, I pushed a small commit to add a unit test.

# Before

When the template was a `block.php` file in a directory with the block name, like `test-image/block.php`, the 'Edit Block' UI didn't show the template as found.

For example, 
`<theme-directory>/blocks/test-image/block.php`

bl-before

# After

As expected, the UI shows this template:

bl-after

php/blocks/class-loader.php
lukecarbis: Great idea!

php/blocks/class-loader.php
lukecarbis: Fixed in 3fc23ef

php/blocks/class-loader.php
lukecarbis: Fixed in 06ad5a0

php/post-types/class-block-post.php
lukecarbis: Fixed in e970caa

php/blocks/class-loader.php
lukecarbis: This is addressed in 96550ae.

lukecarbis:
@kienstra Thanks for reviewing! Could you please take a fresh look now?

RobStino (dismissed):
Tested and this successfully addresses the issue.
Thanks for the efforts on this @lukecarbis and @kienstra

kienstra:
Hi @lukecarbis,
Sorry for the delay. I'm working on this now.

kienstra:
Hi @lukecarbis, if it's alright, I'm going to push a commit to refactor where utility functions are. It's definitely open for discussion 😄

kienstra:
Hi @lukecarbis,
What do you think about https://github.com/getblocklab/block-lab/pull/377/commits/a26d3972655d81c7ad28a17b567fcb571df5f43e, which moves this PR's helper functions to a new class, `Utils`?

Like we talked about, another option might be:
1. Put the helper functions, like `block_lab_template_locations()`, in a class like [Loader](https://github.com/getblocklab/block-lab/blob/905af1bc45e709e91f0faa051a4427c887045dfb/php/blocks/class-loader.php)

2. To enable access to the functions, either change [->plugin->components](https://github.com/getblocklab/block-lab/blob/905af1bc45e709e91f0faa051a4427c887045dfb/php/class-plugin-abstract.php#L22) from `protected` to `public`, or add `->loader` as a `public` property of [Plugin](https://github.com/getblocklab/block-lab/blob/905af1bc45e709e91f0faa051a4427c887045dfb/php/class-plugin-abstract.php#L15).

This would work, but it would essentially move [Loader](https://github.com/getblocklab/block-lab/blob/905af1bc45e709e91f0faa051a4427c887045dfb/php/blocks/class-loader.php) to the global scope. So any plugin could call something like `block_lab()->loader->render_block_template()`

As an alternative, this [commit](https://github.com/getblocklab/block-lab/pull/377/commits/a26d3972655d81c7ad28a17b567fcb571df5f43e) creates a new utility function class. 

The class `Plugin` already has utility functions, like [get_assets_url](https://github.com/getblocklab/block-lab/blob/905af1bc45e709e91f0faa051a4427c887045dfb/php/class-plugin-abstract.php#L227). This does something similar, but adds the new functions to `Plugin` via the `->utils` property.

kienstra:
>I think we tend to go with the latter in other parts of the plugin, and it means that we don’t need the class to contain a plugin property.

Ah, interesting. I hadn't thought of that.

lukecarbis:
Would it make sense to move the `is_pro()` method to utils?

php/blocks/class-loader.php
```diff
@@ -394,7 +394,7 @@ public function block_template( $name, $type = 'block' ) {
 
 		// So lets fix it.
 		if ( empty( $wp_query ) ) {
-			$wp_query = new \WP_Query(); // phpcs: ignore WordPress.WP.GlobalVariablesOverride.Prohibited.
+			$wp_query = new \WP_Query(); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
```
lukecarbis, line 397: Weird that this wasn't picked up in my version of phpcs. I'll update.

kienstra: All good, I was having a lot of trouble running PHPCS locally.

kienstra:
>If we were fancy, we could even create some sort of pass through magic method to skip the utils property. Like:
block_lab()->get_block_lab_stylesheet_location()

>Would it make sense to move the is_pro() method to utils?

Great ideas, https://github.com/getblocklab/block-lab/pull/377/commits/e74eb0c864172be18edcd6efba1d9be6c3c12d11 applies both of them.

php/blocks/class-loader.php
kienstra: Thanks, that looks good.

php/blocks/class-loader.php
kienstra: Thanks, that looks great.

php/post-types/class-block-post.php
kienstra: Thanks, that looks good!

php/blocks/class-loader.php
kienstra: Nice, good idea to create a separate method for stylesheets.

php/blocks/class-utils.php
```diff
@@ -0,0 +1,141 @@
...
+	 * and allows to be called when STYLESHEET_PATH has not been set yet. Handy for async.
+	 *
+	 * @param string|array $template_names Templates to locate.
+	 * @param string       $path           (Optional) Path to locate the templates first.
+	 * @param bool         $single         `true` - Returns only the first found item. Like standard `locate_template`
+	 *                                     `false` - Returns all found templates.
+	 *
+	 * @return string|array
+	 */
+	public function locate_template( $template_names, $path = '', $single = true ) {
```
kienstra: This is simply moved from `helpers.php`, and not changed at all. It'd be good to have a unit test for this soon.

php/helpers.php
```diff
@@ -151,82 +151,6 @@ function block_field_config( $name ) {
 	return (array) $block_lab_config->fields[ $name ];
 }
 
-/**
```
kienstra, line 154: Now that there's a class `Utils`, it'd be good to add any new plugin-wide methods there. Unless the function needs to be a plain function for users, like `block_field()`.

php/blocks/class-utils.php
kienstra: It now has a unit test.

kienstra: Update: there's now a unit test for it.

kienstra:
@lukecarbis,
Good to see you yesterday.

Could you please review this latest changes here when you have a chance? 

I moved more functions into the `Utils` class, and added more tests.

Thanks, Luke!

lukecarbis (commented):
These changes look great! Nice work.

Three suggestions:

1. Should `Utils` class be moved to the global namespace? i.e. `Block_Lab\Utils` instead of `Block_Lab\Blocks\Utils`? This would make more sense to me, especially for methods such as `is_pro()`.

2. Should the `block_lab_allowed_svg_tags()` and `block_lab_get_icons()` functions be moved from `helpers.php` to the `Utils` class? I would prefer this, as it would keep `helpers.php` as a space only used for template tags. There's a very small chance it could break backwards compatibility, but it's unlikely anyone would be using these global functions.

3. If we were worried about breaking backwards compatibility with these functions, this could be a good opportunity to include a `deprecated.php`. We could include via `helpers.php` and re-route old functions… e.g.

```PHP
function block_lab_allowed_svg_tags() {
    // Add deprecated notice here.
    block_lab()->allowed_svg_tags();
}
```

This could also be a good place to deprecate actions / filters.

lukecarbis:
Another thought – how about `Utilities` instead of `Utils`? Is `Utils` an accepted convention?

lukecarbis:
Going with `Util` to match WordPress.

lukecarbis:
@kienstra Could you please take another look? This hotfix is turning into a rewrite!

kienstra:
>@kienstra Could you please take another look? This hotfix is turning into a rewrite!

Haha, yeah. I'm looking at it now.

kienstra:
It looks good so far. I'll continue reviewing in several hours if that's alright.

php/deprecated.php
```diff
@@ -0,0 +1,58 @@
...
+ * Show a PHP error to warn developers using deprecated functions.
+ *
+ * @param string $function    The function that was called.
+ * @param string $version     The version of Block Lab that deprecated the function.
+ * @param string $replacement The function that should have been called.
+ */
+function block_lab_deprecated_function( $function, $version, $replacement ) {
+	_deprecated_function(
+		// filter_var is used for sanitization here as it allows arrow functions ("->").
+		filter_var(
```
kienstra, line 24: What do you think about passing something like `FILTER_SANITIZE_ENCODED` as the 2nd argument of `filter_var()`? 

Good idea to use that function. But so far, it looks like it doesn't sanitize much. 

For example, `filter_var( '<script>doThis();</script>' )` doesn't remove anything.

I'm not sure if `FILTER_SANITIZE_ENCODED` is the right argument, though. It does looks to allow `->`, and it strips `<script>`.

lukecarbis: Great thinking! I went with `FILTER_SANITIZE_STRING` in the end, as:

- It allows arrow functions without encoding them (`->`)
- It removes `<script>` and other malicious things
- We use it elsewhere in the plugin

lukecarbis: Implemented in 7b98dd7.

kienstra: Nice, good idea.

kienstra: Like you mentioned, I also see how passing `FILTER_SANITIZE_STRING` removes `<script>`, but doesn't remove things we need like `->` or `()`

php/deprecated.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+/**
+ * Show a PHP error to warn developers using deprecated functions.
+ *
+ * @param string $function    The function that was called.
+ * @param string $version     The version of Block Lab that deprecated the function.
+ * @param string $replacement The function that should have been called.
+ */
+function block_lab_deprecated_function( $function, $version, $replacement ) {
```
kienstra, line 21: Nice helper function!

kienstra:
@lukecarbis, is it alright if I push a commit to this to fix a mistake of mine?

lukecarbis:
Absolutely

php/class-plugin-abstract.php
```diff
@@ -217,6 +247,14 @@ public function set_version( $file ) {
 		return $this;
 	}
 
+	/**
+	 * Instantiates the Util class, with utility methods.
+	 */
+	public function set_util() {
```
lukecarbis, line 253: Nice move, moving this to the abstract.

kienstra:
Thanks! Your good idea to throw the Error:

```php
throw new \Error( "Call to undefined method $class::$name()" );
```

...reminded me to add something like that to `Plugin::__call()`. And that revealed that `Loader::retrieve_blocks()` was calling `block_lab->locate_template()` before `block_lab->util` was even set. And it failed silently.

So I added an exception like you added, and set `block_lab->util` earlier.

block-lab.php
```diff
@@ -114,7 +117,8 @@ function block_lab() {
 	->set_file( __FILE__ )
 	->set_slug( 'block-lab' )
 	->set_url( plugin_dir_url( __FILE__ ) )
-	->set_version( __FILE__ );
+	->set_version( __FILE__ )
+	->set_util();
```
kienstra, line 121: This is probably hacky how I set `->util` here. Here's a [little background](https://github.com/getblocklab/block-lab/pull/377#issuecomment-521481994).

Maybe we could eventually abstract all of these methods above, including `set_version()` and `set_util()`, into something like `Plugin::init( $file )` 

But I've probably taken this PR down the rabbit hole enough already 😄

lukecarbis:
✅ Looks good to me.

Need your review to merge.

php/class-plugin-abstract.php
kienstra: Thanks!

kienstra (approved):
**Approved**

Hi @lukecarbis,
Thanks for your great work and patience. This looks good!

kienstra:
Like with any big refactoring PR, the plugin could probably use a full regression test after this.

block-lab.php
kienstra: Ah, there's a new `Plugin::init()` method in https://github.com/getblocklab/block-lab/pull/269/files#diff-f162295a163d95e718f07b67b57ff1e1R40

lukecarbis: So we should move the `->set_util()` line to the `init` method in the onboarding PR, once this has been merged.

kienstra: Yeah, good point.

PR #382, opened 2019-08-06
Conditional blocks based on post types

lukecarbis:


php/post-types/class-block-post.php
```diff
@@ -895,6 +900,15 @@ public function save_block( $data ) {
 		check_admin_referer( 'block_lab_save_fields', 'block_lab_fields_nonce' );
 		check_admin_referer( 'block_lab_save_properties', 'block_lab_properties_nonce' );
 
+		// Update post meta with allowed post types.
+		if ( isset( $_POST['block-post-types'] ) ) {
+			$post_types = sanitize_text_field(
+				wp_unslash( $_POST['block-post-types'] )
+			);
+			$post_types = explode( ',', $post_types );
+			update_post_meta( $post_id, 'block-post-types', $post_types );
```
lukecarbis: @kienstra I chose here to use post meta to store which post types a block should appear in (or not).

The alternative is to extend the `Block` class with a new property for post types, in which case the accepted post types would be stored as JSON in the post content.

Which do you think is a better approach?

lukecarbis: I think I just answered my own question. It's got to be an extension of Block, so that Export / Import continues to work properly.

lukecarbis: In addition, rather than saving "allowed post types", we might need to save "disallowed post types".

Use case:
- I create a new block, and don't change any "Post Type" settings, so that the new block works in all post types
- The block saves, storing all current post types as allowed
- I create a new post type, and attempt to add my block
- The block doesn't exist for this post type, since it hasn't been explicitly set
- In order to get my blocks to show for the new post type, I need to edit every block and add the post type

lukecarbis:
@RobStino @kienstra Ready for you to review.

RobStino (approved):
Dude. This is so cool and so needed for a project I'm working on.
So steps I took to test:
- Create a custom post type (for testing)
- Create a custom block and set it to only be for posts (not pages or the new CPT)
- Try to add the block to a new post of each of the 3 post types
- Observed I could only add it to the post type Post
- Created another new CPT
- Observed that the block was now also available to this new CPT
Everything went as expected.

I know we talked about it in today's call, but my only thought is that when it comes to the include/exclude thing, what about:
- When a block is active for all post types, new post types are added automatically
- When a block is _inactive_ for at least one post type, new post types _are not_ added automatically.

I don't think it's that big a deal TBH. My thoughts just come from the idea that blocks that are available to all, you will most likely want available to any new CPTs. And for blocks that are not available to all, are probably built specifically for specific post types, so perhaps auto-adding doesn't make as much sense.

Approving as this does exactly what we set out to do.

lukecarbis:
> - When a block is active for all post types, new post types are added automatically
> - When a block is inactive for at least one post type, new post types are not added automatically.

I agree that this would be really nice, but it's extremely difficult to do. Plus, I'm slightly concerned that it would be unpredictable behaviour, and could cause confusion. "Why did it do this thing, when last time it did that thing?"

kienstra (approved):
**Approved**

Hi @lukecarbis,
This looks good.

As expected, on disabling a post type in its 'Publish' panel, it's not available for that post type:

no-on-page

PR #383, opened 2019-08-14
Remember default value of form field in the Block Builder

### LINKED ISSUE #378: Second time "Add Field" does not update slug for that field when pressed CTRL+V
Issue: When creating a block, when I click on "Add Field" button, if it's the first field, then whenever I paste something into the "Field Label" textbox, it automatically updates the "slug" for that field. But if it's the second field (second time adding a field), then pasting into the "Field Label" does not update the "slug". 

How to replicate:

Go to "add new block" page.
Click "Add field" button.
Paste something into "Field label" textbox.
Then click "Add field" button again.
Then again paste something into the second field's "Field Label" textbox.
- It won't update the slug for the second field.

Expected behavior:

Second field should update its slug just like the first field.

Thanks.

lukecarbis:
This is so that we can keep autoslugging on if the default value hasn't changed.

**Before**

Before

**After**

After

Closes #378.

### DIFF
--- a/js/admin.block-post.js
+++ b/js/admin.block-post.js
@@ -22,7 +22,7 @@
 				edit     = field.find( '.block-fields-actions-edit' ),
 				label    = field.find( '.block-fields-edit-label input' );
 
-				$( '.block-fields-rows' ).append( field );
+			$( '.block-fields-rows' ).append( field );
 			$( '.block-no-fields' ).hide();
 
 			edit.trigger( 'click' );
--- a/js/admin.block-post.js
+++ b/js/admin.block-post.js
@@ -18,11 +18,16 @@
 		$( '#block-add-field' ).on( 'click', function() {
 			let template = wp.template( 'field-repeater' ),
 				data     = { uid: new Date().getTime() },
-				field    = $( template( data ) );
-			$( '.block-fields-rows' ).append( field );
+				field    = $( template( data ) ),
+				edit     = field.find( '.block-fields-actions-edit' ),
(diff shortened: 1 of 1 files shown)


js/admin.block-post.js
```diff
@@ -18,11 +18,16 @@
 		$( '#block-add-field' ).on( 'click', function() {
 			let template = wp.template( 'field-repeater' ),
 				data     = { uid: new Date().getTime() },
-				field    = $( template( data ) );
-			$( '.block-fields-rows' ).append( field );
+				field    = $( template( data ) ),
+				edit     = field.find( '.block-fields-actions-edit' ),
+				label    = field.find( '.block-fields-edit-label input' );
+
+				$( '.block-fields-rows' ).append( field );
```
kienstra: Maybe this should be 1 tab to the left, not being part of the `let` block above.

kienstra:
**Approved**

Hi @lukecarbis,
This looks good! 

Like you showed in the great `.gif`s [above](https://github.com/getblocklab/block-lab/pull/383#issue-307529299), I also saw how when you add a field, don't change the 'Field Label,' add another field, and return to the earlier field, it now auto-slugs the 'Field Label.'

kienstra (dismissed):

js/admin.block-post.js
lukecarbis: Nice pickup. Thanks. Fixed in 57d161c.

kienstra (approved):

kienstra:
Hi @lukecarbis, thanks, this looks great!

PR #384, opened 2019-08-15
Copy the set_util() method into init(), add helper method for post type

kienstra:
* To follow the pattern of the `Plugin::init()` method, this moves the contents of `->set_util()` into that method, and deletes `->set_util()`
* Also, it adds a helper method `Plugin::get_post_type_slug()`
* This looks to remove the need to add `$block_post` as a property of Plugin
* This prevents `block_lab()->block_post` from essentially being in the global scope, so now people can't call something like `block_lab()->block_post->enqueue_scripts()` or `->add_caps()`

### DIFF
--- a/php/class-plugin-abstract.php
+++ b/php/class-plugin-abstract.php
@@ -54,6 +54,14 @@ abstract class Plugin_Abstract implements Plugin_Interface {
 	protected $slug;
 
 	/**
+	 * The slug of the post type that stores the blocks.
+	 *
+	 * @since 1.3.5
+	 * @var string
+	 */
+	protected $post_type_slug = 'block_lab';
+
+	/**
 	 * URL to the main plugin directory.
 	 *
 	 * @since 1.0.0
@@ -68,13 +76,6 @@ abstract class Plugin_Abstract implements Plugin_Interface {
 	 * @var string
 	 */
 	protected $version;
-
-	/**
-	 * Utility methods.
-	 *
-	 * @var Util
-	 */
-	protected $util;
 
 	/**
 	 * Allows calling methods in the Util class, directly in this class.
@@ -183,6 +184,15 @@ abstract class Plugin_Abstract implements Plugin_Interface {
 	}
 
 	/**
+	 * Gets the slug of the post type that stores the blocks.
+	 *
+	 * @return string The slug.
+	 */
+	public function get_post_type_slug() {
+		return $this->post_type_slug;
+	}
+
+	/**
 	 * Set the plugin's slug.
 	 *
 	 * @param string $slug The slug.
@@ -245,14 +255,6 @@ abstract class Plugin_Abstract implements Plugin_Interface {
 		};
 
 		return $this;
-	}
-
-	/**
-	 * Instantiates the Util class, with utility methods.
-	 */
-	public function set_util() {
-		$this->util = new Util();
-		$this->register_component( $this->util );
 	}
 
 	/**
--- a/php/class-plugin-abstract.php
+++ b/php/class-plugin-abstract.php
@@ -54,14 +54,6 @@ abstract class Plugin_Abstract implements Plugin_Interface {
 	protected $slug;
 
 	/**
-	 * The slug of the post type that stores the blocks.
-	 *
-	 * @since 1.3.5
-	 * @var string
-	 */
-	protected $post_type_slug = 'block_lab';
-
-	/**
 	 * URL to the main plugin directory.
 	 *
 	 * @since 1.0.0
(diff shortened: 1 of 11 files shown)


php/class-plugin-abstract.php
```diff
@@ -53,6 +53,14 @@ abstract class Plugin_Abstract implements Plugin_Interface {
 	 */
 	protected $slug;
 
+	/**
+	 * The slug of the post type that stores the blocks.
+	 *
+	 * @since 1.3.5
+	 * @var string
+	 */
+	protected $post_type_slug = 'block_lab';
```
lukecarbis: Do you think this belongs in the plugin abstract? I imagine it would go in the `Plugin` class, not the abstract version of it (which to me, should be portable to different plugins).

php/class-plugin-abstract.php
```diff
@@ -182,6 +183,15 @@ public function get_slug() {
 		return $this->slug;
 	}
 
+	/**
+	 * Gets the slug of the post type that stores the blocks.
+	 *
+	 * @return string The slug.
+	 */
+	public function get_post_type_slug() {
+		return $this->post_type_slug;
```
lukecarbis: This could be moved to the `Util` class?

Inside `Util`, it could just do this:
```
return $this->plugin->post_type_slug;
```

OR we could keep the post type slug in `Block_Post`, and do this instead:

```
return $this->plugin->block_post->slug;
```

The `plugin` property is set automatically when registering any component.

php/class-plugin.php
```diff
@@ -13,38 +13,30 @@
...
 
 	/**
-	 * Block Post Type.
+	 * Utility methods.
 	 *
-	 * @var Post_Types\Block_Post
+	 * @var Util
 	 */
-	public $block_post;
+	protected $util;
```
lukecarbis, line 22: I like this being protected.

php/class-plugin.php
```diff
@@ -13,38 +13,30 @@
...
-		$this->set_util();
-
-		$this->block_post = new Post_Types\Block_Post();
-		$this->register_component( $this->block_post );
+		$this->util = new Util();
+		$this->register_component( $this->util );
 
-		$this->loader = new Blocks\Loader();
-		$this->register_component( $this->loader );
+		$this->register_component( new Post_Types\Block_Post() );
```
lukecarbis, line 46: Nice work removing the `block_post` and `loader` properties.

lukecarbis (changes requested):

php/class-plugin.php
kienstra: Thanks!

php/class-plugin-abstract.php
kienstra: Ah, good point. Abstract should apply to any plugin.

php/class-plugin.php
kienstra: Thanks 😄

php/class-plugin-abstract.php
kienstra: Great idea, this is now moved to `Util` in https://github.com/getblocklab/block-lab/pull/384/commits/4614c78849587a6a2dc0b4ab04e287691212979b

php/class-plugin-abstract.php
kienstra: Good idea, https://github.com/getblocklab/block-lab/pull/384/commits/4614c78849587a6a2dc0b4ab04e287691212979b moves this to `Plugin`.

php/class-plugin.php
```diff
@@ -13,38 +13,38 @@
...
 	/**
-	 * Initiate the loading of new blocks.
+	 * The slug of the post type that stores the blocks.
 	 *
-	 * @var Blocks\Loader
+	 * @since 1.3.5
+	 * @var string
 	 */
-	public $loader;
+	public $post_type_slug = 'block_lab';
```
lukecarbis, line 37: What do you think about just calling this `$slug` instead of `$post_type_slug` – purely for brevity?

lukecarbis (approved):
Nice work!

php/class-plugin.php
kienstra: Yeah, that would be a little simpler. The reasoning was to distinguish it from the `$slug` property of `Plugin_Abstract`. 

Though both of them end up being `'block_lab'`

lukecarbis: Aah! Good point. Where is that used?

lukecarbis: Might be better to just go with the `Plugin_Abstract->$slug` property. The post type can just use that too.

kienstra: Though maybe `Util::get_post_type_slug()` could simply return `Plugin::$slug`, which also happens to be `'block_lab'`

kienstra: Ah, my fault. `Plugin::$slug` is `'block-lab'`, while `Plugin::$post_type_slug` is `'block_lab'`

kienstra:
Thanks, Luke!

PR #387, opened 2019-08-19
Helper functions for the repeater fields

lukecarbis:
This PR adds the following helper functions:

- `block_rows()`
- `block_row()`
- `block_sub_field()`
- `block_sub_value()`


lukecarbis:
@kienstra Ready for you to review.

php/blocks/class-loop.php
```diff
@@ -0,0 +1,92 @@
...
+	 * @var string
+	 */
+	public $active;
+
+	/**
+	 * Set a loop to active.
+	 *
+	 * @param string $name The field name.
+	 */
+	public function active( $name ) {
```
kienstra: OK, nice. It looks like only one row can be active at a time, though there might be a few rows stored in `$this->loops`.

php/blocks/class-loop.php
```diff
@@ -0,0 +1,92 @@
...
+ * @copyright Copyright(c) 2019, Block Lab
+ * @license http://opensource.org/licenses/GPL-2.0 GNU General Public License, version 2 (GPL-2.0)
+ */
+
+namespace Block_Lab\Blocks;
+
+/**
+ * Class Loop
+ */
+class Loop {
```
kienstra, line 15: It's nice how this clas follows the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) really well. It only handles the loop, and doesn't have to know about what it's looping through.

kienstra (approved):
**Approved**

Hi @lukecarbis,
Good work! This looks good, and will help a lot with the progress on the `feature/repeater` branch.

kienstra:
For reference, here's how to use these repeater helper functions in a template:

```php
$repeater_name = 'my-repeater';
if ( block_rows( $repeater_name ) ) :
	while ( block_rows( $repeater_name ) ) :
		block_row( $repeater_name );
		
		block_sub_field( 'your-example-field' );
	endwhile;
endif;

```

PR #395, opened 2019-08-22
Fix an issue in removing repeater rows

kienstra:
* Checks that `attribute` is truthy before accessing a property
* There was a console error from running `removeRow()`
* This blocked actually removing it


kienstra:
**Request For Review**

Hi @lukecarbis,
Could you please review this?

lukecarbis (approved):
Works nicely! Thanks.

PR #398, opened 2019-08-24
Prevent repeated requests to validate the Pro license

### LINKED ISSUE #397: Failed licensing check (maybe?) causing website to hang
Our development environments were seeing super long load times and timeouts this morning. New Relic monitoring revealed external hits to getblocklab.com taking 40,000ms. The Block Lab website appears to be down today. Looking at source, I'm guessing the plugin is trying to re-validate my license, but the super-long timeout is causing the website to hang.

Disabling the Block Lab plugin (via CLI) resolved the issue.

kienstra:
# Before
When the [license transient](https://github.com/getblocklab/block-lab/blob/e71150f505c9da18096742a044abf1d56c0648d5/php/admin/class-license.php#L90) expires, if somehow the [license validation POST request](https://github.com/getblocklab/block-lab/blob/e71150f505c9da18096742a044abf1d56c0648d5/php/admin/class-license.php#L120) to https://getblocklab.com fails (like with a 404), this plugin will make a few request to that URL each page load to validate the license. 

# After
If the POST request fails, it's stored in the transient, and there won't be another request for 24 hours. Users can still trigger a validation request by clicking 'Save Changes':

save-changes-here

# Steps to reproduce
1. Ensure that a valid Pro key is entered and saved in `/wp-admin` > Block Lab > Settings
2. Simulate the validation POST request failing, by setting the URL of the request to one that doesn't exist:

```diff
diff --git a/php/admin/class-license.php b/php/admin/class-license.php
index 66b79df..68e4325 100644
--- a/php/admin/class-license.php
+++ b/php/admin/class-license.php
@@ -43,7 +43,7 @@ class License extends Component_Abstract {
         * Initialise the Pro component.
         */
        public function init() {
-               $this->store_url    = 'https://getblocklab.com';
+               $this->store_url    = 'https://nonexistent-site-here.com';
                $this->product_slug = 'block-lab-pro';
```

3. Delete the transient:

```sh
wp transient delete block_lab_license
```
(Normally this would validate licenses about every 24 hours, once the transient expires and there's a request to WordPress. This simulates the 24 hours expiring by deleting the transient).

4. Go to the front-end of a post (not the block editor). This should cause the validation request to run, as the transient won't be set.

5. On this line, set a debug breakpoint in your IDE:

https://github.com/getblocklab/block-lab/blob/e71150f505c9da18096742a044abf1d56c0648d5/php/admin/class-license.php#L120

6. Like before, go to the front-end of any post (not the block editor)

7. Expected: there is no call to `wp_remote_post()`, as the previous one failed

8. Actual: There are 2 calls to `wp_remote_post()`, the IDE breakpoint will run twice
 
Closes #397


php/admin/class-license.php
```diff
@@ -120,17 +128,18 @@ public function activate_license( $key ) {
 		$response = wp_remote_post(
 			$this->store_url,
 			array(
-				'timeout'   => 15,
+				'timeout'   => 10,
```
kienstra, line 131: This seemed a little high, compared to the default of `5`. But maybe it's appropriate.

kienstra:
**Request For Review**

Hi @lukecarbis,
Could you please review this PR for #397? Thanks!

lukecarbis (approved):
Thanks for putting this together Ryan. Tested and it all looks good.

kienstra:
Thanks, @lukecarbis! While this PR addresses the main issue, I should also add an admin notice on the license page for the request failing. 

This should also help us for debugging support requests. As you know, it's very different when the `POST` request for license validation fails with a `404`, and when it returns that the license is invalid.

PR #402, opened 2019-08-26
Add integration and unit tests

kienstra:
- [x] Add an [integration test](https://github.com/getblocklab/block-lab/pull/402/files#diff-65847bbff742889cb73a7e018e683093) for all of the fields.
- [x] Make the unit tests phpcs-compliant, and remove the exclusion of them in `phpcs.xml`
- [x]  Test all of the fields except for the repeater. 
- [x] Ensure that when they are present as attributes, they work as expected in the block template.

In a future PR
* Add an integration test for the repeater.
* Add more unit tests.


tests/php/test-helpers.php
```diff
@@ -1,129 +0,0 @@
-<?php
```
kienstra, line 1: This is simply moved to [here](https://github.com/getblocklab/block-lab/pull/402/files#diff-beb131e068d21ed66ea78e959245b576R1).

kienstra:
**Request For Review**

Hi @lukecarbis,
Could you please review this? Ideally it would have an integration test for the repeater, and more unit tests.

But I'd like to switch to working on some Repeater issues, like those [reported on Slack](https://block-lab.slack.com/archives/CJ31X6NPN/p1566928925078700).

This PR should help to test PHP regressions in fixes for those.

Almost all of the diff here is from the `tests/` directory and config files like `phpunit.xml.dist`.

composer.json
```diff
@@ -9,6 +9,7 @@
     },
     "require-dev": {
         "brain/monkey": "^2",
-        "squizlabs/php_codesniffer": "^3.4"
+        "squizlabs/php_codesniffer": "^3.4",
+        "wp-coding-standards/wpcs": "2.1.1"
```
kienstra, line 13: This ensures there isn't an error for the WP standards not being installed. They might be globally installed, but I think it's best to have them here, to have a common standard.

php/class-util.php
```diff
@@ -11,7 +11,6 @@
 namespace Block_Lab;
 
 use Block_Lab\Blocks;
-use Block_Lab\Component_Abstract;
```
kienstra, line 14: This is in the same namespace, so it looks like it's not necessary to `use` it.

lukecarbis:
@kienstra This is fantastic! It would be great if at some point the integration test also tested for blank values, and default values.

lukecarbis (approved):

kienstra:
>@kienstra This is fantastic! It would be great if at some point the integration test also tested for blank values, and default values.

Thanks, @lukecarbis! Good point, that's now mentioned in #404.

PR #406, opened 2019-08-29
Add a different admin notice for when the validation request fails

kienstra:
A follow-up to #398 

This displays a message for the case that we saw on Saturday, where the validation request failed (like with a 404).

support-request-failed

>There was a problem activating the license, but it may not be invalid. Please either try again or contact support, mentioning this message.

However, this will only display when the user clicks 'Save Changes' on the Block Lab Settings page below. It won't appear when this makes the daily request to `getblocklab.com` to continue verifying the license. 

That's probably a more common case, as users probably only save their license once, where it makes another request every day.

This is a different case from it being an invalid license, and I think it  should have different feedback. It means our site might be down, and we should have that information right away.

# Steps to test
1. Simulate the validation POST request failing, by setting the URL of the request to one that doesn't exist:

```diff
diff --git a/php/admin/class-license.php b/php/admin/class-license.php
index 66b79df..68e4325 100644
--- a/php/admin/class-license.php
+++ b/php/admin/class-license.php
@@ -43,7 +43,7 @@ class License extends Component_Abstract {
         * Initialise the Pro component.
         */
        public function init() {
-               $this->store_url    = 'https://getblocklab.com';
+               $this->store_url    = 'https://nonexistent-site-here.com';
                $this->product_slug = 'block-lab-pro';
```

2. Delete the transient:

```sh
wp transient delete block_lab_license
```

3. Click 'Save Changes' for the license:

support-request-failed

### DIFF
--- a/php/admin/class-license.php
+++ b/php/admin/class-license.php
@@ -177,10 +177,15 @@ class License extends Component_Abstract {
 	 */
 	public function license_request_failed_message() {
 		$message = sprintf(
-			/* translators: %s is the email link for support */
-			__( 'There was a problem activating the license, but it may not be invalid. If the problem persists, please <a href="%s">contact support</a>.', 'block-lab' ),
-			'mailto:hi@getblocklab.com?subject=There was a problem activating my Block Lab Pro license'
+			/* translators: %s is an HTML link to contact support */
+			__( 'There was a problem activating the license, but it may not be invalid. If the problem persists, please %s.', 'block-lab' ),
+			sprintf(
+				'<a href="%1$s">%2$s</a>',
+				'mailto:hi@getblocklab.com?subject=There was a problem activating my Block Lab Pro license',
+				esc_html__( 'contact support', 'block-lab' )
+			)
 		);
+
 		return sprintf( '<div class="notice notice-error"><p>%s</p></div>', wp_kses_post( $message ) );
 	}
 
--- a/php/admin/class-license.php
+++ b/php/admin/class-license.php
@@ -176,8 +176,12 @@ class License extends Component_Abstract {
 	 * @return string
 	 */
 	public function license_request_failed_message() {
-		$message = __( 'There was a problem activating the license, but it may not be invalid. Please either try again or contact support, mentioning this message.', 'block-lab' );
-		return sprintf( '<div class="notice notice-error"><p>%s</p></div>', esc_html( $message ) );
+		$message = sprintf(
+			/* translators: %s is the email link for support */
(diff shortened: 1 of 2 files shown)


kienstra:
**Question About Notice**

Hi @lukecarbis,
Good to see you earlier.

1. What do you think about the notice shown above?

>There was a problem activating the license, but it may not be invalid. Please either try again or contact support, mentioning this message.

2. Do you think the message should have something like 'contact support,' and if so, what URL should that have?

Thanks, talk to you soon.

lukecarbis:
How about this:

> There was a problem activating the license, but it may not be invalid. If the problem persists, please [contact support](mailto:hi@getblocklab.com?subject=There was a problem activating my Block Lab Pro license).

kienstra:
Wow, great idea. Especially the subject line.

kienstra:
Hi @lukecarbis, how does this look, with your message used verbatim:

settings-message

kienstra:
Also, when you have a chance, could you please review this PR?

php/admin/class-license.php
```diff
@@ -155,12 +167,29 @@ public function license_success_message() {
...
+	 *
+	 * This is for when the validation request fails entirely, like with a 404.
+	 * Not for when it returns that the license is invalid.
+	 *
+	 * @return string
+	 */
+	public function license_request_failed_message() {
+		$message = sprintf(
+			/* translators: %s is the email link for support */
+			__( 'There was a problem activating the license, but it may not be invalid. If the problem persists, please <a href="%s">contact support</a>.', 'block-lab' ),
```
lukecarbis: I know that there's a range of opinion on this, but I'd prefer to put the anchor tag in sprintf variables, like we do elsewhere in the plugin. I know core doesn't always stick to this rule, but I think it makes things much easier for translators. Optional, if you feel strongly against this.

php/admin/class-license.php
```diff
@@ -155,12 +167,29 @@ public function license_success_message() {
...
+			'mailto:hi@getblocklab.com?subject=There was a problem activating my Block Lab Pro license'
```
lukecarbis: It makes sense to leave the subject untranslated. 👍

lukecarbis (approved):
Looks good!

php/admin/class-license.php
kienstra: Haha, yeah.

php/admin/class-license.php
kienstra: Ah, good point. I'll apply that suggestion soon.

kienstra: How does https://github.com/getblocklab/block-lab/pull/406/commits/5c57e18bb39e9fb60991cde3aad4baad70fb3ab2 look?

PR #407, opened 2019-08-30
Add a field width control to blocks

lukecarbis:
This doesn't yet work for repeater rows, but it will be easy to add. I'd suggest merging this one in, and doing repeater later.

Please test on all controls! @RobStino could use your help for this!

Screen Shot 2019-08-30 at 4 00 35 pm

Screen Shot 2019-08-30 at 3 58 47 pm

### DIFF
--- a/js/blocks/components/fields.js
+++ b/js/blocks/components/fields.js
@@ -21,6 +21,22 @@ const getControl = ( field ) => {
 };
 
 /**
+ * Gets the class name for the field.
+ *
+ * @param {Object} field The field to get the class name of.
+ * @return {string} The class name.
+ */
+const getClassName = ( field ) => {
+	let className = 'block-lab-control';
+
+	if ( field.width ) {
+		className += ' width-' + field.width;
+	}
+
+	return className;
+};
+
+/**
  * Renders the fields, using their control functions.
  *
  * @param {Array}  fields           The fields to render.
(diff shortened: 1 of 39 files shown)


RobStino:
This is great Luke!
I've just started testing and noticed that the width controls aren't available on the image field. Is this just because you're yet to get to that one?

RobStino:
What do folk think about adding 33% / 66% in there?

js/blocks/components/fields.js
```diff
@@ -20,6 +20,22 @@ const getControl = ( field ) => {
 	return loadedControls[ field.control ];
 };
 
+/**
+ * Gets the class name for the field.
+ *
+ * @param {Object} field The field to get the class name of.
+ * @return {string} The class name.
+ */
+const getClassName = ( field ) => {
```
kienstra, line 29: Nice DocBlock!

lukecarbis:
@RobStino I understand that having three-up columns would be nice, however I think the UI gets a bit crowded / overcomplicated. I'm disinclined to include the 33% and 67% options.

Screen Shot 2019-09-01 at 2 11 23 pm

Unless we found a different / better way of displaying the options? e.g. small / medium / large / full.

lukecarbis:
@RobStino I didn't include images at first because I wasn't sure it would look great with the droppable area… but after testing, I think it works.

Screen Shot 2019-09-01 at 2 25 45 pm

kienstra:
**Code Looks Good, So Does Display In Editor**
*Question About Display In Inspector*

Hi @lukecarbis,
The Editor display looks really good:

widths-in-editor

In the Inspector, it looks like it essentially ignores the widths. 

inspector-fields

Maybe the only issue with this would be that user would be confused about the widths not applying.

What do you think? Maybe users wouldn't be concerned about widths in the Inspector anyway, as its narrow.

lukecarbis:
@kienstra Great pickup! fabf0ee addresses this by hiding the field width settings when the location is not the editor.

kienstra:
Thanks, @lukecarbis! I'll look at that commit, and probably approve this soon.

kienstra:
Thanks, @lukecarbis. https://github.com/getblocklab/block-lab/commit/fabf0eefa2c9201ef6d7ef9d1810787eb88f32d5 looks good.

On selecting 'Inspector,' the width is hidden:

width-hidden

kienstra (approved):

PR #408, opened 2019-09-02
Repeater columns

lukecarbis:
For your consideration.

This is a good start, but we might want to tweak the UI a little. I'm not 100% happy with it.

Keen for your thoughts!

Screen Shot 2019-09-02 at 3 39 36 pm

Screen Shot 2019-09-02 at 2 27 04 pm

Screen Shot 2019-09-02 at 2 27 49 pm

Screen Shot 2019-09-02 at 2 28 27 pm

### DIFF
--- a/css/src/editor.scss
+++ b/css/src/editor.scss
@@ -205,12 +205,12 @@ div[class^="wp-block-block-lab-"] {
 
     .block-lab-repeater__rows {
       width: 100%;
-      display: flex;
+      display: grid;
       flex-wrap: wrap;
       margin: 1px 0 0 1px;
+      grid-template-columns: 100%;
 
       .block-lab-repeater--row {
-        flex-basis: 100%;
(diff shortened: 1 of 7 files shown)


RobStino:
I've noticed a couple of bugs in testing, but before we dive into fixing the small things, I'm just still thinking through the bigger idea.
I'm not 100% sure on it tbh. It doesn't feel as elegant as so much of the rest of our UX.
I'm doing some more looking around for ideas on how we approach this.
In general, I'm wondering if we should maintain the current implementation of basic *Rows* as the default, and then offer alternative display options (TBD). Even in this though, I'm not sure as I like our slightly opinionated approach to UX. Rather than heaps and heaps of options.
Thanks for working on this though @lukecarbis It helps move the idea along.

lukecarbis:
I think we should abandon this approach, in favour of something more like this: https://d.pr/v/mYxMiF

PR #409, opened 2019-09-02
Fix empty Repeater rows and default values

### LINKED ISSUE #393: Blank repeater rows don't show at all
If a repeater contains a blank row, we would expect it to be output with default values. However, it currently is not even looped through at all.

kienstra:
# WIP

## Before
* On adding empty rows, they did not appear at all in the editor
* This looks to be from the way `<ServerSideRender>` [handles empty {}](https://github.com/WordPress/gutenberg/blob/e2490612decf2f0ac614d6ee9f6d1269d0bfc615/packages/server-side-render/src/server-side-render.js#L22)
* Default values were only applied to the first row

## After 
* This adds a workaround for that, setting the new row to `{ '': '' }`
* This still isn't ideal, and I should keep an eye out for a better approach
* Default values are now applied to all repeater rows, even for completely empty rows

Fixes #393

### DIFF
--- a/js/blocks/components/edit.js
+++ b/js/blocks/components/edit.js
@@ -7,7 +7,7 @@ const { Fragment } = wp.element;
 /**
  * Internal dependencies
  */
-import { AdvancedControls, BlockLabInspector, Fields, FormControls } from './';
+import { AdvancedControls, BlockLabInspector, FormControls } from './';
 import icons from '../../../assets/icons.json';
 
 /**
--- a/js/blocks/components/repeater-rows.js
+++ b/js/blocks/components/repeater-rows.js
@@ -129,26 +129,7 @@ import { Fields } from './';
 			 * Without this, it looks like setAttributes() doesn't recognize a change to the array, and the component doesn't re-render.
 			 */
 			const rows = repeaterRows.slice();
-
-			/*
-			 * Ensure that every row has the required attributes, so that we don't lose blank rows.
-			 */
-			for ( name in this.props.subFields ) {
-				rows.forEach( ( row ) => {
-					if ( ! row.hasOwnProperty( name ) ) {
-						row[ name ] = null;
-					}
-				} );
-			};
-
-			rows.splice(
-				to,
-				0,
-				rows.splice(
-					from,
-					1
-				)[0]
-			);
+			[ rows[ from ], rows[ to ] ] = [ rows[ to ], rows[ from ] ]
 
 			attr[ parentName ] = { rows };
 			parentBlockProps.setAttributes( attr );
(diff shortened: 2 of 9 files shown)


js/blocks/components/edit.js
```diff
@@ -7,7 +7,7 @@ const { Fragment } = wp.element;
 /**
  * Internal dependencies
  */
-import { AdvancedControls, BlockLabInspector, Fields, FormControls } from './';
+import { AdvancedControls, BlockLabInspector, FormControls } from './';
```
kienstra, line 10: I should have removed `Fields` earlier, it's not used in the file.

js/blocks/controls/repeater.js
```diff
@@ -10,14 +10,28 @@ const { BaseControl, IconButton } = wp.components;
 import { RepeaterRows } from '../components';
 
 const BlockLabRepeaterControl = ( props ) => {
-	const { field, getValue, onChange, parentBlock, parentBlockProps } = props;
+	const { field, onChange, parentBlock, parentBlockProps } = props;
```
kienstra, line 13: I should have removed `getValue` earlier, it's not used in the file.

js/blocks/controls/repeater.js
```diff
@@ -10,14 +10,28 @@ const { BaseControl, IconButton } = wp.components;
...
+	/**
+	 * Adds a new empty row, using { '': '' }.
+	 *
+	 * Simply using {} results in <ServerSideRender> not sending an empty row,
+	 * and the empty row isn't rendered in the editor.
+	 *
+	 * @see https://github.com/getblocklab/block-lab/issues/393
+	 */
+	const addEmptyRow = () => {
+		const withAddedRow = rows.concat( { '': '' } );
```
kienstra, line 30: This isn't ideal, but it's worked so far in brief testing. I'll keep testing this.

lukecarbis:
@kienstra I can confirm that this works great for adding and saving new blank rows.

I'm still experiencing this blank rows bug, I wonder if it's related:



Happy to create a new issue / PR to resolve this if it's sufficiently different, but I thought the issues might be tied together in some way.

kienstra:
>I'm still experiencing this blank rows bug, I wonder if it's related:

Ah, interesting. I'll look at that.

kienstra:
So far, it looks like this [problem](https://github.com/getblocklab/block-lab/pull/409#issuecomment-527242700) only exists with fields that have `<TextControl>`, like Text and Email. Though I'm not sure they're causing this.

kienstra:
**Checkbox Issue**

Hi @lukecarbis,
Good talking to you 😄 

I also saw the issue with the checkbox you mentioned, where it appears to be 'half-checked.'

https://github.com/getblocklab/block-lab/pull/409/commits/c55a103ca1c545a47ed20b3b695717c1c5de4839 fixed it in my testing, but let me know if you still see the issue.

Feel free to wait until a code review if you'd like. 

Thanks, Luke!

js/blocks/components/repeater-rows.js
```diff
@@ -129,26 +129,7 @@ import { Fields } from './';
...
-
-			rows.splice(
-				to,
-				0,
-				rows.splice(
-					from,
-					1
-				)[0]
-			);
+			[ rows[ from ], rows[ to ] ] = [ rows[ to ], rows[ from ] ]
```
lukecarbis, line 132: Wow – this is so so obvious. Can't believe how much I over-engineered this! Pickup of the week!

js/blocks/controls/email.js
```diff
@@ -4,6 +4,8 @@ const { TextControl } = wp.components;
 
 const BlockLabEmailControl = ( props ) => {
 	const { field, getValue, onChange } = props;
+	const initialValue = getValue( props );
+	const value = 'undefined' !== typeof initialValue ? initialValue : field.default;
```
lukecarbis, line 8: Great solution. Really glad to get this console error sorted.

lukecarbis (approved):
I messed around for quite a while with new blank rows, saved blank rows, moving blank rows around, etc. As far as I can see, this PR solves all issues with blank rows. Great work! Thank you!

js/blocks/components/repeater-rows.js
kienstra: Haha, I didn't know about that either. I just happened upon it from StackOverflow.

js/blocks/controls/email.js
kienstra: Thanks!

js/blocks/components/repeater-rows.js
kienstra: The previous solution actually worked fine.

js/blocks/controls/email.js
kienstra: Thanks!

kienstra:
Hi @lukecarbis, thanks a lot for reviewing and testing this. I'm going to merge it now if that's alright.

PR #412, opened 2019-09-03
Add a filter to allow getting more values from block_field()

kienstra:
* An alternative to #373 
* This will make it possible for `block_field()` to return or echo additional block attributes that another plugin might have added, like in this [example usage](https://github.com/phpbits/block-options/blob/57a752f703590ea719bb9c85fad1f3a0e97b1fa0/includes/helper.php#L10-L17) of this filter. 
* It's possible to add block attributes [by filtering](https://github.com/phpbits/block-options/blob/9c6aadfc8d7a5153d17acb0bb3630296cfa21c8b/src/extensions/attributes/index.js#L28) the JavaScript function `registerBlockType()`. For example, Editorskit [adds an editorskit object](https://github.com/phpbits/block-options/blob/9c6aadfc8d7a5153d17acb0bb3630296cfa21c8b/src/extensions/attributes/index.js#L28-L43) to the block attributes, and uses it [in the Inspector Controls](https://github.com/phpbits/block-options/blob/aad4e9aa0f49ba29d226ecbf7357f60e7863ed3a/src/extensions/advanced-controls/index.js#L101).
* But before, it wasn't possible to output these attributes with `block_field()`. That only allowed registered Block Lab fields and `className`.

### DIFF
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -25,30 +25,27 @@ function block_field( $name, $echo = true ) {
 	 */
 	global $block_lab_attributes, $block_lab_config;
 
+	$default_fields = array( 'className' );
+
 	if ( ! isset( $block_lab_attributes ) || ! is_array( $block_lab_attributes ) ) {
 		return null;
 	}
 
-	$default_fields = array( 'className' );
+	$is_name_allowed = isset( $block_lab_config->fields[ $name ] ) || in_array( $name, $default_fields, true );
 
 	/**
-	 * Filters the default fields that are allowed in addition to the fields in block attributes.
+	 * Filters whether the $name argument to this function is allowed.
 	 *
-	 * Adding an attribute to this can enable outputting it via block_field().
-	 * Normally, this function only returns or echoes Block Lab attributes (fields), and one default field.
-	 * But this allows getting block attributes that might have been added by other plugins.
-	 * To allow getting another attribute, add it to the $default_fields array.
+	 * By default, this only allows a $name that is in the block fields or default fields, like 'className'.
+	 * Returning true to this filter can prevent exiting from block_field()
+	 * when looking for another value in the block attributes.
 	 *
-	 * @param string[] $default_fields Default block fields to allow.
-	 * @param string   $name The name of value to get.
+	 * @param bool   $is_name_allowed Whether the name is allowed.
+	 * @param string $name The name of value to get.
 	 */
-	$default_fields = apply_filters( 'block_lab_default_fields', $default_fields, $name );
+	$is_block_field_name_allowed = apply_filters( 'is_block_field_name_allowed', $is_name_allowed, $name );
 
-	if ( ! is_array( $default_fields ) ) {
-		return null;
-	}
-
-	if ( ! isset( $block_lab_config->fields[ $name ] ) && ! in_array( $name, $default_fields, true ) ) {
+	if ( ! $is_block_field_name_allowed ) {
 		return null;
 	}
 
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -25,27 +25,30 @@ function block_field( $name, $echo = true ) {
 	 */
 	global $block_lab_attributes, $block_lab_config;
 
-	$default_fields = array( 'className' );
-
 	if ( ! isset( $block_lab_attributes ) || ! is_array( $block_lab_attributes ) ) {
 		return null;
 	}
 
-	$is_name_allowed = isset( $block_lab_config->fields[ $name ] ) || in_array( $name, $default_fields, true );
+	$default_fields = array( 'className' );
 
 	/**
-	 * Filters whether the $name argument to this function is allowed.
+	 * Filters the default fields that are allowed in addition to the fields in block attributes.
 	 *
-	 * By default, this only allows a $name that is in the block fields or default fields, like 'className'.
-	 * Returning true to this filter can prevent exiting from block_field()
-	 * when looking for another value in the block attributes.
+	 * Adding an attribute to this can enable outputting it via block_field().
+	 * Normally, this function only returns or echoes Block Lab attributes (fields), and one default field.
+	 * But this allows getting block attributes that might have been added by other plugins.
+	 * To allow getting another attribute, add it to the $default_fields array.
 	 *
-	 * @param bool   $is_name_allowed Whether the name is allowed.
-	 * @param string $name The name of value to get.
+	 * @param string[] $default_fields Default block fields to allow.
+	 * @param string   $name The name of value to get.
 	 */
-	$is_block_field_name_allowed = apply_filters( 'is_block_field_name_allowed', $is_name_allowed, $name );
+	$default_fields = apply_filters( 'block_lab_default_fields', $default_fields, $name );
 
-	if ( ! $is_block_field_name_allowed ) {
+	if ( ! is_array( $default_fields ) ) {
+		return null;
+	}
+
+	if ( ! isset( $block_lab_config->fields[ $name ] ) && ! in_array( $name, $default_fields, true ) ) {
 		return null;
 	}
 
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -32,12 +32,13 @@ function block_field( $name, $echo = true ) {
 	$default_fields = array( 'className' => 'string' );
 
 	/**
-	 * Filters the default fields that are allowed in addition to the fields in block attributes.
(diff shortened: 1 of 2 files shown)


lukecarbis (approved):
Nice work here. Great test write up as well.

php/helpers.php
```diff
@@ -25,13 +25,27 @@ function block_field( $name, $echo = true ) {
...
+	$default_fields = array( 'className' );
+
 	if ( ! isset( $block_lab_attributes ) || ! is_array( $block_lab_attributes ) ) {
 		return null;
 	}
 
-	$default_fields = array( 'className' );
+	$is_name_allowed = isset( $block_lab_config->fields[ $name ] ) || in_array( $name, $default_fields, true );
+
+	/**
```
lukecarbis, line 34: Should this be a single asterix, since it's an inline comment, not a doc block?

i.e.

```
/*
 * Filters whether the $name argument to this function is allowed.
 *
 * …
 */
```

php/helpers.php
```diff
@@ -25,13 +25,27 @@ function block_field( $name, $echo = true ) {
...
+	 * Filters whether the $name argument to this function is allowed.
+	 *
+	 * By default, this only allows a $name that is in the block fields or default fields, like 'className'.
+	 * Returning true to this filter can prevent exiting from block_field()
+	 * when looking for another value in the block attributes.
+	 *
+	 * @param bool   $is_name_allowed Whether the name is allowed.
+	 * @param string $name The name of value to get.
+	 */
+	$is_block_field_name_allowed = apply_filters( 'is_block_field_name_allowed', $is_name_allowed, $name );
```
lukecarbis: Why are we filtering the `is_name_allowed` result, instead of filtering the `default_fields` array?

php/helpers.php
kienstra: Ah, good question. 

The examples I've seen [in Core](https://github.com/WordPress/wordpress-develop/blob/00b03f2a6f2c21ebf9975c9c08976358681400ba/src/wp-includes/feed.php#L28) have had 2 asterisks, but I'm not sure if there's an official convention for this.

php/helpers.php
kienstra: Ah, that's a good idea. I didn't think of that, as the `$default_fields` array didn't exist when I originally wrote this filter in another branch, about a month ago.

lukecarbis: Ha! I only picked up on this because I thought you pulled me up on this a couple weeks back. :P

lukecarbis: Maybe just do whatever we are already doing elsewhere in the plugin.

kienstra: You make a good point, I think it's simpler to filter the `$default_fields`.

kienstra: And thanks to the simplicity, there's less risk of regression.

kienstra:
Jeffrey should probably also review and test this before we merge it.

lukecarbis:
@phpbits Could you take a look at this PR please?

kienstra:
@phpbits, sorry for the delay on this. The filter changed a little since we [discussed it](https://github.com/phpbits/block-options/blob/57a752f703590ea719bb9c85fad1f3a0e97b1fa0/includes/helper.php#L10-L17).

Here's how to use it now:

```php

/**
 * Adds 'editorskit' to the default fields, allowing it to be passed to block_field().
 *
 * @param array $default_fields The default fields.
 */
add_filter( 'block_lab_default_fields', function( $default_fields ) {
	$default_fields[] = 'editorskit';
	return $default_fields;
} );

```

phpbits:
@kienstra the new one doesn't work but the filter I've tested before is working perfectly. The problem is that `block_field( 'editorskit', false )` returns string `Array` instead of the array values. Thanks!

kienstra:
Hi @phpbits,
Thanks for testing this.

You're right, it looks like it's returning the string `'Array'`:

https://github.com/getblocklab/block-lab/blob/60e63eb6549a7fa1693d64e619278ae422de4b4f/php/helpers.php#L51-L53

I'll keep working on this.

phpbits:
Thanks a lot @kienstra!

kienstra:
**Question About Approach**

Hi @lukecarbis,
Do you think the current state of this PR is alright?

The idea you had of filtering the `$default_fields` was really good.

But it looks like there's something further down the function that prevents this from working. I only noticed that when Jeffrey brought this up.

In the case of Editorskit, it would add a `'editorskit'` to the `$default_fields` via a filter.

But further down in the function, it casts all `$default_fields` into a string, even though `editorskit` is an array:

https://github.com/getblocklab/block-lab/blob/60e63eb6549a7fa1693d64e619278ae422de4b4f/php/helpers.php#L51-L53

Maybe there's a better way to approach this, though.

I reverted all of the commits to this PR, except [the first](https://github.com/getblocklab/block-lab/pull/412/commits/77d01f641e8a865dfd0d65eeaf034a94895482cd).

It'd be good to have this PR merged in the next release, but only if we get this approach right, and Jeffrey gets a chance to test it.

Thanks, Luke!

lukecarbis:
@kienstra I must be missing something – where is `$default_fields` cast as a string?

lukecarbis:
@kienstra Would it work if we instantiated each of the default fields as a `Block_Lab\Blocks\Field`? That way we could call `cast_value` against it, instead of forcing it to be a string.

e.g.
```PHP
$default_fields = apply_filters(
    'block_lab_default_fields',
   array(
       'className' => new Field(
            array(
                'name' => 'className',
                'type' => 'string'
            )
        )
    )
);
```

and later…

```PHP
} elseif ( array_key_exists( $name, $default_fields, true ) ) { 
    // Cast default Editor attributes appropriately. 
    $value = $default_fields[ $name ]->cast_value( $value ); 
}
```

kienstra:
Hi @lukecarbis,
That's a good idea.

The last 2 commits have a slightly modified version of your idea. The `block_field()` method takes care of instantiating a `Field`, as an external plugin would have to do `new Block_Lab\Blocks\Field()`.

Now, adding a `$default_field` would look like:

```php
/**
 * Adds 'editorskit' to the default fields, allowing it to be passed to block_field().
 *
 * @param array $default_fields The default fields.
 */
add_filter( 'block_lab_default_fields', function( $default_fields ) {
	$default_fields['editorskit'] = 'array'; // The name and type.

	return $default_fields;
} );
```

php/helpers.php
```diff
@@ -29,9 +29,22 @@ function block_field( $name, $echo = true ) {
...
-	if ( ! isset( $block_lab_config->fields[ $name ] ) && ! in_array( $name, $default_fields, true ) ) {
+	/**
+	 * Filters the default fields that are allowed in addition to the fields in block attributes.
+	 *
+	 * Adding an attribute to this can enable outputting it via block_field().
+	 * Normally, this function only returns or echoes Block Lab attributes (fields), and one default field.
+	 * But this allows getting block attributes that might have been added by other plugins.
+	 * To allow getting another attribute, add it to the $default_fields array.
+	 *
+	 * @param array  $default_fields An associative array of $field_name => $field_type.
```
kienstra, line 43: This could probably use more description of the types that are possible, like string, array, etc...

lukecarbis (approved):
These changes look great, @kienstra! Nice work. Please merge whenever you and @phpbits are ready.

kienstra:
**Request To Test**

Hi @phpbits,
Sorry to keep changing this, and hope you've been doing great.

When you have a chance, could you please test this PR?

Here's how to use the new filter:

```php
/**
 * Adds 'editorskit' to the default fields, allowing it to be passed to block_field().
 *
 * @param array $default_fields An associative array of $field_name => $field_type.
 */
add_filter( 'block_lab_default_fields', function( $default_fields ) {
	$default_fields['editorskit'] = 'array'; // The name and type.

	return $default_fields;
} );
```

phpbits:
@kienstra My apology for late response. Just tested it and working perfectly 💯 Thanks a lot!

kienstra:
Thanks, @phpbits! Sorry for all of the changes here.

kienstra:
**Going To Merge If That's Alright**

Hi @lukecarbis,
If it's alright, I'm going to merge this soon. 

@phpbits kindly tested this again, and I made a [comment-only commit](https://github.com/getblocklab/block-lab/pull/412/commits/669439f463e4ef2177b2b268549766e0a76449c4) to make the DocBlock a little more clear.

It'd be good to have this in production soon.

Thanks, Luke!

PR #414, opened 2019-09-04
Fix an issue with the Textarea in the Repeater

kienstra:
# Before
Like the issue we saw in #409, if a Repeater has a Textarea with a value, and it's moved, it can replace the value of a blank Textarea:

mo-repeater

# After 
The empty Textarea remains empty, as expected


kienstra:
Hi @lukecarbis,
Could you please review this? It looks like I missed this 😄 

Thanks!

lukecarbis (approved):

PR #415, opened 2019-09-04
Fix a license validation issue for 'lifetime'

kienstra:
# WIP

* Account for 'lifetime' license values
* Instead of an 'expires' time, it looks like the EDD API sometimes returns 'lifetime' for that type of license.

### DIFF
--- a/php/admin/class-license.php
+++ b/php/admin/class-license.php
@@ -95,7 +95,15 @@ class License extends Component_Abstract {
 		$license = $this->get_license();
 
 		if ( isset( $license['license'] ) && 'valid' === $license['license'] ) {
-			if ( isset( $license['expires'] ) && time() < strtotime( $license['expires'] ) ) {
+			if (
+				isset( $license['expires'] )
+				&&
+				(
+					'lifetime' === $license['expires']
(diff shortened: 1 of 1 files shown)


kienstra:
This looks to only apply to a certain kind of license that only developers are using. I'm closing this, for possible review later.

php/admin/class-license.php
```diff
@@ -95,7 +95,15 @@ public function is_valid() {
 		$license = $this->get_license();
 
 		if ( isset( $license['license'] ) && 'valid' === $license['license'] ) {
-			if ( isset( $license['expires'] ) && time() < strtotime( $license['expires'] ) ) {
+			if (
```
lukecarbis: What do you think about turning this into two if statements, for readability:

```
if ( isset( $license['license'] ) && isset( $license['expires'] && 'valid' === $license['license'] ) ) {
    if ( 'lifetime' === $license['expires'] ) {
        return true;
    }
    if ( time() < strtotime( $license['expires'] ) ) {
        return true;
    }
}
```

lukecarbis (approved):

Issue #404, opened 2019-08-28
Continue with integration and unit tests


kienstra:
A follow-up to #402.

- [x] Add an integration test for the repeater.
- [ ] Add more unit tests (likely to be in several PRs, including #433)
- [x] In the [integration test](https://github.com/getblocklab/block-lab/blob/f0f59564d13802ef7185ef3d639bfe4ef1b25c04/tests/php/integration/test-template-output.php), also test for blank and default values (thanks to @lukecarbis for this [suggestion](https://github.com/getblocklab/block-lab/pull/402#issuecomment-525554382)).

PR #419, opened 2019-09-04
Add the validation filter to block_sub_field()

### LINKED ISSUE #418: block_sub_field returns image ID instead of URL
Walkthrough -> https://d.pr/v/DPXBPa

kienstra:
* Adds an integration test for the repeater, with 2 rows of all possible controls.
* Though it doesn't test default values yet.
* [Adds a filter](https://github.com/getblocklab/block-lab/pull/419/files#diff-018b57db3c29f17ff59f382617e39f14R75) similar to the one for `block_field()`
* That filter is how each control's `validate()` function is called when calling `block_sub_field()` or `block_field()`
* For example, for the Image control, [Image::validate()](https://github.com/getblocklab/block-lab/blob/60e63eb6549a7fa1693d64e619278ae422de4b4f/php/blocks/controls/class-image.php#L59) outputs a different value if `$echo` is true (the URL), versus when it's false (the ID).

Closes #418 
Partially fulfills #404 (Add an integration test for the repeater)

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -72,6 +72,7 @@ class Block_Post extends Component_Abstract {
 		add_action( 'wp_insert_post_data', array( $this, 'save_block' ), 10, 2 );
 		add_action( 'init', array( $this, 'register_controls' ) );
 		add_filter( 'block_lab_field_value', array( $this, 'get_field_value' ), 10, 3 );
+		add_filter( 'block_lab_sub_field_value', array( $this, 'get_field_value' ), 10, 3 );
 
 		// Clean up the list table.
 		add_filter( 'disable_months_dropdown', '__return_true', 10, $this->slug );
--- a/tests/php/integration/helpers/class-abstract-attribute.php
+++ b/tests/php/integration/helpers/class-abstract-attribute.php
@@ -1,0 +1,201 @@
+<?php
+/**
+ * Abstract_Attribute
+ *
+ * @package Block_Lab
+ */
+
+/**
+ * Class Abstract_Attribute
+ *
+ * @package Block_Lab
+ */
+abstract class Abstract_Attribute extends \WP_UnitTestCase {
+
+	/**
+	 * The block attributes.
+	 *
+	 * @var array
+	 */
+	public $attributes;
+
+	/**
+	 * All fields that return a string for block_value().
+	 *
+	 * @var array
+	 */
+	public $string_fields;
+
+	/**
+	 * All fields that return either an object or ID for block_value().
+	 *
+	 * @var array
+	 */
+	public $object_fields;
+
+	/**
+	 * The instance of Loader, to render the template.
+	 *
+	 * @var Blocks\Loader
+	 */
+	public $loader;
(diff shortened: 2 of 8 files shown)


kienstra:
This fixed the issue locally, but I should test it more.

tests/php/integration/helpers/class-abstract-attribute.php
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @package Block_Lab
+ */
+
+/**
+ * Class Abstract_Attribute
+ *
+ * @package Block_Lab
+ */
+abstract class Abstract_Attribute extends \WP_UnitTestCase {
```
kienstra, line 13: Most of this class is simply abstracted out of [test-template-output.php](https://github.com/getblocklab/block-lab/pull/419/files#diff-65847bbff742889cb73a7e018e683093L16).

kienstra:
**Request For Review**

Hi @lukecarbis,
Hope you're doing great, I kind of miss Australia 😄 

Could you please review this when you have a chance? Thanks, Luke!

php/post-types/class-block-post.php
```diff
@@ -72,6 +72,7 @@ public function register_hooks() {
 		add_action( 'wp_insert_post_data', array( $this, 'save_block' ), 10, 2 );
 		add_action( 'init', array( $this, 'register_controls' ) );
 		add_filter( 'block_lab_field_value', array( $this, 'get_field_value' ), 10, 3 );
+		add_filter( 'block_lab_sub_field_value', array( $this, 'get_field_value' ), 10, 3 );
```
lukecarbis, line 75: Great pickup!

lukecarbis (approved):
Australia misses you too!

This PR looks great. Here's the template I used to test:

```
if ( block_rows( 'repeater' ) ) {
	while ( block_rows( 'repeater' ) ) {
		block_row( 'repeater' );
		?>
		<p><?php echo block_sub_value( 'background' ); ?></p>
		<p><?php block_sub_field( 'background' ); ?></p>
		<?php
	}
}
reset_block_rows( 'repeater' );
```

Here's the result of that on `develop` (notice an unexpected value from `block_sub_field`): 
Screen Shot 2019-09-09 at 7 39 44 am

Here's the result of the same template on this branch (with the correct value for `block_sub_field`):
Screen Shot 2019-09-09 at 7 40 20 am

kienstra:
Hi @lukecarbis,
Haha, thanks.

Thanks a lot for testing this, also.

If it's alright, I just pushed https://github.com/getblocklab/block-lab/pull/419/commits/7e7af7d3064ce6bb9d927b7abc6a83899fc367a7, which adds an integration test for default values (but not yet for the repeater). It only touches files in `tests/php/`

lukecarbis:
7e7af7d looks good! Feel free to merge.

kienstra:
Thanks, @lukecarbis!

php/post-types/class-block-post.php
kienstra: Thanks!

PR #421, opened 2019-09-09
Add min and max row setting to repeater field

### LINKED ISSUE #399: Repeater row limit
There should be a row limit to repeaters. Use case: Adding buttons to the bottom of a hero section. e.g. I might have a hero that allows for _up to_ 2 buttons.

lukecarbis:
Hey @kienstra – thanks for reviewing this.

This PR forces a minimum and maximum amount of rows for the repeater.

The example below is a repeater with a min setting of 4, and a max of 6. Notice how it pre-populates the repeater with the minimum amount of rows. The "New Row" button is disabled when the maximum limit is reached, and the "delete" button is hidden when the minimum limit is reached.

@RobStino Please review this for UX.

Closes #399.

### DIFF
--- a/js/blocks/components/repeater-rows.js
+++ b/js/blocks/components/repeater-rows.js
@@ -170,7 +170,7 @@ import { Fields } from './';
 											className="button-delete"
 											label={ __( 'Delete', 'block-lab' ) }
 											onClick={ this.removeRow( rowIndex ) }
-											disabled={ rows.length <= field.min ? true : false }
+											disabled={ !! field.min && rows.length <= field.min }
 											isSmall
 										/>
 									</div>
--- a/js/blocks/components/repeater-rows.js
+++ b/js/blocks/components/repeater-rows.js
@@ -152,7 +152,7 @@ import { Fields } from './';
 	 * Renders the repeater rows.
 	 */
 	render() {
-		const { rows, subFields, parentBlockProps, parentBlock } = this.props;
+		const { rows, field, subFields, parentBlockProps, parentBlock } = this.props;
 
 		return (
 			<Fragment>
@@ -170,6 +170,7 @@ import { Fields } from './';
 											className="button-delete"
 											label={ __( 'Delete', 'block-lab' ) }
 											onClick={ this.removeRow( rowIndex ) }
+											disabled={ rows.length <= field.min ? true : false }
 											isSmall
 										/>
 									</div>
--- a/js/blocks/controls/repeater.js
+++ b/js/blocks/controls/repeater.js
@@ -14,7 +14,7 @@ const BlockLabRepeaterControl = ( props ) => {
 	const { attributes, setAttributes } = parentBlockProps;
 	const attr = { ...attributes };
 	const value = attr[ field.name ];
-	const defaultRows = [ {} ];
+	const defaultRows = new Array( field.min ? field.min : 1 ).fill( { '': '' } );
 	const hasRows = value && value.hasOwnProperty( 'rows' );
 	const rows = hasRows ? value.rows : defaultRows;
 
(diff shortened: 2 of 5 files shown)


js/blocks/components/repeater-rows.js
```diff
@@ -170,6 +170,7 @@ import { Fields } from './';
 											className="button-delete"
 											label={ __( 'Delete', 'block-lab' ) }
 											onClick={ this.removeRow( rowIndex ) }
+											disabled={ rows.length <= field.min ? true : false }
```
kienstra: Maybe this should first check that `field.min` isn't `null`, like:

```js
disabled={ !! field.min && rows.length <= field.min }
```

`field.min` can be `null` if no value is selected in the UI.

Still, this also works locally:

```js
disabled={ rows.length <= field.min }
```
...as no positive number will be less than `null`

js/blocks/controls/repeater.js
```diff
@@ -50,7 +52,7 @@ const BlockLabRepeaterControl = ( props ) => {
 					label={ __( 'Add new', 'block-lab' ) }
 					labelPosition="bottom"
 					onClick={ addEmptyRow }
-					disabled={ false }
+					disabled={ rows.length >= field.max ? true : false }
```
kienstra: Likewise, it looks like this works, without the ternary expression:
```js
disabled={ !! field.max && rows.length >= field.max  }
```

RobStino (approved):
This is great! The UX in the Block Builder, as well as the Block Editor, makes sense.
The only thing I was thinking was that a notice of _why_ you can't add/delete more rows might be helpful to the end-user, but I realised that this is the job of the field help text and can be easily added by the block creator.

php/blocks/controls/class-repeater.php
```diff
@@ -46,5 +46,23 @@ public function __construct() {
 	 */
 	public function register_settings() {
 		$this->settings[] = new Control_Setting( $this->settings_config['help'] );
+		$this->settings[] = new Control_Setting(
+			array(
+				'name'     => 'min',
+				'label'    => __( 'Minimum Rows', 'block-lab' ),
+				'type'     => 'number_non_negative',
+				'default'  => '',
```
kienstra: This is a nitpick, but it looks like `'default' => ''` is already present [as a default value](https://github.com/getblocklab/block-lab/blob/a93d9a7b52b7aadf17d8541770778dfc86a7e610/php/blocks/controls/class-control-setting.php#L43) of `Control_Setting`, so this might not be needed.

kienstra (changes requested):
**Looks Good**
*Some Minor Points*

Hi @lukecarbis,
This looks good:

disabled-button-styling

There are some minor points, but I think this is close to being ready to merge.

js/blocks/components/repeater-rows.js
lukecarbis: GREAT idea!

js/blocks/controls/repeater.js
lukecarbis: Addressed in https://github.com/getblocklab/block-lab/pull/421/commits/92075e8fd76cc52150164d12a0e1211fe0a0dd17.

js/blocks/components/repeater-rows.js
lukecarbis: Addressed in https://github.com/getblocklab/block-lab/pull/421/commits/92075e8fd76cc52150164d12a0e1211fe0a0dd17.

php/blocks/controls/class-repeater.php
lukecarbis: Addressed in https://github.com/getblocklab/block-lab/pull/421/commits/056200ea2f500e069777327691156d842c81b455.

js/blocks/components/repeater-rows.js
kienstra: Thanks, this looks good.

js/blocks/controls/repeater.js
kienstra: Thanks, this also looks good.

kienstra:
This looks good. I'm going to test this really briefly for good measure, and probably approve it.

kienstra:
**Approved**

Hi @lukecarbis,
Testing of this looks good. Ship it 😄

kienstra (approved):

PR #422, opened 2019-09-10
Rename phpunit.xml.dist to phpunit.xml

lukecarbis:
I think this should pass travis without needing to changing `.travis.yml`, but we'll find out soon.


kienstra (approved):
**Approved**

Hi @lukecarbis,
Good idea, this looks good.

PR #429, opened 2019-09-13
Repeater row count and row index function.

### LINKED ISSUE #403: Repeater row count
Add the ability to get the number of rows in a repeater

lukecarbis:
This adds some helper functions for retrieving the total count of repeater rows, and the index of an individual repeater row.

The helper functions are:

`block_row_count( $name )`
This takes a repeater field name as the argument, and returns the row count for that repeater. (It can also return false, if you pass it an invalid repeater name.)

`block_row_index()`
This should be used inside a repeater loop (after the `block_row()` call). This returns the index of the current row.

Here's a code example:
```PHP
if ( block_rows( 'repeater' ) && block_row_count( 'repeater' ) > 3 ):

    while ( block_rows( 'repeater' ) ) :
        block_row( 'repeater' );

        // Note: The index is zero-based, so we should add 1 to get a non-programmer count.
        echo '<li>This is row number ' . block_row_index() + 1 . '</li>';
    endwhile;

    echo '</ul>';
endif;
```

Closes #403.


php/helpers.php
```diff
@@ -150,6 +150,47 @@ function reset_block_rows( $name ) {
...
+/**
+ * Return the total amount of rows in a repeater.
+ *
+ * @param string $name The name of the repeater field.
+ * @return int|bool The total amount of rows. False if the repeater isn't found.
+ */
+function block_row_count( $name ) {
+	global $block_lab_attributes;
+
+	if ( ! isset( $block_lab_attributes[ $name ] ) ) {
```
kienstra: This function looks good overall.

Is this conditional necessary?

It seems that if it's not set, the block below will return false:

```php
if ( ! isset( $block_lab_attributes[ $name ]['rows'] ) ) {
	return false;
}
```

lukecarbis: That's a really good point.

The thinking here is:

1. Check if a row even exists
2. Check if it's a repeater (has rows)

Given this, do you still think it's worth only doing the single (second) condition? I'm more than happy to make that change. Just wanted to explain my approach.

kienstra: Sure, it's kind of a nitpick 😄 

My guess is that there's not much performance difference between the existing approach, and an approach with only one `isset()` check.

php/blocks/class-field.php
```diff
@@ -183,6 +183,18 @@ public function cast_value( $value ) {
 				break;
 		}
 
+		if ( 'repeater' === $this->control ) {
```
kienstra: Could the logic in this class be in the [Repeater class](https://github.com/getblocklab/block-lab/blob/41fd3922f44ce1dbb1c4677d5b7f54317c5959a6/php/blocks/controls/class-repeater.php#L1), in a `validate()` method?

php/helpers.php
lukecarbis: Addressed in https://github.com/getblocklab/block-lab/pull/429/commits/6300f331a2abbf27078188488a2290dd4f3d29fd.

php/blocks/class-field.php
lukecarbis: Done in https://github.com/getblocklab/block-lab/pull/429/commits/a1bb136adad13a6a7f02dd557771f5ce1ad3b2db.

lukecarbis:
@kienstra Thanks for your review. Ready for you to take another look.

Wanted to make note of https://github.com/getblocklab/block-lab/pull/429/commits/8d0b4f6fc7bd78feb038d04c9607f3dbddc664fd. I was trying to remove repeater specific functionality from the `Field` class, and move it into the `Repeater` control. If you don't think it makes much sense, though, I'd be happy to revert this commit.

php/blocks/class-field.php
```diff
@@ -194,6 +194,17 @@ public function cast_value( $value ) {
...
+		/**
+		 * Filters the value of a field when being output as a string.
+		 *
+		 * The result of this filter should be a string. Even if it isn't, it will be run
+		 * through strval() anyway.
+		 *
+		 * @param mixed $value  The value to cast as a string.
+		 * @param Field $field  This field.
+		 */
+		$value = apply_filters( 'block_lab_cast_field_value_to_string', $value, $this );
```
kienstra: This filter looks good, great documentation of it.

What do you think of moving it to right above the return statement:

```php
return strval( $value );
```

php/blocks/controls/class-repeater.php
```diff
@@ -24,19 +24,28 @@ class Repeater extends Control_Abstract {
...
-	 * The Repeater control is an array of objects, with each row being an object.
-	 * For example, a repeater with one row might be [ { 'example-text': 'Foo', 'example-image': 4232 } ].
+	 * The Repeater control is an array of arrays, with each row being its own array.
+	 * For example, a repeater with two rows might be:
+	 * [ 'rows': [ 0: [ 'example-text': 'Foo', 'example-image': 42 ], 1: [ 'example-text': 'Bar', 'example-image': 32 ] ] ].
 	 *
 	 * @var string
 	 */
-	public $type = 'object';
+	public $type = 'array';
```
kienstra: Good thing to bring up. In PHP, this would be an `array()`, and in JS it'd be an object. I don't see many examples of array or object attributes in Core blocks.

php/blocks/class-loader.php
```diff
@@ -463,6 +466,10 @@ public function block_template( $name, $type = 'block' ) {
 			if ( ! current_user_can( 'edit_posts' ) || ! isset( $templates[0] ) ) {
 				return;
 			}
+			// Hide the template not found notice on the frontend, unless WP_DEBUG is enabled.
```
kienstra, line 466: Good idea

kienstra:
Hi @lukecarbis,
Nice work here. I have to go to bed, and I'll take a look at it again tomorrow.

php/blocks/class-field.php
kienstra: Actually, don't worry about this comment for now.

php/blocks/controls/class-repeater.php
lukecarbis: Ah – makes much more sense for `$type` to represent the Javascript type, rather than the PHP type. Addressed in https://github.com/getblocklab/block-lab/pull/429/commits/2b61b6476075fa1c01ec3ae866065fe1ba33fc7a.

php/blocks/class-loader.php
```diff
@@ -308,7 +308,10 @@ public function render_block_template( $block, $attributes ) {
 			foreach ( $block->fields as $field ) {
 				if ( isset( $field->settings['sub_fields'] ) ) {
 					$sub_field_settings = $field->settings['sub_fields'];
-					$rows               = $attributes[ $field->name ]['rows'];
+				}
+
+				if ( isset( $attributes[ $field->name ]['rows'] ) ) {
```
kienstra: Maybe this conditional could be moved above to line `309`, instead of creating a new `if` block:

```php
if ( isset( $field->settings['sub_fields'] ) && isset( $attributes[ $field->name ]['rows'] ) ) {
```

Right now, it's possible (though unlikely) for `$sub_field_settings` to be unset at line 318.

php/blocks/controls/class-repeater.php
```diff
@@ -63,4 +63,31 @@ public function register_settings() {
...
+
+	/**
+	 * Remove empty placeholder rows.
+	 *
+	 * @param mixed $value The value to either make available as a variable or echoed on the front-end template.
+	 * @param bool  $echo Whether this will be echoed.
+	 * @return mixed $value The value to be made available or echoed on the front-end template.
+	 */
+	public function validate( $value, $echo ) {
+		if ( isset( $value['rows'] ) ) {
```
kienstra, line 75: How about something like:

```php
array_filter( $value['rows'] );
```

It [looks like](https://www.thecodedeveloper.com/remove-empty-array-elements-in-php/) that removes empty array values, like `'' => ''`

kienstra (approved):
**Looks Good**
*Some Minor Points*

Hi @lukecarbis,
This looks good. The new functions should be really useful.

There are just a few minor points. I'll be in and out today, so feel free to merge when you're happy.

php/blocks/class-loader.php
lukecarbis: Good catch.

php/blocks/controls/class-repeater.php
lukecarbis: This would be nice, except that `array_filter` filters empty values (not empty keys) – so if there were a repeater with an empty value, that item in the array would be removed.

We could supply a custom callback to array_filter which targets empty keys, or alternatively use something like this:
```
array_diff_key( $value['rows'][ $key ], array_flip( array( '', 0 ) ) );
```

but ultimately, I think just unsetting the two is a better, simpler, solution.

php/blocks/class-loader.php
lukecarbis: Addressed in https://github.com/getblocklab/block-lab/pull/429/commits/39fade9da38e99c9da5562128ac3189776233cb2.

kienstra: Thanks 😄

php/blocks/controls/class-repeater.php
kienstra: Ah, it removes empty values. That's not good.

kienstra:
Like we talked about, this should be good to merge whenever you'd like 😄

PR #432, opened 2019-09-18
Add Classic Text control, including list controls

### LINKED ISSUE #405: Consider adding list controls to the Rich Text control
The [Rich Text control](https://github.com/getblocklab/block-lab/blob/fa36740ce6c0de3019187186dff6f82cee038689/js/blocks/controls/rich-text.js) doesn't have a way to add lists, like `<ul>` or `<ol>`:

rich-text-here

...though the Core List block does have list controls:

list-controls-here

Though these are [marked as unstable](https://github.com/WordPress/gutenberg/blob/db70326242c303f676a6127385e4d785b3604d5a/packages/block-library/src/list/edit.js#L14-L20
), it'd be good to consider using something like them for this plugin's Rich Text control.

For reference, here is the Rich Text control [when it had alignment controls](https://github.com/getblocklab/block-lab/blob/1b7c97fc50218a72b63d4c068c9b50329f1af8fc/js/blocks/controls/rich-text.js).

kienstra:
### Update: this has changed, please see this [comment below](https://github.com/getblocklab/block-lab/pull/432#issuecomment-538701506).
* This begins to add list buttons to the Rich Text controls
* Still, it might not be possible to get this to work as we'd expect. For example, it looks like it's not possible to have the list buttons next to the bold and italic buttons:

rich-text-control

Also, it might be really hard to make only part of the text a list.

For example, in the Core List block, it simply makes its entire `<RichText>` component a `<ul>` or `<ol>`. I'm not sure if the `<RichText>` has a good way to make only part of the text into a `<ul>` or `<ol>`.

Closes #405 (if this works)

### DIFF
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -185,7 +185,7 @@ class TinyMCE extends Component {
 				id={ `toolbar-${ editorId }` }
 				className="classic-text__toolbar"
 				onClick={ this.focus }
-				data-placeholder={ __( 'Rich text', 'block-lab' ) }
+				data-placeholder={ __( 'Classic text', 'block-lab' ) }
 				onKeyDown={ this.onToolbarKeyDown }
 			/>,
 			<div
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -183,7 +183,7 @@ class TinyMCE extends Component {
 			<div
 				key="toolbar"
 				id={ `toolbar-${ editorId }` }
-				className="rich-text__toolbar"
+				className="classic-text__toolbar"
 				onClick={ this.focus }
 				data-placeholder={ __( 'Rich text', 'block-lab' ) }
 				onKeyDown={ this.onToolbarKeyDown }
@@ -191,7 +191,7 @@ class TinyMCE extends Component {
 			<div
 				key="editor"
 				id={ `editor-${ editorId }` }
-				className="rich-text__edit"
+				className="classic-text__edit"
 			/>,
 		];
 		/* eslint-enable jsx-a11y/no-static-element-interactions */
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -137,13 +137,7 @@ class TinyMCE extends Component {
 		} );
 
 		editor.on( 'init', () => {
-			const rootNode = this.editor.getBody();
-
-			// Create the toolbar by refocussing the editor.
-			if ( document.activeElement === rootNode ) {
-				rootNode.blur();
-				this.editor.focus();
-			}
+			this.editor.focus();
 		} );
 	}
 
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -131,7 +131,7 @@ class TinyMCE extends Component {
 		} );
 
 		editor.addButton( 'wp_add_media', {
-			tooltip: __( 'Insert Media' ),
+			tooltip: __( 'Insert Media', 'block-lab' ),
 			icon: 'dashicon dashicons-admin-media',
 			cmd: 'WP_Medialib',
 		} );
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -137,7 +137,13 @@ class TinyMCE extends Component {
 		} );
 
 		editor.on( 'init', () => {
-			this.editor.focus();
+			const rootNode = this.editor.getBody();
+
+			// Create the toolbar by refocussing the editor.
+			if ( document.activeElement === rootNode ) {
+				rootNode.blur();
+				this.editor.focus();
+			}
 		} );
 	}
 
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -137,11 +137,18 @@ class TinyMCE extends Component {
 		} );
 
 		editor.on( 'init', () => {
-			const rootNode = this.editor.getBody();
+			const { activeElement } = document;
 
-			// Create the toolbar by refocussing the editor.
-			rootNode.blur();
-			this.editor.focus();
+			// Force focus to the editor, as this seems to be the only way to display the toolbar.
+			// @see https://github.com/WordPress/gutenberg/pull/4948
+			this.editor.fire( 'focus' );
+
+			// Refocus to whatever was in focus before the editor, likely most of the block.
+			editor.once( 'focus', () => {
+				if ( activeElement !== this.editor.getBody() ) {
+					activeElement.focus();
+				}
+			} );
 		} );
 	}
 
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -27,6 +27,11 @@ function isTmceEmpty( editor ) {
  * @see https://github.com/WordPress/gutenberg/blob/416c9bc9eaef6e6bceea923b6f36642746c01aba/packages/block-library/src/classic/edit.js
  */
 export default class TinyMCE extends Component {
+	/**
+	 * Constructs the class.
+	 *
+	 * @param {Object} props The component properties.
+	 */
 	constructor( props ) {
 		super( props );
 		this.initialize = this.initialize.bind( this );
@@ -34,6 +39,9 @@ export default class TinyMCE extends Component {
 		this.focus = this.focus.bind( this );
 	}
 
+	/**
+	 * A lifecycle method, called after the component mounts.
+	 */
 	componentDidMount() {
 		const { baseURL, suffix } = window.wpEditorL10n.tinymce;
 
@@ -49,30 +57,42 @@ export default class TinyMCE extends Component {
 		}
 	}
 
+	/**
+	 * A lifecycle method, called before the component unmounts.
+	 */
 	componentWillUnmount() {
 		window.addEventListener( 'DOMContentLoaded', this.initialize );
-		wp.oldEditor.remove( `editor-${ this.props.clientId }` );
+		wp.oldEditor.remove( `editor-${ this.props.editorId }` );
 	}
 
+	/**
+	 * A lifecycle method, called after the component updates.
+	 *
+	 * @param {Object} prevProps The previous props of the component, before the update.
+	 */
 	componentDidUpdate( prevProps ) {
-		const { clientId, content } = this.props;
+		const { content, editorId, onChange } = this.props;
 
-		const editor = window.tinymce.get( `editor-${ clientId }` );
+		const editor = window.tinymce.get( `editor-${ editorId }` );
 
 		if ( prevProps.content !== content ) {
 			editor.setContent( content || '' );
+			onChange( content || '' );
 		}
 	}
 
+	/**
+	 * Initializes the TinyMCE.
+	 */
 	initialize() {
-		const { clientId } = this.props;
+		const { editorId } = this.props;
 		const { settings } = window.wpEditorL10n.tinymce;
-		wp.oldEditor.initialize( `editor-${ clientId }`, {
+		wp.oldEditor.initialize( `editor-${ editorId }`, {
 			tinymce: {
 				...settings,
 				inline: true,
 				content_css: false,
-				fixed_toolbar_container: `#toolbar-${ clientId }`,
+				fixed_toolbar_container: `#toolbar-${ editorId }`,
 				setup: this.onSetup,
 				toolbar1: 'bold,italic,bullist,numlist,outdent,indent,alignleft,aligncenter,alignright,link,unlink,wp_add_media',
 				toolbar2: '',
@@ -80,9 +100,13 @@ export default class TinyMCE extends Component {
 		} );
 	}
 
+	/**
+	 * Handles events in the editor.
+	 *
+	 * @param {Object} editor The editor.
+	 */
 	onSetup( editor ) {
 		const { content, onChange } = this.props;
-		const { ref } = this;
 		let bookmark;
 
 		this.editor = editor;
@@ -109,6 +133,10 @@ export default class TinyMCE extends Component {
 			bookmark = null;
 		} );
 
+		editor.on( 'change', () => {
+			onChange( editor.getContent() );
+		} );
+
 		editor.on( 'keydown', ( event ) => {
 			if ( ( event.keyCode === BACKSPACE || event.keyCode === DELETE ) && isTmceEmpty( editor ) ) {
 				// delete the block
@@ -124,13 +152,6 @@ export default class TinyMCE extends Component {
 			 */
 			if ( altKey && event.keyCode === F10 ) {
 				event.stopPropagation();
-			}
-		} );
-
-		// Show the second, third, etc. toolbars when the `kitchensink` button is removed by a plugin.
-		editor.on( 'init', function() {
-			if ( editor.settings.toolbar1 && editor.settings.toolbar1.indexOf( 'kitchensink' ) === -1 ) {
-				editor.dom.addClass( ref, 'has-advanced-toolbar' );
 			}
 		} );
 
@@ -151,12 +172,20 @@ export default class TinyMCE extends Component {
 		} );
 	}
 
+	/**
+	 * If the editor exists, this focuses it.
+	 */
 	focus() {
 		if ( this.editor ) {
 			this.editor.focus();
 		}
 	}
 
+	/**
+	 * Handles a keydown event for the toolbar.
+	 *
+	 * @param {Object} event The keydown event.
+	 */
 	onToolbarKeyDown( event ) {
 		// Prevent WritingFlow from kicking in and allow arrows navigation on the toolbar.
 		event.stopPropagation();
@@ -164,8 +193,11 @@ export default class TinyMCE extends Component {
 		event.nativeEvent.stopImmediatePropagation();
 	}
 
+	/**
+	 * Renders the component.
+	 */
 	render() {
-		const { clientId } = this.props;
+		const { editorId } = this.props;
 
 		// Disable reasons:
 		//
@@ -177,7 +209,7 @@ export default class TinyMCE extends Component {
 		return [
 			<div
 				key="toolbar"
-				id={ `toolbar-${ clientId }` }
+				id={ `toolbar-${ editorId }` }
 				ref={ ( ref ) => this.ref = ref }
 				className="rich-text__toolbar"
 				onClick={ this.focus }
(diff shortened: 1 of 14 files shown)


kienstra:
**Use TinyMCE?**

Hi @lukecarbis,
What do you think about https://github.com/getblocklab/block-lab/pull/432/commits/0f48d3d0e411fbd5456e2471f2452af5785ebac6, which uses TinyMCE instead of the `<RichText>` component? 

The list buttons work well so far, and it even has alignment controls. Though this PR could use some more polish.

So far, it looks to be backwards-compatible with the earlier Rich Text control. It's mainly forked from the [Core Classic block](https://github.com/WordPress/gutenberg/blob/416c9bc9eaef6e6bceea923b6f36642746c01aba/packages/block-library/src/classic/edit.js).

Also, it's much easier to use than `<RichText>` so far. 

rich-text-tiny

kienstra:
**Lists**

Lists are working, but they could use better styling:

lists-work-here

kienstra:
Another bonus is that we could later add controls that the Classic block has, like text color, strikethrough, or horizontal line.

kienstra:
If you have time, feel free to commit styling changes on this PR.

kienstra:
**Request For Review**

Hi @lukecarbis and @RobStino,
Could you please review this? It adds list and alignment controls to the Rich Text control.

# Features Of New Rich Text

new-showing

As far as I've tested this, it's backwards-compatible with the previous control. But the 'Code' button doesn't look to be available in the new TinyMCE implementation:

code-button

Still, text that was code-formatted with the previous control still looks the same on loading this new implementation.

# Before

before-rt

# After 

after-rt

This also surfaced a bug with the previous Rich Text implementation, which I think this new version fixes.

Before, if there was an empty line:

previous-rich-text

...it didn't display in the `<ServerSideRender>` or the front-end:

server-side-render-bad-display

This PR fixes that. That might be considered a backwards-compatibility break, but I think it's more of a bug fix.

lukecarbis (changes requested):
Looking great Ryan! I can't believe I'm saying this, but I think switching to TinyMCE might actually be a good idea.

Couple of things I noticed:

1. The toolbar isn't displayed until the textarea is focused. After that it remains, even on blur. My suggestion is to make it always visible. Alternatively make it only show when the textarea is focused, and the mouse is moving (hide again when you start typing), similar to the previous implementation of rich text.



2. A few styling suggestions for the toolbar:
- Remove the box-shadow from `.mce-tinymce`
- Center align
- White background
- 2px bottom margin
- 4px border radius on `.mce-toolbar .mce-btn-group .mce-btn`
- Make the buttons match the current rich text style (different hover / active effect)

Screen Shot 2019-09-23 at 8 47 38 am

3. Placeholder text doesn't seem to work.

———

More on the use of TinyMCE. What do you think about this idea:

- We use the `<TinyMCE>` component in our  "Rich Text" control, as per this implementation
- Rather than removing the `<RichText>` component entirely, we switch our current Textarea control to use the `<RichText>` component
- Possibly, we simplify the `<RichText>` component in Textarea to remove the Inline Image

kienstra:
**Thanks for reviewing this!**

Hi @lukecarbis,
Good point here!

1. Ah, good catch. I'll work on that tomorrow if that's alright. Though it'd be really good to have this PR in a release this week. There's a user that needs this for a project.

2. Thanks, those suggestions are applied in https://github.com/getblocklab/block-lab/pull/432/commits/9714ee781eeb7f1ae3db65730d2190d7cc0f763e:

toolbar-styling

3. Yeah, good catch. What do you think of removing the placeholder? 

I didn't see a good way to add a placeholder for TinyMCE. There are [workarounds](https://stackoverflow.com/questions/27237876/how-do-i-add-placeholder-text-to-tinymce), like adding a TinyMCE plugin. 

UX is definitely not my speciality, but I'm not sure if a placeholder applies to a full rich text editor.

kienstra:
>Rather than removing the <RichText> component entirely, we switch our current Textarea control to use the <RichText> component

Interesting 😄 It's probably possible, we could probably `wpautop()` the previous Textarea value, to make it compatible with a `<RichText>` component.

UX isn't my specialty 😄 But I'm not sure if people would expect bold, italic, and link controls in a Textarea field.

kienstra:
Hi @lukecarbis,
You've been really productive today already 😄 

But if you happen to have time and want to dig into it, feel free to work on [point 1 above](https://github.com/getblocklab/block-lab/pull/432#pullrequestreview-291531559) and commit to this branch.

I'll have a pretty busy day tomorrow, and might not be able to completely fix it.

lukecarbis:
@kienstra Thanks for looking at all that. Larger reply still incoming, but just wanted to mention that I noticed that although https://github.com/getblocklab/block-lab/pull/432/commits/effe21d425e1a6b6f2679d57a8bfd4c60ca7ae1c helps keep the toolbar visible after focusing, it's not visible when first adding the block to a post.

lukecarbis:
Also, the "selected" state doesn't match the regular <RichText> component. See the "Bold" below.

Screen Shot 2019-09-24 at 8 12 08 am

kienstra:
Here's the current display of buttons with `.is-active` and `:hover`:

active-hov

kienstra:
Sorry, I probably won't have more time today to work on this.

kienstra:
Hm, I might have made some progress on this.

kienstra:
This looks to fix the issue of the toolbar not displaying, but it has its own issues:

```diff
diff --git a/js/blocks/components/tiny-mce.js b/js/blocks/components/tiny-mce.js
index 9923331..38ddc76 100644
--- a/js/blocks/components/tiny-mce.js
+++ b/js/blocks/components/tiny-mce.js
@@ -71,10 +71,12 @@ class TinyMCE extends Component {
                wp.oldEditor.initialize( `editor-${ editorId }`, {
                        tinymce: {
                                ...settings,
-                               inline: true,
                                content_css: false,
                                fixed_toolbar_container: `#toolbar-${ editorId }`,
+                               inline: false,
+                               resize: false,
                                setup: this.onSetup,
+                               statusbar: false,
                                toolbar1: 'bold,italic,bullist,numlist,outdent,indent,alignleft,aligncenter,alignright,link,unlink,wp_add_media,strikethrough',
                                toolbar2: '',
                        },
```

kienstra:
Sorry, that's probably all I can work on this today.

lukecarbis:
My suggestion at this point is to create a new control to house this `<TinyMCE>` component (maybe called "Classic Text"), and keep the "Rich Text" control as it currently is.

RobStino:
Hey all.
Tested this fairly thoroughly and overall I really like this!
I liked how it didn't wrap the output on the frontend with any divs or anything.

Here's the field in the editor:
image
Looked good in the Editor when deselected:
image

Looked and behaved as expected on the frontend as well.

My suggestions for improvements (if possible/practical) include:
- Set the default height to 5 rows. Currently, it's only 1 row.
- Include Heading control
- Include code font style control

Possible other controls, although maybe not...
- Font color
- Font size

As far as moving forward I can see how this is experience is what we want, but can't yet deliver in the Rich Text. I think I'm in favour of having both, and then potentially merging/migrating at some point in the future when we can achieve feature parity.

My only concern is that it kind of goes against our "it just works" product philosophy. Having two options that do kinda the same thing doesn't feel great. But, it is what it is. Happy to defer on this one.

RobStino:
I also observed the bug where the controls bar doesn't display on initial load.

kienstra:
With https://github.com/getblocklab/block-lab/pull/432/commits/28b88d52e86c725d60381233b8e24e738aebae7a, the `min-height` of this Control is the equivalent of 3 lines, even if there's no text:

min-height-rich-text

Feel free to change the `min-height`.

Also, it looks like the toolbar now displays on first creating a block, and on focusing the block. It doesn't require clicking in the edit area anymore.

kienstra:
Once #444 is merged into `develop`, merging `develop` into this branch should make the failing unit tests above pass.

Update: I applied the same fix in this PR, so it should be good.

kienstra:
With https://github.com/getblocklab/block-lab/pull/432/commits/3cd1f5285ab0e1aa6e11fc8b8c4e7d64e8518aca, there is now a separate control, Classic Text. 

This PR's changes for the Rich Text control are moved into that, and the Rich Text control is mainly unchanged in this PR.

new-ct

This doesn't have a placeholder, though it has help text and a default value.

kienstra:
There is now a 'formatselect' in the toolbar:

here-is-a-heading

tests/php/unit/helpers/trait-testing-helper.php
```diff
@@ -42,7 +42,7 @@ public function set_license_validity( $is_valid ) {
 		if ( $is_valid ) {
 			$transient_value = array(
 				'license' => 'valid',
-				'expires' => date( '+1 month' ),
+				'expires' => date( 'D, d M Y H:i:s', time() + 1000 ),
```
kienstra: This fixes the failed unit tests. I'm not sure how this worked before 😄

kienstra:
**Request For Review**

Hi @lukecarbis,
Hope you had a great weekend.

Could you please review this Classic Text control?

Like we talked about, this creates a new control, and the Rich Text control is hardly changed in this PR.

classic-text-control

Thanks, Luke!

kienstra:
**Thanks For Reviewing**

Hi @RobStino,
Sorry for the delay in replying to your [review](https://github.com/getblocklab/block-lab/pull/432#issuecomment-535337069).

>My suggestions for improvements (if possible/practical) include:

> Set the default height to 5 rows. Currently, it's only 1 row.

Good idea. Sorry, I missed your comment, and only made it 3 rows. How is that? It's easy to change here if you'd prefer:

min-height-rich-text


> Include Heading control

Good suggestion, that's applied [here](https://github.com/getblocklab/block-lab/pull/432#issuecomment-538772431).

> Include code font style control

Sure, that's now included in the [formatselect](https://github.com/getblocklab/block-lab/pull/432#issuecomment-538772431), as 'Preformatted'.

> Possible other controls, although maybe not...
> Font color
> Font size

Sure, any controls in the Core Classic block are available here. Though adding more might require another toolbar row, like the Core Classic block has:

classic-text-control

js/blocks/components/tiny-mce.js
```diff
@@ -0,0 +1,201 @@
+/**
+ * WordPress dependencies
+ */
+const { Component } = wp.element;
+const { __ } = wp.i18n;
+const { F10 } = wp.keycodes;
+
+/**
+ * Forked from the Core Classic Editor's Edit component.
```
lukecarbis, line 9: What do we get by forking instead of just including?

php/post-types/class-block-post.php
```diff
@@ -41,6 +41,7 @@ class Block_Post extends Component_Abstract {
 	public $pro_controls = array(
 		'repeater',
 		'post',
+		'classic_text',
```
lukecarbis: Can we please switch these two lines so that Rich Text appears before Classic?

lukecarbis:
@kienstra This is looking great! I left a couple of questions, but at this point I think it's just about getting the styles looking right.

I know we've gone to some effort to get the controls to match the Rich Text, but I think we should actually change that now! :/ Sorry.

Partly, this is due to the width of the toolbar now much wider due to the formatselect.

Also, we were trying to replicate the RichText style earlier because we intended to _replace_ it. Now that we're embracing the "Classic"ness, maybe we should style the toolbar more like this:

Screen Shot 2019-10-07 at 9 31 56 am

(Not saying we need all these controls, just that our toolbar could look / be styled similarly).

@RobStino , do you have any opinion on this?

js/blocks/components/tiny-mce.js
kienstra: Yeah, it'd be good to include it somehow.

But I don't think there's a good way to do that.

For example, the Classic block's `index.js` file [imports it](https://github.com/WordPress/gutenberg/blob/416c9bc9eaef6e6bceea923b6f36642746c01aba/packages/block-library/src/classic/index.js#L9):


```js
import edit from './edit';
```

php/post-types/class-block-post.php
kienstra: Sure

kienstra: It's moved in https://github.com/getblocklab/block-lab/pull/432/commits/1e6aac76933c3a661c69c73d3bb9c9664b81c269

kienstra:
Yeah, it'll probably be good to remove the styling that made this look like the `<RichText>` component. It should also make this more maintainable, with less styling.

RobStino:
Hey fellas. I agree with Luke. Let's maintain the styling of the Classic block, so to speak. Maintainability it makes sense, and it also differentiates it from the RichText further.

kienstra:
Thanks, @RobStino! That sounds good. I'll apply that to this PR in the next few days.

kienstra:
**Question About New Styling**

Hi @lukecarbis and @RobStino,
What do you think about this new styling for the toolbar? https://github.com/getblocklab/block-lab/pull/432/commits/44e95d20d2c3dd587cade0017ab4a3298f96bdc7 removes all of the custom styling, except the rules that centered it and made it white:

buttons-s

We could also remove the styling that centers the toolbar and makes it white. I'm not sure if that looks good, but that's definitely not my specialty 😄 

classic-text-distinct-buttons

lukecarbis:
@kienstra – Hope you don't mind, I pushed a commit (https://github.com/getblocklab/block-lab/pull/432/commits/674f66207c5537c0fe7bece9915b18625b3df9c7) which has some style updates. 

Here's what it looks like now:
Screen Shot 2019-10-13 at 9 00 51 am

And focused:
Screen Shot 2019-10-13 at 9 00 37 am

@RobStino Can I get your sign-off on these designs?

lukecarbis:
Found one last bug! Related to the Classic Text stealing focus.

1. Create a new block, which includes a Classic Text field, as well as some other text controls.
2. Make sure the Classic Text field is not the _first_ one.
3. Add the Block to a post
4. Notice that the Classic Text field is focused first when the Block is selected

kienstra:
Ah, good catch! I'll probably get to this within a few days.

kienstra:
It looks like there might not be an easy fix for the issue with the [Classic editor's toolbar focusing](https://github.com/getblocklab/block-lab/pull/432#issuecomment-541368486):

https://github.com/WordPress/gutenberg/pull/4948

kienstra:
I'll look at this a little more, but there [might not be a fix](https://github.com/WordPress/gutenberg/pull/4948#issuecomment-365242220) for the [Classic editor's toolbar focusing](https://github.com/getblocklab/block-lab/pull/432#issuecomment-541368486):

>I don't really see a way of fixing the focus problem without having to copy over a significant amount of code to the classic block to overwrite the theme. It would be nicer if the inline theme just had a setting to show toolbars all the time instead of just when the editor is focussed.

kienstra:
One option would be to only display the toolbar on focusing the TinyMCE. This would at least prevent the [issue with focusing](https://github.com/getblocklab/block-lab/pull/432#issuecomment-541368486).

kwfricke:
This is awesome. I CAN'T WAIT for this to be released :D :D

kienstra:
Thanks, @kwfricke! Looking forward to getting it to you 😄

kienstra:
**Question About Toolbar**

Hi @lukecarbis,
Hope you're doing great. Could you please help with a question?

I couldn't find a fix for the [issue with the Classic Text control's toolbar focusing](https://github.com/getblocklab/block-lab/pull/432#issuecomment-541368486).

It looks like the TinyMCE that this uses requires a focus event for the toolbar to display.

The [workaround in Gutenberg](https://github.com/WordPress/gutenberg/blob/853b3cb100469f9d71bad93bb38ece5210dae0b2/packages/block-library/src/classic/edit.js#L157) seems to work because when you add a Classic block, the TinyMCE is focused. That's pretty much all the block has 😄 

But when you add a Block Lab block that has several fields, including a TinyMCE (Classic Text), it's more complex.

Do any of these options sound acceptable?

1. Only display the toolbar on clicking in (focusing) the Classic Text field (like [how it was in this PR before](https://github.com/getblocklab/block-lab/pull/432#pullrequestreview-291531559))
2. Keep this PR as is, with the known [issue with focusing](https://github.com/getblocklab/block-lab/pull/432#issuecomment-541368486)

Thanks, Luke!

lukecarbis:
@kienstra Let's go with option 1: Only display the toolbar on clicking in (focusing) the Classic Text field (like how it was in this PR before).

@kwfricke Sorry this is taking so long! We want to make sure it's right! :)

kienstra:
Thanks, Luke! That sounds good.

kienstra:
**Request For Some Styling Work And Final Review**

Hi @lukecarbis,
Thanks, https://github.com/getblocklab/block-lab/pull/432/commits/fd8e87f9166bbe55b66b24aebde75de2c127bc67 makes the toolbar only display on clicking in (focusing) the Classic Text field (like [how it was in this PR before](https://github.com/getblocklab/block-lab/pull/432#pullrequestreview-291531559)).

Your [styling commit](https://github.com/getblocklab/block-lab/pull/432/commits/674f66207c5537c0fe7bece9915b18625b3df9c7) looked great, and unfortunately, that commit I just made to the JavaScript messed it up 😄 

So I made a few edits to `editor.scss`, but could you please work on the styling more? 

class-text-control

Thanks, Luke!

lukecarbis:
@kienstra Styles are in. Is it possible to hide the toolbar when the classic text area loses focus? Or was that one of the limitations of this approach?

lukecarbis:
FWIW, the Rich Text toolbar doesn't disappear when it loses focus either. I would normally create a separate issue for this – but if there's a simple win here, may as well include it in this PR along with the Classic Text toolbar.

kienstra:
>@kienstra Styles are in. Is it possible to hide the toolbar when the classic text area loses focus? Or was that one of the limitations of this approach?

Hi @lukecarbis, 
It might be possible, but if it's alright, the weekend might be the earliest I could work on this. Would it be acceptable to ship this as is?

kienstra:
**Request For Final Review**

Hi @RobStino,

There's a [known issue](https://github.com/getblocklab/block-lab/pull/432#issuecomment-544179824) where the toolbar only displays on clicking the edit area. There doesn't look to be a workaround.

Thanks, Rob!

kienstra:
>@kienstra Styles are in. Is it possible to hide the toolbar when the classic text area loses focus? Or was that one of the limitations of this approach?

I'll look at this more this weekend, but I don't think it'll be easy to hide the toolbar. There doesn't look to be an API for the toolbar, it's only shown on focus with a workaround (maybe hack).

lukecarbis (approved):

lukecarbis:
@kienstra As discussed in our last conversation, I'm happy with where is at, and we're all happy to move forward with this PR understanding that there's a limitation with the toolbar not hiding.

kienstra:
Sounds good, Luke! Thanks for all of your work on this.

RobStino:
Works perfectly!

**Form**
image

**Preview**
image

**Front end**
image

kienstra:
Wow, great! Thanks a lot for testing this.

PR #434, opened 2019-09-23
PHP API for registering blocks with Block Lab

### LINKED ISSUE #400: Block builder API
A feature request I hear a lot is a way of programatically adding new blocks, using PHP to define them.

For example:
```PHP
block_lab()
    ->new_block( 'Team Member', 'team-member' )
        ->add_field( 'Name', 'name', 'text' )
        ->add_field( 'Position', 'position', 'text' )
        ->add_field( 'Photo', 'photo', 'image' )
        ->add_field( 'Twitter', 'twitter', 'url' );
```

Inspired by [this](https://github.com/StoutLogic/acf-builder/wiki) and [this](https://www.advancedcustomfields.com/resources/register-fields-via-php/).

lukecarbis:
I've implemented the following functions, which enable to ability to register blocks programatically:

`block_lab_add_block( $block_name, $block_config )`

and

`block_lab_add_field( $block_name, $field_name, $field_config )`

Both are well documented in their respective doc blocks, if you'd like to learn more about these functions and how to use them.

They need to run during the `block_lab_add_blocks` action hook.

Here's some example code I have been testing with:

```PHP
// One huge long array with lots defined.
block_lab_add_block(
	'one-fish',
	array(
		'title'    => 'One Fish',
		'category' => 'common',
		'icon'     => 'waves',
		'excluded' => array( 'page' ),
		'keywords' => array( 'sad', 'glad', 'bad' ),
		'fields'   => array(
			'thin' => array(
				'label'   => 'Thin',
				'control' => 'toggle',
				'width'   => '25',
				'default' => true,
			),
			'fat' => array(
				'label'   => 'Fat',
				'control' => 'toggle',
				'width'   => '25',
				'default' => false,
			),
			'hat'  => array(
				'label'   => 'Hat',
				'control' => 'select',
				'width'   => '50',
				'options' => array(
					array(
						'label' => 'Yellow',
						'value' => 'yellow',
					),
					array(
						'label' => 'Red',
						'value' => 'red',
					),
					array(
						'label' => 'Blue',
						'value' => 'blue',
					),
				),
			),
		),
	)
);

// A more minimalist approach.
block_lab_add_block( 'two-fish', array( 'icon' => 'waves' ) );
block_lab_add_field( 'two-fish', 'hello-there-ned-how-do-you-do' );
block_lab_add_field( 'two-fish', 'tell-me-tell-me-what-is-new', array( 'placeholder' => 'My hat is old, my teeth are gold.' ) );
```

Closes #400.


kienstra:
Hi @lukecarbis,
Nice, this will be really useful.

If it's alright, tomorrow is probably the earliest I could review this.

php/helpers.php
```diff
@@ -309,3 +309,136 @@ function block_field_config( $name ) {
...
+
+	if ( ! isset( $block_config['excluded'] ) ) {
+		$block_config['excluded'] = array();
+	}
+
+	if ( ! isset( $block_config['keywords'] ) ) {
+		$block_config['keywords'] = array();
+	}
+
+	if ( ! isset( $block_config['fields'] ) ) {
```
kienstra: [wp_parse_args()](https://codex.wordpress.org/Function_Reference/wp_parse_args) might make this a little easier

php/helpers.php
```diff
@@ -309,3 +309,136 @@ function block_field_config( $name ) {
...
+		$block_config['fields'] = array();
+	}
+
+	add_filter(
```
kienstra: The logic in the callback looks good.

What do you think about adding a public function to class `Loader` like:

`block_lab()->loader->add_block( $block_config );`

Then, it could store the registered blocks in a protected property of `Loader`.

That could have most of the logic in this callback, but this particular `add_filter()` call wouldn't be needed.

The `apply_filters( 'block_lab_blocks' )` call in `Loader` would still be useful, as would this `block_lab_add_block()` function.

php/helpers.php
```diff
@@ -309,3 +309,136 @@ function block_field_config( $name ) {
...
+
+	if ( ! isset( $field_config['order'] ) ) {
+		$field_config['order'] = 0;
+	}
+
+	if ( ! isset( $field_config['settings'] ) ) {
+		$field_config['settings'] = array();
+	}
+
+	add_filter(
```
kienstra: Like the comment above, what do you think about turning this `add_filter()` call into something like:

```php
block_lab()->loader->add_field( $block_name, $field_config );
```
...or whatever method name.

It could then add fields to blocks added via the [possible function above](https://github.com/getblocklab/block-lab/pull/434#discussion_r329407225): `block_lab()->loader->add_block( $block_config );`

kienstra (commented):
**Looks Good, Question About Adding Method To Loader**

Hi @lukecarbis,
This looks good, it's a big win for the plugin.

What do you think about moving the logic from the `add_filter()` calls into a new method in the `Loader` class?

This and #435 depend on the `Loader` class, but before, it wasn't really available in any other file.

#440 moves it into `block_lab()->loader`, and makes most of the method protected.

Either way, hope you're doing great, and look forward to talking tomorrow.

php/helpers.php
kienstra: Ah, that wouldn't be able to add fields to existing blocks created with `/wp-admin` > Block Lab > Add New.

Would that be a common use case, to add a block via via  `/wp-admin` > Block Lab > Add New, then later add a field to it with `block_lab_add_field()` ?

lukecarbis: I don't think it would be a very common use case, no.

php/helpers.php
lukecarbis: Address in https://github.com/getblocklab/block-lab/pull/434/commits/e547cc32c00f439325e53d68d840695dbce54f2e.

php/helpers.php
lukecarbis: Addressed in https://github.com/getblocklab/block-lab/pull/434/commits/5b98f71fd97a3e6fb6973b89c98f15bbb1993e10.

php/helpers.php
lukecarbis: Addressed in https://github.com/getblocklab/block-lab/pull/434/commits/5b98f71fd97a3e6fb6973b89c98f15bbb1993e10.

lukecarbis:
@kienstra Thanks for you're review! Great to have your eyes on it, great suggestions.

I've made some changes to address your comments. Could you please take another look?

php/helpers.php
kienstra: Nice, that looks good. Thanks for pushing it.

kienstra:
Hi @lukecarbis, this looks really good. I think the logic is good to go 😄 

Would you mind if I pushed some unit tests to this branch just for completeness? For example, I was thinking of suggesting we eventually change to `array()` to the `[]` syntax, and it'd be good to have as much coverage as possible if we ever make that change.

lukecarbis:
Please feel free to push unit tests.

kienstra:
Thanks, Luke!

composer.json
```diff
@@ -9,6 +9,7 @@
     },
     "require-dev": {
         "brain/monkey": "^2",
+        "mockery/mockery": "^1.2.4",
```
kienstra, line 12: This allows mocking entire classes, like `Loader`. It doesn't look like [Brain Monkey](https://brain-wp.github.io/BrainMonkey/docs/wordpress-hooks-added.html) can mock classes.

kienstra:
Using the PHP API [example above](https://github.com/getblocklab/block-lab/pull/434#issue-320112300), the block:

1. Looks as expected in the Edit component
2. Saves its values as expected
3. Looks as expected in the `<ServerSideRender>` component and on the front end

expected-block-good

kienstra (approved):
**Great work!**

Hi @lukecarbis,
This looks really good.

The API worked as expected, and I think it's ready to merge when you're ready.

php/blocks/class-loader.php
```diff
@@ -569,15 +569,15 @@ public function add_block( $block_config ) {
...
+	 * @param string $block_name The name of the block that the field is added to.
 	 * @param array $field_config The config of the field to add.
 	 */
 	public function add_field( $block_name, $field_config ) {
 		if ( ! isset( $this->blocks[ "block-lab/{$block_name}" ] ) ) {
 			return;
 		}
 		if ( ! isset( $field_config['name'] ) ) {
-			return $blocks;
+			return;
```
kienstra, line 616: Is it alright that I made this change? Maybe something else was intended.

lukecarbis: No, nothing else intended. Good pickup.

PR #435, opened 2019-09-23
Replace global variables with data array in Plugin class

### LINKED ISSUE #425: Replace globals $block_lab_attributes and $block_lab_config
The globals `$block_lab_attributes` and `$block_lab_config` are used in the class [Loader](https://github.com/getblocklab/block-lab/blob/25d16644074dfdce1b639d0b272cd17d9c745e18/php/blocks/class-loader.php#L284) and in [helpers.php](https://github.com/getblocklab/block-lab/blob/25d16644074dfdce1b639d0b272cd17d9c745e18/php/helpers.php#L26).

It'd be best to not use globals, and instead move them into an OOP structure.

At the moment, there aren't any references to those globals, when searching WordPress plugins for them on https://wpdirectory.net/

If we're going to replace them, it'd be good to do that soon, before other plugins or themes use them.

lukecarbis:
Closes #425.

### DIFF
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -78,7 +78,7 @@ class Loader extends Component_Abstract {
 	}
 
 	/**
-	 * Retreive data from the Loader's data store.
+	 * Retrieve data from the Loader's data store.
 	 *
 	 * @param string $key The data key to retrieve.
 	 * @return mixed
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -33,7 +33,7 @@ class Loader extends Component_Abstract {
 	 *
 	 * @var array
 	 */
-	protected $data = array();
+	protected $data = [];
 
 	/**
 	 * Load the Loader.
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -29,6 +29,13 @@ class Loader extends Component_Abstract {
 	protected $blocks = '';
 
 	/**
+	 * A data store for sharing data to helper functions.
+	 *
+	 * @var array
+	 */
+	protected $data = array();
+
+	/**
 	 * Load the Loader.
 	 *
 	 * @return $this
@@ -68,6 +75,37 @@ class Loader extends Component_Abstract {
 		 * PHP block loading.
 		 */
 		add_action( 'plugins_loaded', $this->get_callback( 'dynamic_block_loader' ) );
+	}
+
+	/**
+	 * Retreive data from the Loader's data store.
+	 *
+	 * @param string $key The data key to retrieve.
+	 * @return mixed
+	 */
+	public function get_data( $key ) {
+		$data = false;
+
+		if ( isset( $this->data[ $key ] ) ) {
+			$data = $this->data[ $key ];
+		}
+
+		/**
+		 * Filters the data that gets returned.
+		 *
+		 * @param mixed  $data The data from the Loader's data store.
+		 * @param string $key  The key for the data being retreived.
+		 */
+		$data = apply_filters( 'block_lab_data', $data, $key );
+
+		/**
+		 * Filters the data that gets returned, specifically for a single key.
+		 *
+		 * @param mixed  $data The data from the Loader's data store.
+		 */
+		$data = apply_filters( "block_lab_data_{$key}", $data );
+
+		return $data;
 	}
 
 	/**
@@ -344,8 +382,8 @@ class Loader extends Component_Abstract {
 			$this->enqueue_global_styles();
 		}
 
-		block_lab()->data['attributes'] = $attributes;
-		block_lab()->data['config']     = $block;
+		$this->data['attributes'] = $attributes;
+		$this->data['config']     = $block;
 
 		if ( ! is_admin() && ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST ) && ! wp_doing_ajax() ) {
 
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -281,8 +281,6 @@ class Loader extends Component_Abstract {
 	 * @return mixed
 	 */
 	public function render_block_template( $block, $attributes ) {
-		global $block_lab_attributes, $block_lab_config;
-
 		$type = 'block';
 
 		// This is hacky, but the editor doesn't send the original request along.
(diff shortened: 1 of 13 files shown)


kienstra:
Hi @lukecarbis,
Nice, you had a productive day!

If it's alright, tomorrow is probably the earliest I could review this.

php/blocks/class-loader.php
```diff
@@ -332,8 +330,8 @@ public function render_block_template( $block, $attributes ) {
 			$this->enqueue_global_styles();
 		}
 
-		$block_lab_attributes = $attributes;
-		$block_lab_config     = $block;
+		block_lab()->data['attributes'] = $attributes;
+		block_lab()->data['config']     = $block;
```
kienstra: What do you think about storing `->data` as a protected property of this `Loader` class? 

That way, it wouldn't be possible to change the data outside this class.

Then, `helpers.php` could access the data with something like `block_lab()->loader->get_data( 'attributes' )`, or whatever method name.

#440 has infrastructure for this, including storing `Loader` as a property of `Plugin`, not in the protected property `->components`.

kienstra (commented):
**Question About Storing Data In Loader**

Hi @lukecarbis,
This PR looks good, and it'll keep us from being stuck with these globals.

What do you think about storing the `->data` in the `Loader` class? That should keep the data from being changed anywhere else, and it could still be accessed in `helpers.php`.

Hope you had a great weekend, it was definitely productive 😄

php/blocks/class-loader.php
kienstra: If it's stored in the `Plugin` class, it has to be publicly accessible. But if it's stored in `Loader`, it can be set inside that class, and no other class can set it. They can only get it.

lukecarbis:
@kienstra I've implemented your suggestion (great idea by the way), but I'm struggling with some failing integration tests. I'm kind of stuck – could you please take a look?

kienstra:
Sure, @lukecarbis! The integration tests aren't very easy to maintain 😄 

It might be a few days before I get to this, if that's alright.

php/helpers.php
```diff
@@ -18,14 +18,25 @@
...
-	/*
-	 * Defined in Block_Lab\Blocks\Loader->render_block_template().
+	/**
+	 * An array containing the block attributes.
 	 *
-	 * @var array
+	 * @var array $attributes
 	 */
-	global $block_lab_attributes, $block_lab_config;
+	$attributes = block_lab()->loader->get_data( 'attributes' );
```
kienstra, line 21: This DocBlock was definitely helpful when there was a global, but do you think it's needed now that it's getting `$attributes` with a normal method call?

And likewise for some other places in this PR.

php/helpers.php
```diff
@@ -126,9 +137,14 @@ function block_row( $name ) {
  * @return bool
  */
 function block_rows( $name ) {
-	global $block_lab_attributes;
```
kienstra, line 129: Nice, it's great to see the `global` removed.

php/helpers.php
```diff
@@ -126,9 +137,14 @@ function block_row( $name ) {
...
+	/**
+	 * An array containing the block attributes.
+	 *
+	 * @var array $attributes
+	 */
+	$attributes = block_lab()->loader->get_data( 'attributes' );
 
-	if ( ! isset( $block_lab_attributes[ $name ] ) ) {
+	if ( ! $attributes || ! isset( $attributes[ $name ] ) ) {
```
kienstra: `! isset( $attributes[ $name ] )` should be enough here, as it'll be false if `! isset( $attributes )`

php/helpers.php
```diff
@@ -172,13 +188,18 @@ function reset_block_rows( $name ) {
...
-	global $block_lab_attributes;
+	/**
+	 * An array containing the block attributes.
+	 *
+	 * @var array $attributes
+	 */
+	$attributes = block_lab()->loader->get_data( 'attributes' );
 
-	if ( ! isset( $block_lab_attributes[ $name ]['rows'] ) ) {
+	if ( ! $attributes || ! isset( $attributes[ $name ]['rows'] ) ) {
```
kienstra: `! isset( $attributes[ $name ]['rows'] )` should be enough, as it'll also be false if `! $attributes`

php/helpers.php
```diff
@@ -303,9 +345,16 @@ function block_config() {
  * @return array|null
  */
 function block_field_config( $name ) {
-	global $block_lab_config;
```
kienstra, line 306: It's great to see `global` removed 😄

kienstra (changes requested):
**Looks Good**
*A Few Small Points*

Hi @lukecarbis,
Nice work! This looks good.

There are just a few points.

If it's alright, I [pushed to this branch](https://github.com/getblocklab/block-lab/pull/435/commits/fe6e5835e2af38ff1bc245a6021d8aa8750cb099) to get the Integration tests to pass. Sorry, I try not to do that without asking.

Would you mind if I pushed a PHPUnit test for `Loader::get_data()` to this branch? 

Thanks, Luke!

lukecarbis:
@kienstra please go ahead!

kienstra:
Thanks, https://github.com/getblocklab/block-lab/pull/435/commits/3e143382926ff3c33da53298e570afe0726aabee adds a test for `Loader::get_data()`

php/helpers.php
lukecarbis: Done in https://github.com/getblocklab/block-lab/pull/435/commits/06d78c727f1a014a6905dd05bb64d2725795ca75

php/helpers.php
lukecarbis: Done in https://github.com/getblocklab/block-lab/pull/435/commits/0fbf45adb15a4ce1a8fd10112939b68a67249365

php/helpers.php
lukecarbis: Done in https://github.com/getblocklab/block-lab/pull/435/commits/0fbf45adb15a4ce1a8fd10112939b68a67249365

lukecarbis:
@kienstra Ready for you to take another look. :)

kienstra (approved):
**Approved**

Hi @lukecarbis,
This looks great! It's to merge when you'd like. 

It'd be good to mention in the Changelog the new API for this, like
`block_lab->loader->get_data( 'attributes' )` instead of `$block_lab_attributes`

PR #439, opened 2019-09-29
[WIP] Create button group for field types instead of select input

lukecarbis:
_Very much_ a work in progress. I started on something, and just looking for some early feedback before I put more hours in.

… what do we think about this approach?

Need to clean up borders and border radiuses.

Screen Shot 2019-09-29 at 12 39 55 pm

### DIFF
--- a/css/admin.block-post.css
+++ b/css/admin.block-post.css
@@ -161,12 +161,30 @@
 	border-bottom: 1px solid #e1e1e1;
 	border-top: 1px solid #e1e1e1;
 }
+#block_fields .block-fields-rows .block-fields-edit-control .button-group {
+	white-space: normal;
+}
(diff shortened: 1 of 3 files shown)


RobStino:
I think this could work. I reckon if we could reduce the size of the buttons and the button group, that could help. Possibly with icons and text inline?
I jumped into the inspector to play around with it, but the HTML structure was a bit funky. Or was for me at least.
I did spot this layout issue as well.
image

CSS Grid could help here nicely with layout.

kienstra:
Hi @lukecarbis,
This looks pretty good.

UX definitely isn't my specialty. But it seems like this takes a lot of space. 

use-space-here

My guess is that users won't change the field types much once they've selected them, though maybe they'd change from text to rich text, or text to email.

lukecarbis:
Going to abandon this in favour of a select2 style solution.

PR #440, opened 2019-09-30
Experiment with moving Loader class into a property of Plugin, making most methods protected

kienstra:
* This should allow easier access to `Loader`
* `Loader` is a dependency of other files, like `helpers.php`, which uses the (recently changed) globals `$block_lab_attributes` and `$block_lab_config`
* This allows adding public methods to `Loader` that other classes can access, like `block_lab()->loader->add_block()`
* Most methods of `Loader` are now `protected`

Related to #434, #435

### DIFF
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -19,14 +19,14 @@ class Loader extends Component_Abstract {
 	 *
 	 * @var array
 	 */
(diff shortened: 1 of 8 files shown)


lukecarbis (approved):
👌

kienstra:
Thanks, Luke!

PR #444, opened 2019-10-05
Prevent `</p>` output in Rich Text control

### LINKED ISSUE #437: Empty Rich Text field outputs </p>
Hi, if you add a Rich Text field to a block but don't enter any text then it outputs a `</p>` tag in the frontend.  The expected behaviour would be a blank string similar to how the Text field functions.  This happens when the RTE is a top level field and also when it is added as a sub field to a repeater.

Black Lab 1.4.1

kienstra:
## Steps to reproduce
1. Create a new post
2. Add a Block Lab block that has a Rich Text field without a default value
3. Don't enter any text in the Rich Text field
4. Publish the post, and go to the front-end
5. Expected: The output is well-formed html
6. Actual: The output is `</p>`, as reported in #437:

closing-p

It looks like when the <RichText> component has no text, it saves `<p></p>`. But passing that to wpautop() returns `</p>`:

wpautop-here

So in that case, this PR simply returns the string.

Closes #437

### DIFF
--- a/php/blocks/controls/class-rich-text.php
+++ b/php/blocks/controls/class-rich-text.php
@@ -51,6 +51,12 @@ class Rich_Text extends Control_Abstract {
 	 */
 	public function validate( $value, $echo ) {
 		unset( $echo );
+
+		// For this value, wpautop() returns '</p>', so simply return the value.
+		if ( '<p></p>' === $value ) {
+			return $value;
+		}
+
 		return wpautop( $value );
 	}
 }
--- a/php/blocks/controls/class-rich-text.php
+++ b/php/blocks/controls/class-rich-text.php
@@ -52,9 +52,9 @@ class Rich_Text extends Control_Abstract {
 	public function validate( $value, $echo ) {
 		unset( $echo );
 
-		// For this value, wpautop() returns '</p>', so simply return the value.
+		// If there's no text entered, Rich Text saves '<p></p>', so instead return ''.
 		if ( '<p></p>' === $value ) {
-			return $value;
+			return '';
 		}
 
 		return wpautop( $value );
--- a/tests/php/unit/blocks/controls/test-class-rich-text.php
+++ b/tests/php/unit/blocks/controls/test-class-rich-text.php
@@ -95,5 +95,8 @@ class Test_Rich_Text extends \WP_UnitTestCase {
 		// This should have the same results, whether the second $echo argument is true or false.
 		$this->assertEquals( $expected_markup_with_p_tags, $this->instance->validate( $markup_with_br_tags, true ) );
 		$this->assertEquals( $expected_markup_with_p_tags, $this->instance->validate( $markup_with_br_tags, false ) );
+
+		$empty_paragraph = '<p></p>';
+		$this->assertEquals( $empty_paragraph, $this->instance->validate( $empty_paragraph, false ) );
 	}
 }
--- a/tests/php/unit/blocks/controls/test-class-rich-text.php
+++ b/tests/php/unit/blocks/controls/test-class-rich-text.php
@@ -97,6 +97,6 @@ class Test_Rich_Text extends \WP_UnitTestCase {
 		$this->assertEquals( $expected_markup_with_p_tags, $this->instance->validate( $markup_with_br_tags, false ) );
 
 		$empty_paragraph = '<p></p>';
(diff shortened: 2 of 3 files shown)


kienstra:
**Request For Review**

Hi @lukecarbis,

Thanks!

php/blocks/controls/class-rich-text.php
```diff
@@ -51,6 +51,12 @@ public function register_settings() {
 	 */
 	public function validate( $value, $echo ) {
 		unset( $echo );
+
+		// For this value, wpautop() returns '</p>', so simply return the value.
+		if ( '<p></p>' === $value ) {
```
lukecarbis, line 56: Something like this might handle more weird edge cases (such as `<span></span>` or `<p><strong></strong></p>`):

```PHP
if ( empty( esc_html( $value ) ) ) {
    return $value;
}
// or…
if ( '' === esc_html( $value ) ) {
    return $value;
}
// or even…
if ( '' === esc_html( $value ) ) {
    return '';
}
```

lukecarbis (approved):
One small suggestion here, feel free to merge with or without it.

php/blocks/controls/class-rich-text.php
kienstra: Ah, interesting point.

I'm not sure if there will be a case where `empty( esc_html( $value ) )`, or `'' === esc_html( $value )`, though maybe I just haven't seen it. 

It looks like even if there's no value entered in the Rich Text control, its value is `<p></p>`

lukecarbis: I could be missing something here.
- If the Rich Text control value is empty, the value is set to `<p></p>`
- The expected value would be an empty string, right?
- The logic above says "After removing all the HTML, if I'm left with an empty string, return an empty string"

kienstra: Hi @lukecarbis,
All good, those are good points.

>If the Rich Text control value is empty, the value is set to `<p></p>`

Right 😄 

>The expected value would be an empty string, right?

Good question, maybe 😄 

The way I thought of this was like the Paragraph block, which also uses the `<RichText>` component. It outputs `<p></p>` when it's empty, just like this Rich Text control does:

empty-paragraph-block

This is definitely debatable, though. The paragraph block isn't exactly like this.


>The logic above says "After removing all the HTML, if I'm left with an empty string, return an empty string"

Ah, `esc_html()` doesn't usually remove HTML, it just escapes it. 

For example, `esc_html( '<div></div>' )` returns `&lt;div&gt;&lt;/div&gt;`

kienstra: Maybe [strip_tags()](https://www.php.net/manual/en/function.strip-tags.php)

lukecarbis: Ah – of course! Sorry, I did mean strip_tags. Even better, though: [`wp_strip_all_tags`](https://developer.wordpress.org/reference/functions/wp_strip_all_tags/).

kienstra: Sure, I thought that's what you meant.

kienstra: Though it's unlikely, it's possible that a Rich Text field has only an image, in which case it would be stripped entirely, as there's no text other than the markup:

image-in-rich-text

kienstra: https://github.com/getblocklab/block-lab/pull/444/commits/a65299c07cbe1432f27b760aa511c2dcc12b4ef6 changes this to return `''` if the Rich Text field is empty (if it saves `<p></p>`).

kienstra:
Hi @lukecarbis,
Happy weekend!

What do you think of the [latest changes](https://github.com/getblocklab/block-lab/pull/444#discussion_r332307316) here?

Thanks!

php/blocks/controls/class-rich-text.php
lukecarbis: Aah! Great pickup on the image!

lukecarbis:
Great! Going to merge this on you behalf, if you don't mind, so we can get that unit test fix in a couple of other PRs.

php/blocks/controls/class-rich-text.php
kienstra: Thanks, it's kind of an edge case. Otherwise, `strip_tags()` would work really well.

kienstra:
Sounds good, thanks for merging this.

PR #447, opened 2019-10-11
Ensure that adding capabilities is robust

### LINKED ISSUE #445: Crashes site if administrator role renamed
Hi, we use the Members plugin to manage user roles and permissions.  We have two levels of administrator and so renamed the default.  When we try to view the site we get the "The site is experiencing technical difficulties." message when trying to view any pages.  The error log shows:

PHP Fatal error:  Uncaught Error: Call to a member function add_cap() on null in [snip]/plugins/block-lab/php/post-types/class-block-post.php:243

So digging down to that line shows:

$admins = get_role( 'administrator' );

The only way I could get back into the admin was to move the block-lab directory outside the project, login to the WP admin and rename one of the two custom admin roles back to "administrator".  Then move the block-lab directory back.

It's quite common practise to rename the administrator account.

kienstra:
* Address an issue issue where `$admins` was `null`, as it can be changed in user management plugins.
* Also, adds an integration test for capabilities. This test passes before this PR is applied, and with it. 
* So there's no expected change or backwards-compatibility break.

Fixes #445


kienstra:
This might take a while, I'm working on getting the integration test to pass on Travis, though it passes locally.

php/post-types/class-block-post.php
```diff
@@ -205,26 +205,13 @@ public function register_post_type() {
...
-			'capabilities'  => array(
-				'edit_post'          => 'block_lab_edit_block',
-				'edit_posts'         => 'block_lab_edit_blocks',
-				'edit_others_posts'  => 'block_lab_edit_others_blocks',
-				'publish_posts'      => 'block_lab_publish_blocks',
-				'read_post'          => 'block_lab_read_block',
-				'read_private_posts' => 'block_lab_read_private_blocks',
-				'delete_post'        => 'block_lab_delete_block',
-			),
+			'capabilities'  => $this->get_capabilities(),
```
kienstra, line 214: There should be no change here, `$this->get_capabilities()` simply wraps the previous code.

tests/php/integration/test-post-capabilities.php
```diff
@@ -0,0 +1,119 @@
...
+use Block_Lab\Post_Types;
+
+/**
+ * Class Test_Post_Capabilities
+ *
+ * Tests the capabilities for the 'block_lab' post type.
+ *
+ * @package Block_Lab
+ */
+class Test_Post_Capabilities extends \WP_UnitTestCase {
```
kienstra, line 17: This test passes even with the previous implementation. For example:

```bash
$ git checkout develop php/post-types/class-block-post.php
$ phpunit --filter Test_Post_Capabilities
OK (42 tests, 42 assertions)
```

php/post-types/class-block-post.php
```diff
@@ -238,15 +225,31 @@ public function register_post_type() {
...
-		$admins->add_cap( 'block_lab_delete_block' );
+		$admin = get_role( 'administrator' );
+		if ( ! $admin ) {
+			return;
+		}
+
+		foreach ( $this->get_capabilities() as $capability => $custom_capability ) {
+			$admin->add_cap( $custom_capability );
+		}
+	}
```
kienstra, line 236: This essentially does the same thing as before, though it exits if `! $admin`

kienstra:
**Request For Review**

Hi @lukecarbis,
When you can, could you please review this PR?

There's no intended change to capabilities, it's only intended to prevent an issue when the `administrator` role is changed by a plugin.

Thanks, Luke!

php/post-types/class-block-post.php
```diff
@@ -205,26 +205,13 @@ public function register_post_type() {
...
 			'show_in_menu'  => true,
 			'menu_position' => 100,
-			// This SVG icon is being included from the plugin directory, so using file_get_contents is okay.
-			// @codingStandardsIgnoreStart
-			'menu_icon'     => 'data:image/svg+xml;base64,' . base64_encode(
-				file_get_contents(
-					$this->plugin->get_assets_path( 'images/admin-menu-icon.svg' )
-				)
+			'menu_icon'     => 'data:image/svg+xml;base64,' . base64_encode( // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
+				file_get_contents( $this->plugin->get_assets_path( 'images/admin-menu-icon.svg' ) ) // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- This SVG icon is being included from the plugin directory, so using file_get_contents is okay.
```
kienstra, line 209: No intended change here, just to the `// phpcs: ignore` comments

lukecarbis (approved):
@kienstra Love the introduction of the `get_capabilities` method. Great work!

kienstra:
Thanks, @lukecarbis!

PR #448, opened 2019-10-22
Enable selective import updates

jasperSpeicher:
Allow the user to selectively update existing block configurations during import, rather than refusing to update them at all.


lukecarbis:
I'll merge this into a separate branch, and continue work from there. Thanks so much @jasperSpeicher.

PR #454, opened 2019-10-28
Fix rich text toolbar styles

lukecarbis:
Fixes a regression with the toolbar styles for the Rich Text field.

Before:
Screen Shot 2019-10-28 at 2 04 43 pm

After:
Screen Shot 2019-10-28 at 2 14 38 pm

### DIFF
--- a/css/src/editor.scss
+++ b/css/src/editor.scss
@@ -421,7 +421,7 @@ div[class^="wp-block-block-lab-"] {
 
 /* Miscellaneous global styles */
 .edit-post-layout {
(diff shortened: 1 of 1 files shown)


lukecarbis:
Just realised that this problem only occurs when the Gutenberg plugin is active. Still, the fix is simple and worth doing.

kienstra:
Hi @lukecarbis,
This looks good. I couldn't reproduce the original issue, but you have a really good handle on styling, much better than me 😄

kienstra (approved):

PR #463, opened 2019-10-30
Bugfix for the PHP API – allow themes to hook in via the functions.php file

lukecarbis:
In my testing of the PHP API I was always just using code that I threw into a quick little plugin.

Just now, I tried registering a block using the PHP API via the `functions.php` file that is loaded with the theme… and **it didn't work**!

This is because the action was firing before the theme was ready.

This PR fixes that by retrieving the blocks a little later (during `init`), which also requires that we load the blocks a little later (moved from `plugins_ready` to `init`).

I have tested fairly extensively (previous blocks still work, new blocks work, blocks are loaded from a JSON file and work, blocks can be registered via the PHP API through a theme, blocks load in the editor, blocks load on the frontend).

That said, this is a small change that could have a large impact. Please test thoroughly!

### DIFF
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -54,8 +54,6 @@ class Loader extends Component_Abstract {
 			],
 		];
 
-		$this->retrieve_blocks();
-
 		return $this;
 	}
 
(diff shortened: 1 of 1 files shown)


php/blocks/class-loader.php
```diff
@@ -606,7 +609,7 @@ public function add_block( $block_config ) {
 	 * that the block isn't added too late.
 	 *
 	 * @param string $block_name The name of the block that the field is added to.
-	 * @param array $field_config The config of the field to add.
+	 * @param array  $field_config The config of the field to add.
```
kienstra, line 612: Ah, good one 😄

kienstra:
The blocks still load as expected everywhere I tested them:

## Content Blocks UI

blocks-load-as-expected

## Edit Block UI

editing-block-still-works

## Block Editor

block-registered-in-functions

## Front-end 

fe-values

kienstra (approved):
**Approved**

Hi @lukecarbis,
This looks good.

Like you mentioned, it allows registering a block in a theme's `functions.php`.

Also, block still work everywhere I tested them.

Like you mentioned, this could use some more testing, with how important it is. But I couldn't find an issue.

PR #467, opened 2019-11-02
Fixes for WP 5.3 compatibility

### LINKED ISSUE #459: Post field doesn't go 100% wide
TBH, 100% wide would probably look not as good as it is, but it breaks the standardised UX of all the other fields.

We should either make them go 100% or remove the width settings from those fields

In Block Builder
image

In editor
image

### LINKED ISSUE #460: Controls overlap with Rich Text
The controls for the Rich text field overlap the field above.
image

kienstra:
Fixes for WP Core 5.3 compatibility

The block editor in WP Core 5.3-RC3 now looks good, other than the points below.

I tested that all of the controls look good with WP 5.3-RC3, and display their entered values in the `<ServerSideRender>`:

* In the main editor
* In the Inspector Controls 
* In a repeater
* With all 25%, 50%, 75%, and 100% width

I didn't see an issue, other than those mentioned below. Though of course this could use more testing.

This also fixes #460 
This also fixes #459

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Contributors: lukecarbis, ryankienstra, Stino11, rheinardkorf
 Tags: Gutenberg, Blocks
 Requires at least: 5.0
-Tested up to: 5.2
+Tested up to: 5.3
 Requires PHP: 5.6
 Stable tag: trunk
 License: GPLv2 or later
--- a/css/src/editor.scss
+++ b/css/src/editor.scss
@@ -158,6 +158,7 @@ div[class^="wp-block-block-lab-"] {
       margin: 1em 0 0;
 
       .components-base-control__label {
+        display: block;
         font-weight: 600;
       }
     }
@@ -167,11 +168,11 @@ div[class^="wp-block-block-lab-"] {
       font-size: 1em;
       color: rgba(0, 0, 0, 0.8);
     }
-  }
 
-  /* FetchInput component */
-  .bl-fetch-input {
-    max-width: $input-width;
+    /* Override a max-width: 25rem style rule in Core that can prevent displaying at the selected with, like 75% */
+    .components-select-control__input {
+      max-width: unset;
+    }
   }
 
   /* Rich Text Control Component */
--- a/css/src/editor.scss
+++ b/css/src/editor.scss
@@ -418,6 +418,10 @@ div[class^="wp-block-block-lab-"] {
     }
   }
 
+  .block-lab-color-control .components-base-control__label {
+    display: block;
+  }
+
   /* The FetchInput component */
   .bl-fetch--input input[type="text"] {
     width: 100%;
--- a/css/src/editor.scss
+++ b/css/src/editor.scss
@@ -4,6 +4,7 @@ $input-padding: 8px;
 $input-width: 300px;
 $default-font-size: 13px;
 $break-small: 600px;
+$font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
 
 @mixin break-small() {
 	@media (min-width: #{ ($break-small) }) {
@@ -85,6 +86,10 @@ div[class^="wp-block-block-lab-"],
       position: relative;
     }
 
+    .bl-image__placeholder .components-button.is-button {
+      white-space: normal;
+    }
+
     .components-form-file-upload,
     .components-media-library-button,
     .bl-image__remove {
@@ -129,6 +134,7 @@ div[class^="wp-block-block-lab-"] {
 
     p {
       font-size: 1rem;
+      font-family: $font-family;
     }
 
     .block-lab-control {
@@ -152,6 +158,10 @@ div[class^="wp-block-block-lab-"] {
       .block-lab-control.width-75 {
         flex-basis: 100%;
       }
+    }
+
+    .components-base-control {
+      font-family: $font-family;
     }
 
     .components-base-control__field {
--- a/css/src/editor.scss
+++ b/css/src/editor.scss
@@ -177,6 +177,10 @@ div[class^="wp-block-block-lab-"] {
 
   /* Rich Text Control Component */
   .block-lab-rich-text-control {
+    .components-base-control__field {
+      margin-top: 20px;
+    }
+
     .input-control {
       background: #fff;
       min-height: 7em;


css/src/editor.scss
```diff
@@ -167,15 +168,19 @@ div[class^="wp-block-block-lab-"] {
...
+    /* Override a max-width: 25rem style rule in Core that can prevent displaying at the selected with, like 75% */
+    .components-select-control__input {
+      max-width: unset;
+    }
   }
 
   /* Rich Text Control Component */
   .block-lab-rich-text-control {
+    .components-base-control__field {
+      margin-top: 20px;
```
kienstra, line 191: This addresses https://github.com/getblocklab/block-lab/issues/460. Both screenshots are with WP 5.3.

# Before 

before-little-or-no-space

# After

after-more-space

css/src/editor.scss
```diff
@@ -158,6 +158,7 @@ div[class^="wp-block-block-lab-"] {
       margin: 1em 0 0;
 
       .components-base-control__label {
+        display: block;
```
kienstra, line 171: This fixes an issue in WP 5.3 with a few fields, like Color and Select:

above-here

css/src/editor.scss
```diff
@@ -167,15 +168,19 @@ div[class^="wp-block-block-lab-"] {
       font-size: 1em;
       color: rgba(0, 0, 0, 0.8);
     }
-  }
 
-  /* FetchInput component */
-  .bl-fetch-input {
-    max-width: $input-width;
```
kienstra, line 174: This style rule didn't allow `<FetchInput>` fields like Post and Taxonomy to respect the selected width, like 75% or 100%.

kienstra: Though maybe that's intended. Maybe we could also remove the 25%, 50%, 75%, 100% option from those fields.

kienstra:
Also, the Color field almost always appears at this width, in WP 5.2.4 (current stable) and 5.3-RC3, no matter if a width is selected, like 50% or 75%:

Is this intentional? Most color values won't need much more space, though it's possible to have something like:

```
rgba(66, 191, 66, 0.26666666666666666)
```

color-at-width-here

css/src/editor.scss
```diff
@@ -413,6 +418,10 @@ div[class^="wp-block-block-lab-"] {
     }
   }
 
+  .block-lab-color-control .components-base-control__label {
+    display: block;
```
kienstra, line 432: This is a similar fix to [this](https://github.com/getblocklab/block-lab/pull/467#discussion_r341829224), but for the Color field in the Inspector.

kienstra:
There looks to be an issue with closing the color picker in the Image control when in the Inspector, but that is also present in the current stable WP Core (5.2.4).

kienstra:
There also looks to be an issue with having a Rich Text and Classic Text control in the same block. For example, when using shortcuts like Command B in the Classic Text block, sometimes they actually apply to the Rich Text block. Though this issue also exists in the current stable WordPress Core.

kienstra:
There's a little more space below the Image control and the next block. Do you think we should 'correct' for this?

# WP Core Current Stable

un-poco-menos

# Upcoming WP Core

un-poco-mas

kienstra:
The block editor in WP Core 5.3-RC3 now looks good, other than the points above.

I tested the functionality of all of the blocks with WP 5.3-RC3:

* In the main editor
* In the Inspector Controls
* In a repeater
* With all 25%, 50%, 75%, and 100% width

I didn't see an issue, other than those mentioned above. Though of course this could use more testing.

kienstra:
**Request For Review**

Hi @lukecarbis,
Hope you had a great weekend. Could you please review this? 

Thanks, Luke!

README.md
```diff
@@ -3,7 +3,7 @@
 Contributors: lukecarbis, ryankienstra, Stino11, rheinardkorf
 Tags: Gutenberg, Blocks
 Requires at least: 5.0
-Tested up to: 5.2
+Tested up to: 5.3
```
kienstra, line 6: It should be fine to bump this to `5.3.` whenever. The AMP plugin deployed last week [with this value](https://github.com/ampproject/amp-wp/blob/394b2023aaf7bb83146c3968d5b43e8be194cac7/readme.txt#L5), and its 'Tested up to' on wp.org [is 5.3](https://wordpress.org/plugins/amp/).

lukecarbis:
Here's the Block I'm using to test: https://gist.github.com/lukecarbis/0a63be52b4d0c13f3951d9addf78896a

lukecarbis:
Testing in `5.3-RC4-46673`.

css/src/editor.scss
lukecarbis: The 25% field width is still useful if you want to, say, have 4 text fields in a row. But having seen the fields taking up 100% of the space, I like it better the way you've got it in this PR.

lukecarbis:
@kienstra This is a great PR. Fixes pretty much all the issues for me. Couple of things I noticed.

**1. Fonts**

Font used by some elements (including the help text) can be overridden by the theme, which is unnecessary in this "editor" style view.

Screen Shot 2019-11-07 at 10 12 34 am

Screen Shot 2019-11-07 at 10 17 02 am
_Using the TwentyTwenty theme_

To fix this, we should add a style rule similar to this:
```CSS
.block-form {
    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
}
```

**2. Image Field with small editor widths**

TwentyTwenty and a few other themes I tried set the editor width to 610px wide. When this happens, a 25% Image field takes up more room than 25%, creating unexpected layouts.

Screen Shot 2019-11-07 at 10 21 08 am

While this isn't really possible to fix properly, CSS like this would help make it work with editor widths as thin as 610px.

```CSS
.bl-image__placeholder .components-button.is-button {
    white-space: normal;
}
```

lukecarbis (approved):

kienstra:
Hi @lukecarbis, thanks for your great code suggestions. I committed both of them in https://github.com/getblocklab/block-lab/pull/467/commits/c9abd5972b64725410dbff5cff189fd28ce0dc3c. 

font-looks-better


If by chance that commit isn't what you had in mind, please let me know 😄

kienstra:
>Here's the Block I'm using to test: https://gist.github.com/lukecarbis/0a63be52b4d0c13f3951d9addf78896a

That's another great thing about the PHP API, it's much easier to share our setup for tests.

PR #470, opened 2019-11-08
Feature/remove onboarding help prompt after adding first block

### LINKED ISSUE #464: Onboarding help prompts persisting after adding first block
Started a new site with a new install of BL. Added my first block etc, but the onboarding prompts are hanging around.

image

byronkeet:
I found the issue affecting everything which was that a conditional statement was missing that made the example or 'tutorial' message appear on every block that was created. I fixed this by adding the condition that it must ONLY show on on the example block. This fixed all issues and now the location message shows correctly as well on each block even if the example block 'tutorial' wasn't followed and/or not deleted.

Fixes #464

### DIFF
--- a/php/admin/class-onboarding.php
+++ b/php/admin/class-onboarding.php
@@ -89,7 +89,7 @@ class Onboarding extends Component_Abstract {
 		/*
 		 * On the edit post screen, editing the draft Example Block.
 		 */
-		if ( $slug === $screen->id && 'post' === $screen->base ) {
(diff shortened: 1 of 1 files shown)


lukecarbis:
Thanks for this PR!

This fixes the button on the first Welcome onboarding screen for me.

But it doesn't fix the issue with the additional notices appearing unexpectedly when creating a new block.



Could you take another look?

lukecarbis:
@byronius-maximus I just re-reviewed this, and it's working for me now. 🤷‍♂

lukecarbis (approved):

PR #478, opened 2019-11-24
[WIP] React refactoring of the 'Edit Block' UI 

### LINKED ISSUE #469: Refactor 'Edit Block' UI to use React
The 'Edit Block' UI has served the plugin well so far. There have been very few bugs, and I think it's a great UI.

edit-block-ui

But as the UI adds more features, it's going to get harder to maintain. There's a [jQuery-based file](https://github.com/getblocklab/block-lab/blob/204fc296c8167288afc6f7b654610668e86bd44c/js/admin.block-post.js) for it that's over 500 lines.

And some things could be really hard without React. Like [Field conditions](https://github.com/getblocklab/block-lab/issues/278). 

I think there's a risk that if the [jQuery-based file](https://github.com/getblocklab/block-lab/blob/204fc296c8167288afc6f7b654610668e86bd44c/js/admin.block-post.js) file gets much bigger, we might never find time to refactor it.

Still, this might not reduce the lines of code, I just think it'll be easier to maintain. And it could have more bugs than the previous implementation, which had very few.

I'm not proposing here that we change the actual design or UI any more than necessary.

kienstra:
## WIP

There's a lot of work remaining, including porting the previous PHP rendering and JS logic

Closes #469


kienstra:
Here's the current state:

current-editor

kienstra:
**More PHP ported**

react-rendered

js/src/index.js
```diff
@@ -1 +0,0 @@
-// Import components.
```
kienstra, line 1: This didn't look to be used.

php/post-types/class-block-post.php
```diff
@@ -954,241 +896,6 @@ public function ajax_field_settings() {
...
 
-	/**
-	 * Save block meta boxes as a json blob in post content.
-	 *
-	 * @param array $data An array of slashed post data.
-	 *
-	 * @return array
-	 */
-	public function save_block( $data ) {
-		if ( ! isset( $_POST['post_ID'] ) ) {
```
kienstra, line 965: The saving will probably be handled client-side now.

kienstra:
This now has a basic way of getting the post content on the client-side and rendering field rows from it.

But so far, this only works with existing blocks. I still need to add the ability to add rows. And editing fields doesn't work yet.

rendering-existing-blocks

webpack.config.js
```diff
@@ -40,7 +40,7 @@ const extractConfig = {
 module.exports = {
 	entry: {
 		'./js/editor.blocks': './js/blocks/index.js',
-		'./js/scripts': './js/src/index.js',
+		'./js/admin.edit-block': './js/src/edit-block.js',
```
kienstra: It looks like `js/src/index.js` wasn't used.

kienstra:
**Current State**

There's a lot of work left. But it's now possible to add fields, enter basic values, 'autoslug' the name, and save the field.

react-ui

PR #479, opened 2019-11-24
[Do not merge] Remove additional styles added to the block editor

lukecarbis:
@kienstra This is a small suggestion for the react refactor. It's kind of a hack, but it overcomes the problem of themes modifying the "style" of the Block Editor.

This is intended just as an idea – no need to merge if you don't think it's needed.

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -48,6 +48,15 @@ class Block_Post extends Component_Abstract {
 	);
 
 	/**
+	 * The core registered styles that are loaded before any additional block editor styles.
+	 *
+	 * We need to keep track of this so that we can remove styles that aren't needed.
+	 *
+	 * @var array
+	 */
+	private $core_styles = array();
+
+	/**
 	 * Block Post constructor.
 	 */
 	public function __construct() {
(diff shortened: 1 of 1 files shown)


lukecarbis:
@kienstra I've just realised that this approach won't really work.

Sure, it cuts out any styles that are loaded on the editor screen.

But then, what about plugins that extend Gutenberg in various ways?

Yoast might want to load in the sidebar, ACF might try to insert custom fields somewhere, we don't really need the top toolbar, etc.

Maybe we should consider:

1. Remove `editor` support when defining the post type (!)
2. Hide or remove the "classic" metaboxes on the right hand side, so we've got a blank canvas
3. Add a react `<div>` to render react stuff in
4. Add Gutenberg's sidebar component (inspector? I don't know what it's called)
5. Add our own custom component

This would prevent any extra editor css loading, and it would also stop plugins from extending the block editor interface.

Thoughts?

kienstra:
**Good thoughts**

Hi @lukecarbis,
That's a good idea to consider this. Other plugins adding meta boxes or scripts to the 'Edit Block' UI would probably be either useless or harmful.

>4. Add Gutenberg's sidebar component (inspector? I don't know what it's called)

It looks like the `<SettingsSidebar>` component imports [the components it uses](https://github.com/WordPress/gutenberg/blob/58a78e6f81b148ab5460a11d84b94578a7b4b370/packages/edit-post/src/components/sidebar/settings-sidebar/index.js#L12-L23
).

So to use that, we'd have to fork all of those components (or the ones we want).

nent-ssidebar

For the top toolbar, of course the block icon makes no sense for the 'Edit Block' UI. But undo and redo should work in #478, once it's finished:

pr-rea

Also, we could [filter out](https://github.com/WordPress/wordpress-develop/blob/6cf77dbbb6d6b289d81dd2d33c742d150129b774/src/wp-admin/includes/post.php#L2252) all meta boxes but ours.

One great thing about Gutenberg is that it has a good API for [getting](https://github.com/getblocklab/block-lab/pull/478/files#diff-a5bbefd2ff269cbba377da9b5690280cR81) and [changing](https://github.com/getblocklab/block-lab/pull/478/files#diff-a5bbefd2ff269cbba377da9b5690280cR88) the post content client-side. 

As you know, the classic editor, like we're using right now, relies on traditional form elements for POST requests, like `<input name="foo" id="foo">`.

It's probably possible to build our new UI on the classic editor foundation. I might even find a blocker that makes that necessary.

But for now, it looks like Gutenberg gives a lot of advantages for a React-based UI.

PR #484, opened 2019-11-27
Add a `<p>` about where the JSON file should come from

kienstra:
This adds some more explanation to the Block Lab importer:

>This JSON file should come from the export link or bulk action in the Content Blocks screen, not from the main Export tool.

this-is-added

This clarifies that it's not possible to export the blocks from here:

import-from-here

...and then import them using this:

sing-impor

Maybe this isn't the best wording, feel free to suggest something else.

### DIFF
--- a/php/admin/class-import.php
+++ b/php/admin/class-import.php
@@ -96,8 +96,8 @@ class Import extends Component_Abstract {
 			<?php
 			echo wp_kses(
 				sprintf(
-					/* translators: %1$s: an opening anchor tag, %2$s: a closing anchor tag, %3$s: a different opening anchor tag, %4$s: a closing anchor tag */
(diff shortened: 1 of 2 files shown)


kienstra:
Hi @lukecarbis,
Could you please review this when you have a chance? It's not a very high priority, in my opinion.

lukecarbis (changes requested):
@kienstra This is great. My only suggestion is that we shouldn't link to the Export Tool. It doesn't make sense to provide a way of quickly accessing a discouraged action. Someone skimming might accidentally click the wrong link.

kienstra:
That's a good point, we shouldn't encourage going to the Export tool.

kienstra:
There's now no link to the 'Export tool':

here-link

lukecarbis (approved):

Issue #485, opened 2019-11-28
PHP performance profiling


kienstra:
It'd be good to have an idea of this plugin's effect on front-end performance. 

This might also point out areas that we could improve. For example, if a certain function is taking too long.

kienstra:
# Block Lab Compared To Implementing Your Own Dynamic Blocks

### What's the PHP performance cost of using Block Lab for 1 block on the front-end (not the block editor), compared to implementing your own dynamic block?

About 87 milliseconds. For reference, this is slightly less than the cost for [Gutenberg](https://github.com/wordPress/gutenberg/) or [Yoast](https://github.com/Yoast/wordpress-seo).

Here's a Blackfire profile of Block Lab:
https://blackfire.io/profiles/e4978c69-4bf2-462e-a195-69fa81d93372/graph

## Methodology

Create 2 posts, one with a Block Lab block, and one with an equivalent dynamic block that doesn't use Block Lab


### With a Block Lab block
1. Activate [Twenty Nineteen](https://wordpress.org/themes/twentynineteen/), and deactivate all plugins except Block Lab, and a plugin that adds the snippets below.

2. Add the Block Lab block:
```php
add_action(
	'block_lab_add_blocks',
	function() {
		block_lab_add_block(
			'test-hero',
			array(
				'title'    => 'Test Hero',
				'category' => 'common',
				'fields'   => array(
					'headline' => array(
						'label'   => 'Headline',
						'control' => 'text',
					),
					'copy'     => array(
						'label'   => 'Copy',
						'control' => 'textarea',
					),
					'url'      => array(
						'label'   => 'URL',
						'control' => 'url',
					),
				),
			)
		);
	}
);
```

2. Then, add the template for it in the theme or plugin, at `blocks/block-test-hero.php`:

```
<?php
/**
 * Template for test-hero block.
 */

?>
<h1><?php esc_html_e( 'Testing Block Lab dynamic block', 'bl-testing' ) ?></h1>
<p><?php esc_html_e( 'Here is the headline attribute:', 'bl-testing' ) ?></p>
<p><?php echo esc_html( block_value( 'headline' ) ); ?></p>
<p><?php esc_html_e( 'And the copy attribute:', 'bl-testing' ) ?></p>
<p><?php echo esc_html( block_value( 'copy' ) ); ?></p>
<p><?php esc_html_e( 'And the url attribute:', 'bl-testing' ) ?></p>
[...]

PR #486, opened 2019-11-29
Refactor import error and success methods

lukecarbis:
Inspired by this comment: https://github.com/getblocklab/block-lab/pull/449#discussion_r349942152

I got a little carried away and refactored our `render_block_import_error` and `render_block_import_success` methods so that they could be used in more places.

PR #487, opened 2019-12-01
Set the posts_per_page to 100

kienstra:
This should help the case that Block Lab performs well, and is scalable (#485).

The `-1` value probably wouldn't get past a VIP review. 

Ideally, people creating more than 100 blocks in the 'Edit Block' UI would use the PHP API.

### DIFF
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -554,7 +554,7 @@ class Loader extends Component_Abstract {
 			[
 				'post_type'      => block_lab()->get_post_type_slug(),
 				'post_status'    => 'publish',
-				'posts_per_page' => - 1,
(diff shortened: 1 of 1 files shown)


kienstra:
**Request For Review**

Hi @lukecarbis,
Hope you had a great weekend.

Sorry, I still owe you a review for #486.

When you have a chance, could you please review this? It should help our case that Block Lab performs well and scales (#485).

lukecarbis (approved):
Looks good to me.

lukecarbis:
@kienstra Maybe consider adding a comment on that line to explain why it's `100`, not `-1`.

kienstra:
Sure, that's a good point.

PR #489, opened 2019-12-03
Add rulesets to phpcs.xml, correct some phpcs issues from them

kienstra:
* Adds more rulesets to `phpcs.xml`, including checks for PHP compatibility and `'block-lab'` textdomain arguments for functions like `__()`
* Adds PHP 7.3 and 7.4 to the Travis build matrix
* Converts `array()` syntax to `[]`


composer.json
```diff
@@ -9,8 +9,10 @@
     },
     "require-dev": {
         "brain/monkey": "^2",
+        "dealerdirect/phpcodesniffer-composer-installer": "0.5.0",
```
kienstra, line 12: Without this, sometimes there are error messages when doing `npm run lint:php`, like:

>ERROR: Referenced sniff "WordPress" does not exist

phpcs.xml
```diff
@@ -1,22 +1,39 @@
...
+
+	<arg value="s"/>
+	<arg name="extensions" value="php"/>
+
+	<file>.</file>
+	<exclude-pattern>*/node_modules/*</exclude-pattern>
+	<exclude-pattern>*/vendor/*</exclude-pattern>
+	<exclude-pattern>*/package/*</exclude-pattern>
+	<exclude-pattern>*.js</exclude-pattern>
+	<exclude-pattern>*.css</exclude-pattern>
```
kienstra, line 38: A lot of this diff is converting the file from spaces to tabs.

php/admin/class-settings.php
```diff
@@ -122,7 +122,7 @@ public function prepare_notice( $notice ) {
 	 * Show any admin notices after saving the settings.
 	 */
 	public function show_notices() {
-		$notices = get_option( 'block_lab_notices', array() );
+		$notices = get_option( 'block_lab_notices', [] );
```
kienstra, line 125: I realized that by changing `phpcs.xml`, doing `npm run lint:php:fix` can convert all of the `array()` syntax to `[]`.

This produced a huge diff, and will of course need regression testing.

But I think it's important to use this more modern syntax. The `[]` syntax [started in PHP 5.4](https://www.php.net/manual/en/language.types.array.php), and we only support PHP 5.6+.

kienstra:
Hi @lukecarbis, great talking to you earlier.

Do you like the `[]` syntax here instead of `array()`? 

I think it's important to use the `[]` syntax for the plugin to be more modern, but that may be a personal preference.

lukecarbis:
I prefer `array()`, but only because that's what I've always used / seen / been taught.

I'm all for making the plugin more modern, though. I have seen `[]` around the place. Could you point me to a couple of good examples of modern WordPress plugins that prefer this syntax?

lukecarbis:
Also, WP core uses `array()`.

kienstra:
Right, Core does use `array()`. 

It looked strange to me also before the AMP plugin started using it:

https://github.com/ampproject/amp-wp/blob/09f1461b896c921d4afd2ffbeee1597043c890e8/includes/class-amp-theme-support.php#L105

I'm looking for more plugin examples.

lukecarbis:
Sorry, you'll have to find something more than a Google project to impress me. ;)

kienstra:
Haha 😄 

Yoast: https://github.com/Yoast/wordpress-seo/blob/9c43132323adfcc7e748a1f30af994c50b62f89e/inc/health-check.php#L149

kienstra:
Calypso uses `[]`: https://github.com/Automattic/wp-calypso/blob/fba2d80dc733828ff3fff47af86f0944cca83bff/apps/full-site-editing/full-site-editing-plugin/global-styles/includes/class-data-set.php#L20

To be honest, there are probably more plugins using `array()` than `[]`. [Jetpack](https://github.com/Automattic/jetpack/) (at least mainly) uses `array()`.

But I think it makes our plugin look more modern, and there's no concern for PHP supporting it, as we support 5.6+, while `[]` goes back to `5.4`. 

Still, I'm not passionate about this. It'd be easy to revert the one commit on this PR and discuss this separately.

lukecarbis:
Let's leave it in for now. I've asked around on Twitter and Slack, see what a few other people say. I'm also happy to be a trend-setter and just push for a new standard here too.

kienstra:
OK, sounds good. I need to do some good regression testing for this PR. It might be a few days, at least.

lukecarbis:
Screen Shot 2019-12-05 at 8 27 59 am
Screen Shot 2019-12-05 at 8 28 19 am

Looks like the answer is pretty clear.

Worth mentioning this response, from @peterwilsoncc, on why this hasn't been adopted in WordPress core:

> It was reviewed but long form was decided upon for the reason that people with impaired vision can have trouble seeing the difference between [(

Also, @visser voted twice. 🙄

lukecarbis:
@kienstra If you put the letters WIP in the title of the PR (I usually use `[WIP]` at the start), we have a WIP check that will prevent merging this PR accidentally.

kienstra:
Nice, thanks for tweeting that poll!

kienstra:
>@kienstra If you put the letters WIP in the title of the PR (I usually use [WIP] at the start), we have a WIP check that will prevent merging this PR accidentally.

Good point that it's safer, as it blocks merging.

I've avoided doing that because there's no green check build icon, so at first glance it seems like the builds are failing. Of course, it's not hard to see that it's just because of the 'WIP' Travis check.

php/views/license.php
```diff
@@ -33,9 +33,9 @@
 						$count = $license['site_count'];
 					}
 
-					$expiry = date( get_option( 'date_format' ) );
+					$expiry = gmdate( get_option( 'date_format' ) );
```
kienstra, line 36: This is one of the few changes to a PHP file in this PR, other than `array()` to `[]`.

This appears in 8 other places, and is in response to a PHPCS error.

kienstra:
**Request For Review**

Hi @lukecarbis,
Could you please review this when you can?

* Adds more rulesets to `phpcs.xml`, including checks for PHP compatibility. 
* Adds PHP 7.3 and 7.4 to the `.travis.yml` build matrix
* Changes `array()` to `[]` 

Regression testing looks good. I created a new block, and it displayed and saved as expected in the block editor:

new-block

And a block with almost all of the fields saved its values and displayed them as expected:
fields-block

Also, saving a license key worked as expected, and the Pro fields were available.

Thanks, Luke!

lukecarbis (approved):
It's going to take some time for me to get used to square bracket arrays, but that's okay! Looking good.

kienstra:
Thanks, @lukecarbis!

PR #492, opened 2019-12-18
Update versions in package.json, use @wordpress/scripts for build process

### LINKED ISSUE #266: Align build workflow with that of Gutenberg 5.3+
New official recommended build workflows have been introduced with Gutenberg 5.3. It would be good to align the BlockLab build process.

See: https://make.wordpress.org/core/2019/03/25/building-javascript/

kienstra:
* There are some packages that are very out of date...I should have been better about updating them.
* Updates the packages in `package.json` to the latest version in https://www.npmjs.com/. For example, https://www.npmjs.com/package/gulp-run.
* Also uses the [recommended](https://make.wordpress.org/core/2019/03/25/building-javascript/) `@wordpress/scripts` package in the build process. This will make Block Lab more in line with Gutenberg-based plugins.
* Adds the `package-lock.json` file to the repo. This should ensure consistency of versions. For example, in `package.json`, the `gulp` version is `^4.0.2`. The `^` means that doing `npm install` can update `gulp` to `4.1.0` or `4.2.0` if and when those exist. So `package-lock.json` should ensure that we have the same package tree.

Closes #266


js/blocks/components/advanced-controls.js
```diff
@@ -20,7 +20,7 @@ const AdvancedControls = ( { block } ) => {
 
 	const tip = sprintf(
 		__( 'The Additional CSS Class can be included in your block template with the %1$s field.', 'block-lab' ),
-		'<code>className</code>'
+		'<code>className</code>',
```
kienstra: These trailing commas aren't related to this PR, but are from doing `npm run lint:js:fix`

webpack.config.js
```diff
@@ -1,43 +1,11 @@
 const path = require( 'path' );
-const ExtractTextPlugin = require( 'extract-text-webpack-plugin' );
+const MiniCssExtractPlugin = require( 'mini-css-extract-plugin' );
```
kienstra, line 2: The [extract-text-webpack-plugin](https://www.npmjs.com/package/extract-text-webpack-plugin) is deprecated, and it recommends using [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin). Most of the diff in this file is due to the different syntax for that. 

I'm not sure I got the syntax right, so I'll do some regression testing on the built JS and CSS.

webpack.config.js
```diff
@@ -57,19 +25,42 @@ module.exports = {
 					loader: 'babel-loader',
 				},
 			},
-			{
-				test: /style\.s?css$/,
```
kienstra, line 61: It looks like this plugin doesn't use `style.css` or `style.scss`. I couldn't find a reference to it, though maybe I missed something.

.gitignore
```diff
@@ -88,6 +88,7 @@ package-lock.json
 *.swp
 *.tmp
 *.bak
+*.map
```
kienstra, line 90: The new build process creates `.map` files, like `block-lab/css/blocks.editor.css.map`.

kienstra:
To test this, I've been deleting `block-lab/css/blocks.editor.css` and `block-lab/js/editor.blocks.js`, and ensuring that `npm run dev` rebuilds those files as expected.

kienstra:
I still need to do a lot of regression testing here, as it upgrades almost every package in `package.json`, and refactors the build system.

kienstra:
@todo: set up something like `dependabot` to automatically open PRs when packages need to be updated.

See #493

js/admin.block-post.js
```diff
@@ -352,7 +352,7 @@
 			const postType = $( input ).val();
 			if ( -1 === excludedPostTypes.indexOf( postType ) ) {
 				displayList.push(
-					$( input ).next( 'label' ).text()
+					$( input ).next( 'label' ).text(),
```
kienstra: I should probably revert this, as this file isn't compiled. And [support](https://caniuse.com/#feat=mdn-javascript_grammar_trailing_commas_trailing_commas_in_functions) for trailing commas in functions doesn't look good enough.

.eslintrc
```diff
@@ -10,6 +10,22 @@
...
     "@wordpress/dependency-group": "error",
-    "@wordpress/react-no-unsafe-timeout": "error"
+    "@wordpress/react-no-unsafe-timeout": "error",
+    "comma-dangle": [
+      "error",
+      {
+        "arrays": "always-multiline",
+        "objects": "always-multiline",
+        "imports": "only-multiline",
+        "exports": "only-multiline",
```
kienstra: This allows the comma after `bar`, but doesn't require it:
```js
const {
    foo,
    bar, 
} = wp.components;
```

.eslintrc
```diff
@@ -10,6 +10,22 @@
   },
   "rules": {
     "@wordpress/dependency-group": "error",
-    "@wordpress/react-no-unsafe-timeout": "error"
+    "@wordpress/react-no-unsafe-timeout": "error",
+    "comma-dangle": [
+      "error",
+      {
+        "arrays": "always-multiline",
+        "objects": "always-multiline",
```
kienstra, line 18: This requires the trailing comma after 'bar' here:

```js
const foo = {
    one: 'baz',
    two: 'bar',
};
```

...but doesn't allow a trailing comma after 'bar' here:

```js
const foo = { one: 'baz', two: 'bar' };
```

.eslintrc
```diff
@@ -10,6 +10,22 @@
...
+        "imports": "always-multiline",
+        "exports": "always-multiline",
+        "functions": "never"
```
kienstra, line 21: A trailing comma after the last argument looks strange to me: `doThis( 'arg1', 'arg2', )`

...and it doesn't have great [browser support](https://caniuse.com/#feat=mdn-javascript_grammar_trailing_commas_trailing_commas_in_functions). This `.eslintrc` is also used for a non-compiled file: `admin.block-post.js`. 

Ideally, soon that file will be ported to the React logic in the refactoring.

.eslintrc
```diff
@@ -10,6 +10,22 @@
...
+      }
+    ],
+    "no-unused-vars": [
+      "error",
+      {
+        "ignoreRestSiblings": true
+      }
+    ]
```
kienstra, line 29: This file could probably use more rules, but that's not in scope for this PR.

kienstra:
**CSS and JS Compilation**

Because this edits `webpack.config.js` so much, it's important to check that the CSS is compiling correctly to `blocks.editor.css`.

[Here's a diff](https://www.diffchecker.com/MVWNukSa) of the compiled `blocks.editor.css` file. On the left is the `develop` branch, and the right is this PR's branch. This is from doing `npm install && npm run build` on both branches.

Much of the diff is from vendor prefixes:
transition-webkit

So far, I didn't see a CSS or JS problem on:

MacOS with Chrome 79
Windows 10 with Edge 18.0 (with BrowserStack)

I tested all fields, including the repeater. It's possible that I missed a styling issue, but I didn't see anything obvious.

kienstra:
**Request For Review**

Hi @lukecarbis,
Good seeing you earlier. Could you please review this when you have a chance?

It updates the packages in `package.json`, and uses the [Gutenberg-recommended](https://make.wordpress.org/core/2019/03/25/building-javascript/) build process: `wp scripts start`.

Regression testing of all of the fields looks good so far. 

Thanks, Luke!

package.json
```diff
@@ -2,50 +2,65 @@
...
   "title": "Block Lab",
   "version": "1.5.2",
+  "description": "WordPress plugin with a simple templating system for building custom blocks.",
+  "author": "Block Lab",
+  "license": "GPL-2.0-or-later",
+  "keywords": [
+    "block-editor",
+    "wordpress",
+    "wordpress-plugin"
+  ],
```
kienstra, line 12: These edits are from the `lint:pkg-json` script below.

package.json
```diff
@@ -2,50 +2,65 @@
   "name": "block-lab",
   "title": "Block Lab",
   "version": "1.5.2",
+  "description": "WordPress plugin with a simple templating system for building custom blocks.",
```
kienstra, line 5: Feel free to suggest something different, I added this because there was an error from not having a description.

js/blocks/components/advanced-controls.js
```diff
@@ -25,12 +25,12 @@ const AdvancedControls = ( { block } ) => {
...
 	return (
 		<InspectorAdvancedControls key={ `inspector-advanced-controls-${ block.name }` }>
-			<DotTip tipId="block-lab/additional-css-class">
+			<Guide tipId="block-lab/additional-css-class">
 				<p className="bl-dot-tip" dangerouslySetInnerHTML={ { __html: tip } }></p>
 				<p className="bl-dot-tip read-more">
 					<a href="https://getblocklab.com/docs/faqs/" target="_blank" rel="noopener noreferrer">{ __( 'Read more', 'block-lab' ) }</a>
 				</p>
-			</DotTip>
+			</Guide>
```
kienstra: This is one of the reasons that this PR now imports modules, like `@wordpress/components`, instead of using something like:

```js
const { Guide } = wp.components;
```

The `nux` package was removed, and the `DotTip` was moved to the `components` package. So we can control which packages we're using, and don't need conditionals like:

```js
const { DotTip } = wp.nux ? wp.nux.DotTip : wp.components.Guide;
```

kienstra: Also, there used to be deprecation notices of components, which we would have to deal with at some point.

kienstra:
Just opened https://github.com/getblocklab/block-lab/issues/497 to possibly consider an alternative to the DotTip, which is removed in https://github.com/getblocklab/block-lab/pull/492/commits/440f50abd405e010193cec4258904fd5d686cd48

js/blocks/components/fields.js
```diff
@@ -1,8 +1,8 @@
 /**
  * WordPress dependencies
  */
-const { applyFilters } = wp.hooks;
-const { select } = wp.data;
+import { applyFilters } from '@wordpress/hooks';
```
kienstra, line 4: The reason for importing these as modules instead of using something like `wp.hooks` is so that we can control the module versions.

For example, `InspectorControls` and `InspectorAdvancedControls` were moved from `wp.editor` to `wp.block-editor`. And `wp.nux` was deprecated entirely.

If we control the versions of the modules, we should be less dependent on the version of Core and Gutenberg users have.

kienstra: A user reported an error from `wp.nux.DotTips` being deprecated in Gutenberg. I probably would have seen that if we imported that deprecated module.

kienstra:
**Request For Review**

Hi @lukecarbis,
When you have a chance, could you please review this? With the latest commits, it's ready for review.

https://github.com/getblocklab/block-lab/pull/492/commits/440f50abd405e010193cec4258904fd5d686cd48 in this PR fixes an issue raised in the #support Slack channel, and in https://wordpress.org/support/topic/blocks-not-visible-in-admin-browser-console-error-wp-nux-not-defined/

Regression testing of all fields looks good. Though of course this is a huge diff.

Thanks, Luke!

lukecarbis (approved):
TBH I don't understand most of this, but it mostly seems to work. The parts that don't could be my machine. After running `npm install`, I also tried:

- npm run check-licenses ✅
- npm run dev ✅
- npm run build ✅
- npm run lint ❌ (`ERROR: Referenced sniff "WordPress-Core" does not exist`)
- npm run lint:js ✅
- npm run lint:pkg-json ✅
- npm run lint:php ❌ (`ERROR: Referenced sniff "WordPress-Core" does not exist`)

lukecarbis:
FWIW, if I run `phpcs -i`, this is the output:

```
The installed coding standards are PEAR, Zend, PSR2, MySource, Squiz, PSR1, PSR12, WordPress, WordPress-Extra, WordPress-Docs, WordPress-Core, PHPCompatibility, PHPCompatibilityParagonieRandomCompat, PHPCompatibilityParagonieSodiumCompat and PHPCompatibilityWP
```

WordPress-Core is obviously installed on my system.

kienstra:
Thanks for bringing that up, it's important that `npm run lint` works. I'll look at that.

kienstra:
Hi @lukecarbis,
Thanks a lot for reviewing this. Maybe you tried this already, but does that error still exist after running:

```sh
rm -rf vendor && composer install && npm run lint:php
```

There's a [package](https://github.com/getblocklab/block-lab/blob/20e0b89fcc852a5585d3af905658cdc2eb397dc2/composer.json#L12) in `composer.json` that should ensure that `WordPress-Core` is available.

lukecarbis:
Just tried the following:
```
$ rm -rf vendor
$ rm -rf composer.lock
$ composer install
$ npm run lint:php
```

That fixed the problem. Note, I had to also remove the composer.lock file.

Ready to merge!

kienstra:
Ah, thanks. Good idea to remove `composer.lock`.

kienstra:
Here's a build of the `develop` branch, probably very close to what the next version will be.

[block-lab.zip](https://github.com/getblocklab/block-lab/files/4074481/block-lab.zip)

PR #495, opened 2019-12-31
Use get_post() instead of the_post() as an argument

kienstra:
Sometimes, there is a fatal error on saving a Block Lab post. But I couldn't reproduce this on a machine with PHP 7.2.

## Steps to reproduce
1. Create a new Block Lab block
2. Name the block
3. Add a field and give it any name
4. Save the block
setup-error-here

5. Expected: the block should save
6. Actual: there's a PHP error:
>( ! ) Notice: Trying to access array offset on value of type null in bl/wp-includes/class-wp-query.php on line 3252

error-here

Block Lab branch: `develop`
WP Core: 5.3, 5.3.1, 5.3.2 (haven't tested others)
PHP version: 7.4.1 (didn't occur on another machine with 7.2.25)

### DIFF
--- a/php/post-types/class-block-post.php
+++ b/php/post-types/class-block-post.php
@@ -1040,7 +1040,7 @@ class Block_Post extends Component_Abstract {
 		// Block category.
 		if ( isset( $_POST['block-properties-category'] ) ) {
 			$category_slug = sanitize_key( $_POST['block-properties-category'] );
-			$categories    = get_block_categories( the_post() );
+			$categories    = get_block_categories( get_post() );
 
 			if ( '__custom' === $category_slug && isset( $_POST['block-properties-category-name'] ) ) {
 				$category = [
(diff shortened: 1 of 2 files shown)


php/post-types/class-block-post.php
```diff
@@ -1040,7 +1040,7 @@ public function save_block( $data ) {
 		// Block category.
 		if ( isset( $_POST['block-properties-category'] ) ) {
 			$category_slug = sanitize_key( $_POST['block-properties-category'] );
-			$categories    = get_block_categories( the_post() );
+			$categories    = get_block_categories( get_post() );
```
kienstra, line 1043: It looks like [the_post()](https://developer.wordpress.org/reference/functions/the_post/) doesn't return a value, so [get_post()](https://developer.wordpress.org/reference/functions/get_post/) should work.

kienstra:
**Request For Review**

Hi @lukecarbis,
Happy New Year! 

Could you please review this?

It might fix the problem here, but I'm waiting for a response:
https://wordpress.org/support/topic/add-field-not-working/

Thanks, Luke!

lukecarbis (approved):

PR #499, opened 2020-01-14
Add a pre-commit hook that only lints staged JS and PHP files

kienstra:
#### Changes

* Inspired by Weston Ruter's example pre-commit hook
* Uses composer-git-hooks to run this
* So on running composer install, this hook will be installed
* This only lints staged JS and PHP files, so it can run quickly

#### Testing Instructions
1. do `composer install`
2. Create an intentional `phpcs` error in any PHP file, like adding an extra space inside parentheses
3. Stage the change
4. Attempt to commit it
5. The pre-commit hook should prevent committing it
6. Repeat steps 2-5, but for a file in `js/`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,10 +9,20 @@
     },
     "require-dev": {
         "brain/monkey": "^2",
+        "brainmaestro/composer-git-hooks": "^2.8",
         "dealerdirect/phpcodesniffer-composer-installer": "0.5.0",
         "mockery/mockery": "^1.2.4",
         "phpcompatibility/phpcompatibility-wp": "2.1.0",
         "squizlabs/php_codesniffer": "^3.4",
         "wp-coding-standards/wpcs": "2.2.0"
+    },
+    "extra": {
+        "hooks": {
+            "pre-commit": "bash bin/pre-commit.sh"
+        }
+    },
+    "scripts": {
+        "post-install-cmd": "cghooks add --ignore-lock",
+        "post-update-cmd": "cghooks update"
     }
 }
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -9,10 +9,20 @@
     },
     "require-dev": {
         "brain/monkey": "^2",
+        "brainmaestro/composer-git-hooks": "^2.8",
```
kienstra, line 12: This package installs the pre-commit hook on running `composer install` or `composer update`. Otherwise, users would probably have to manually install it, or run a bash script to install it.

composer.json
```diff
@@ -9,10 +9,20 @@
...
         "wp-coding-standards/wpcs": "2.2.0"
+    },
+    "extra": {
+        "hooks": {
+            "pre-commit": "bash bin/pre-commit.sh"
+        }
+    },
+    "scripts": {
+        "post-install-cmd": "cghooks add --ignore-lock",
+        "post-update-cmd": "cghooks update"
```
kienstra, line 26: I think this means that on running `composer update`, `cghooks` will update the pre-commit hook.

package.json
```diff
@@ -71,6 +71,7 @@
     "build": "wp-scripts build",
     "lint": "npm-run-all --parallel lint:*",
     "lint:js": "wp-scripts lint-js js",
+    "lint:js:files": "wp-scripts lint-js",
```
kienstra, line 74: This command allows passing specific files, like `npm run lint:js:files js/foo.js`. 

`npm run lint:js` lints all of the JS files in `js/`.

lukecarbis (approved):
I completed the steps as suggested, and could not commit my PHPCS mistake. Nice work!

kienstra:
Thanks, Luke!

package.json
InfiniteLove2020: n and p to navigate between commits in a pull request.

PR #502, opened 2020-01-17
Version bump to 1.5.3

kienstra:
### Changes
* Updates the version in the normal places for [version bump PRs](https://github.com/getblocklab/block-lab/pull/488)
* Adds an entry in `CHANGELOG.md` 
* Fixes so Travis passes

### Testing
*(in addition to normal regression testing)*
1. `npm install && composer install`
2. Repeat testing steps in https://github.com/getblocklab/block-lab/pull/499#issue-362447992

PR #507, opened 2020-01-25
Prevent an edge case notice, and a unit test

kienstra:
Once, I think when importing a block, this notice appeared:
>Notice: Trying to get property 'post_type' of non-object in /path/to/wp-content/plugins/block-lab/php/blocks/class-loader.php on line 183

This looks to be from `get_post()` returning something like `null`, and trying to access the `->post_type` property of that.

This is probably an edge case, I didn't see it before.

Sorry, I don't remember the steps to reproduce this.

#### Changes
* Handles a case of the global `$post` being `null`, or not having a `->post_type`
* Adds a basic test

### DIFF
--- a/php/blocks/class-loader.php
+++ b/php/blocks/class-loader.php
@@ -132,8 +132,7 @@ class Loader extends Component_Abstract {
 	 * Launch the blocks inside Gutenberg.
 	 */
 	protected function editor_assets() {
-		$asset_config_file = $this->plugin->get_path( 'js/editor.blocks.asset.php' );
-		$asset_config      = require $asset_config_file;
+		$asset_config = require $this->plugin->get_path( 'js/editor.blocks.asset.php' );
 
 		wp_enqueue_script(
 			'block-lab-blocks',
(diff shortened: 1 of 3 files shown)


php/blocks/class-loader.php
```diff
@@ -150,22 +149,6 @@ protected function editor_assets() {
...
-		);
-
-		$block_names = wp_list_pluck( $this->blocks, 'name' );
-
-		foreach ( $block_names as $block_name ) {
-			$this->enqueue_block_styles( $block_name, [ 'preview', 'block' ] );
-		}
-
-		$this->enqueue_global_styles();
-
```
kienstra, line 168: This is simply moved below, so the script-related logic is first, then the style logic. There's no change to it.

php/blocks/class-loader.php
```diff
@@ -177,19 +160,30 @@ protected function editor_assets() {
...
-		$post      = get_post();
-		$post_type = $post->post_type;
-
 		wp_localize_script(
 			'block-lab-blocks',
 			'blockLab',
 			[
 				'authorBlocks' => $author_block_slugs,
-				'postType'     => $post_type,
+				'postType'     => get_post_type(), // To conditionally exclude blocks from certain post types.
```
kienstra, line 168: This is the main edit in the PR. `get_post_type()` will return `false` if `get_post()` returns something falsey. This should prevent a PHP notice.

kienstra:
Travis should pass once we merge #502 and merge in the `develop` branch to this.

kienstra:
A review can wait until #502 is merged.

php/blocks/class-loader.php
```diff
@@ -132,8 +132,7 @@ protected function get_callback( $method_name ) {
 	 * Launch the blocks inside Gutenberg.
 	 */
 	protected function editor_assets() {
-		$asset_config_file = $this->plugin->get_path( 'js/editor.blocks.asset.php' );
-		$asset_config      = require $asset_config_file;
+		$asset_config = require $this->plugin->get_path( 'js/editor.blocks.asset.php' );
```
kienstra, line 135: This isn't related to this PR, it's just something I noticed could be more concise.

lukecarbis (approved):

PR #508, opened 2020-01-28
Use a virtual filesystem in the PHPUnit tests

kienstra:
* Some tests wrote to the actual filesystem, like in the `WP_TESTS_DIR` directory theme
* For example, the integration tests add template files to it in `example-theme/blocks/`
* This can create problems, when sometimes the cleanup doesn't work:
links-here

#### Changes

- [x] Use a virtual filesystem for the integration tests
- [ ] Use it for all other tests that write to the filesystem

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,20 +9,11 @@
     },
     "require-dev": {
         "brain/monkey": "^2",
-        "brainmaestro/composer-git-hooks": "^2.8",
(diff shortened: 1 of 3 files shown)


kienstra:
In the future, it'd be good to prevent the unit tests from having to bootstrap WordPress, like https://github.com/xwp/block-scaffolding-wp

lukecarbis (approved):

PR #514, opened 2020-02-09
Add component Jest tests

kienstra:
#### Changes
* This uses the [react-testing-library](https://testing-library.com/docs/react-testing-library/).
* These component tests don't look to be flaky. They've been passing when they should in this repo and in another [one of my plugins](https://github.com/kienstra/augmented-reality).
* Ideally, all of the components would have tests. But I think this PR is a small improvement.

#### Testing instructions
The only non-test files that this touches are `fetch-input.js` and `repeater-rows.js`. So this is just regression testing for those.

1. Create a Block Lab block with a Repeater
2. Populate the fields, and add a second row
3. Enter values for the fields
4. Ensure that the values save, and they appear in the `<ServerSideRender>` and the front-end
5. Repeat steps 1-4 for Post and Taxonomy fields


js/blocks/components/fetch-input.js
```diff
@@ -13,44 +13,29 @@ import classNames from 'classnames';
...
-		this.suggestionNodes = [];
-
-		this.state = {
-			loading: false,
-			results: [],
-			showSuggestions: false,
-			selectedSuggestion: null,
-		};
-	}
+const FetchInput = ( props ) => {
```
kienstra: Here's the class refactored to be a function with hooks.

js/blocks/components/test/fetch-input.js
```diff
@@ -0,0 +1,80 @@
...
+const baseProps = {
+	value,
+	displayValue,
+	field: { label },
+	onChange: mockOnChange,
+	getValueFromAPI: ( fromAPi ) => fromAPi,
+	getDisplayValueFromAPI: ( fromApi ) => fromApi,
+};
+
+describe( 'FetchInput', () => {
```
kienstra, line 43: The unit tests for these components might be the more interesting part of this PR.

js/blocks/components/fetch-input.js
kienstra: There's a unit test for the component:

```bash
npm install && npm run test:js fetch-input
```

kienstra: The [test-js Travis job](https://travis-ci.com/getblocklab/block-lab/jobs/285249839) passed 5 times in a row, which makes me a little less concerned that these tests will be flaky.

js/blocks/components/fetch-input.js
```diff
@@ -388,7 +388,7 @@ class FetchInput extends Component {
 
 								return !! buttonValue && (
 									<button
-										key={ result.id }
+										key={ `bl-fetch-suggestion-${ index }` }
```
kienstra, line 391: Fixes an error in the tests.

kienstra:
**Request For Testing**

Hi @lukecarbis,
Hope you're having a great day.

There's very little diff to non-test files, just what was necessary.

The [testing instructions above](https://github.com/getblocklab/block-lab/pull/514#issue-372793293) are up-to-date.

These tests are pretty similar to the [helper function tests](https://github.com/getblocklab/block-lab/blob/693c950aac4c7946be6efb1dae5d41af26f5d959/js/blocks/helpers/test/getBlockLabAttributes.js). Though they test rendered components, not just functions.

This should give us more confidence when editing component files. 

Thanks, Luke!

kienstra:
Sorry, those are the last commits I'm planning on making before review.

kienstra:
Hi @lukecarbis,
Good talking to you earlier.

Would you be able to review this in the next few days? It does affect some non-test files, but most of the diff is tests.

Thanks, Luke!

rheinardkorf:
Nice work @kienstra.

Also, good idea replacing the `wp.` functions with its package equivalents.

kienstra:
Thanks, @rheinardkorf!

I'm still trying out the component tests, but so far they look reliable, and not too prone to fail on refactoring.

lukecarbis (approved):
Looks good Ryan. Everything works as per your testing instructions, and the tests pass as well. Nice job!

kienstra:
Thanks, @lukecarbis!

PR #517, opened 2020-03-07
Make `block_field()` return `null` if second argument is `true`

kienstra:
#### Changes
`block_field( 'foo' )` now returns `null`

#### Background
Before,  `block_field( 'foo' )` returned the same value that it echoed.

This can lead people to believe that it can be used to get the value.

For example:

```php
$text = block_field( 'example-text' );
if ( $text ) {
        echo '<p>' . esc_html( $text ) . '</p>';
}
```

...leads to output like:
```
Here is the text field<p>Here is the text field</p>
```

There have been some support topics about this:
https://wordpress.org/support/topic/variables-are-echoing-out/
https://wordpress.org/support/topic/php-block_field-name-of-field-does-not-display/

This won't fix those issues, as `block_value()` was usually needed instead of `block_field()`.

But it'll probably make it more clear from the beginning that `block_field( 'foo' )` shouldn't be used to get a value, only to echo. 

#### Testing instructions
1. Create a new post
2. Add a Block Lab block that has several fields
3. Ensure there's a template for the block that uses `block_field()`
4. Ensure the template still displays as expected in the `ServerSideRender` and the front-end

### DIFF
--- a/php/helpers.php
+++ b/php/helpers.php
@@ -89,6 +89,8 @@ function block_field( $name, $echo = true ) {
 		 * and then output the field with a more suitable escaping function.
 		 */
 		echo wp_kses_post( $value );
+
+		return null;
 	}
 
 	return $value;
--- a/tests/php/integration/fixtures/all-fields-except-repeater.php
+++ b/tests/php/integration/fixtures/all-fields-except-repeater.php
@@ -74,7 +74,5 @@ foreach ( $non_string_fields as $name => $value ) :
 	endforeach;
 endforeach;
 
-printf(
-	'Here is the result of block_field() for multiselect: %s',
-	block_field( 'multiselect' )
-);
+echo 'Here is the result of block_field() for multiselect: ';
+block_field( 'multiselect' );
--- a/tests/php/unit/test-helpers.php
+++ b/tests/php/unit/test-helpers.php
@@ -87,7 +87,7 @@ class Test_Helpers extends \WP_UnitTestCase {
 
 		// Because block_field() has a second argument of true, this should echo the user login and return it.
 		$this->assertEquals( $mock_text, $actual_user_login );
-		$this->assertEquals( $return_value, $actual_user_login );
+		$this->assertEquals( null, $return_value );
 
 		ob_start();
 		$return_value = block_field( $class_key, true );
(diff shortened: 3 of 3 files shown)


php/helpers.php
```diff
@@ -89,6 +89,8 @@ function block_field( $name, $echo = true ) {
 		 * and then output the field with a more suitable escaping function.
 		 */
 		echo wp_kses_post( $value );
+
+		return null;
```
kienstra, line 93: This can be a breaking change for something like:

```php
$url = get_field( 'url-field' );
if ( $url ) {
        echo '<a href="' . esc_url( $url ) . '">Click here</a>'
}
```

But in that case, there would have already be unexpected output:

```
https://example.com <a href="https://example.com">Click here</a>
```

Maybe there are cases where this would break something that was working.

kienstra: Maybe something like this would be a problem:

```php
$image = get_field( 'foo-image' );
if ( $image ) {
        echo '<p>Example caption of the image</p>';
}
```

We could note in the Changelog that `block_field( 'foo' )` will not return anything anymore. So if there are conditionals that depend on it, please migrate to `block_value()`.

kienstra:
Hi @lukecarbis,
What do you think about this? 

It's a breaking change for [some cases](https://github.com/getblocklab/block-lab/pull/517#discussion_r389304261), but overall I think it'll reduce confusion.

Thanks, Luke!

php/helpers.php
lukecarbis: I don't think that this is a breaking change worth worrying about.

lukecarbis:
Yeah, I think this is a great call.

php/helpers.php
lukecarbis, line 93: We should probably add `@return mixed` to the docblock.

lukecarbis (approved):

php/helpers.php
kienstra: OK, thanks a lot for reviewing this.

kienstra: OK, add it to the `block_field()` DocBlock? It looks like it's already there, but I could have misunderstood this.

lukecarbis: Ah my bad I was looking at the wrong lines, sorry!

kienstra: All good, thanks for reviewing this

PR #538, opened 2020-04-22
Remove Upgrade screen

lukecarbis:
#### Changes
* Upgrade screen is hidden by default
* If the upgrade screen is filtered on intentionally, it now shows a "Not available" message.

#### Testing instructions
1. Remove your Pro license key (if you've got one)
2. Check that there's no "Go Pro" item in the menu
3. Filter `block_lab_show_pro_nag` to return `true`
4. Make sure there's no "upgrade" messaging on the Go Pro screen

#### Note
I didn't just pull the Upgrade class out entirely, because I didn't want to have to deprecate the `block_lab_show_pro_nag` filter. The class might be useful in some other context.

Screen Shot 2020-04-22 at 1 23 58 pm
(This screen is no longer accessible by default)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,11 @@ Simple to use functions, ready to render and work with the values stored through
 ### Connect with Community ###
 Join The Lab, our free Block Lab Community on Slack. The Lab is our real-time home for sharing resources, learning, and making connections. [Learn more.](https://getblocklab.com/welcome-to-the-lab/)
 
-### Go Pro ###
-Block Lab Pro is here, with powerful features to take your block building to the next level. [Learn more.](https://getblocklab.com/block-lab-pro)
-
 ## Links ##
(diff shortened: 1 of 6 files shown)


kienstra:
Also, it'd probably be good to delete the 'Get Block Lab Pro' link:

link-bl-pro

kienstra (approved):
**Looks Good**

Hi @lukecarbis,
Good idea to do this. The only comment I have [is above](https://github.com/getblocklab/block-lab/pull/538#issuecomment-618033572).

kienstra:
Ah, it'd probably also be good to remove the 'Go Pro' section in `README.md`.

lukecarbis:
@kienstra Great pickups! Thank you! Addressed in the commits above.

kienstra:
Thanks, @lukecarbis! This looks great. Merge whenever you'd like.

kienstra (approved):

PR #588, opened 2020-05-27
Rename the custom post type and block namespace

kienstra:
#### Changes
* Changes the slug of the `block_lab` CPT
* Changes the block namespace in the CPT `post_content`. It's now `block-lab/`

#### Testing instructions
* Run PHPUnit tests
* There's no way to functionally test this yet. It isn't connected to anything now, as the UI is not yet decided

### DIFF
--- a/php/blocks/migration/class-post-type.php
+++ b/php/blocks/migration/class-post-type.php
@@ -123,6 +123,9 @@ class Post_Type {
 	/**
 	 * Gets the posts of the previous post_type.
 	 *
+	 * This doesn't have an 'offset' parameter, as the migration changes the post_type.
+	 * So this query won't find posts that were already migrated.
+	 *
 	 * @return WP_Post[] The posts that were found.
 	 */
 	private function query_for_posts() {
--- a/php/blocks/migration/class-post-type.php
+++ b/php/blocks/migration/class-post-type.php
@@ -15,7 +15,7 @@ use WP_Query;
 /**
  * Class Post_Type
  */
-final class Post_Type {
+class Post_Type {
 
 	/**
 	 * The previous slug of the custom post type (in Block Lab).
--- a/php/blocks/migration/class-post-type.php
+++ b/php/blocks/migration/class-post-type.php
@@ -66,7 +66,10 @@ final class Post_Type {
 	}
 
 	/**
-	 * Migrates all of the custom posts post to the new slug and block namespace.
+	 * Migrates all of the custom post type posts to the new slug and block namespace.
+	 *
+	 * These each store a config for a custom block,
+	 * they aren't actual posts as entered into the block editor.
 	 */
 	public function migrate_all() {
 		$offset = 0;
--- a/php/blocks/migration/class-post-type.php
+++ b/php/blocks/migration/class-post-type.php
@@ -1,0 +1,143 @@
+<?php
+/**
+ * Post_Type.
+ *
+ * @package   Block_Lab
+ * @copyright Copyright(c) 2020, Block Lab
+ * @license http://opensource.org/licenses/GPL-2.0 GNU General Public License, version 2 (GPL-2.0)
+ */
+
+namespace Block_Lab\Blocks\Migration;
+
+use WP_Post;
+use WP_Query;
+
+/**
+ * Class Post_Type
+ */
+final class Post_Type {
+
+	/**
+	 * The previous slug of the custom post type (in Block Lab).
+	 *
+	 * @var string
+	 */
+	private $previous_post_type_slug;
+
+	/**
+	 * The new slug of the custom post type (not in Block Lab).
+	 *
+	 * @var string
+	 */
+	private $new_post_type_slug;
+
+	/**
+	 * The previous namespace of the block.
+	 *
+	 * @var string
+	 */
+	private $previous_block_namespace;
+
+	/**
+	 * The new namespace of the block.
+	 *
+	 * @var string
+	 */
+	private $new_block_namespace;
+
+	/**
+	 * The batch size (posts_per_page) for the post queries.
+	 *
+	 * @var int
+	 */
+	private $query_batch_size = 10;
+
+	/**
+	 * Post_Type constructor.
+	 *
+	 * @param string $new_post_type_slug  The new slug of the custom post type.
+	 * @param string $new_block_namespace The new namespace of the block.
+	 */
+	public function __construct( $new_post_type_slug, $new_block_namespace ) {
+		$this->previous_post_type_slug  = block_lab()->get_post_type_slug();
+		$this->new_post_type_slug       = $new_post_type_slug;
+		$this->previous_block_namespace = 'block-lab';
+		$this->new_block_namespace      = $new_block_namespace;
+	}
+
+	/**
+	 * Migrates all of the custom posts post to the new slug and block namespace.
+	 */
+	public function migrate_all() {
+		$offset = 0;
+		$posts  = $this->query_for_posts( $offset );
+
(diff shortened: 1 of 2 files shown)


kienstra:
# Request for review

Hi @lukecarbis,

This is just migrating the `block_lab` CPT. There's no way to actually use this yet, as there isn't a UI yet.

So this shouldn't be merged yet, maybe it'll remain a feature branch for the migration work.

This will probably be changed slightly as the UI develops. Maybe there'll be error messages, or some indication of completion.

php/blocks/migration/class-post-type.php
```diff
@@ -0,0 +1,146 @@
...
+
+namespace Block_Lab\Blocks\Migration;
+
+use WP_Post;
+use WP_Query;
+
+/**
+ * Class Post_Type
+ */
+final class Post_Type {
```
lukecarbis: We haven't used `final` anywhere else. in Block Lab. Happy to include it – I can't imagine this class ever needing a child, but just wanted to call that out for consistency.

php/blocks/migration/class-post-type.php
```diff
@@ -0,0 +1,146 @@
...
+	 * These each store a config for a custom block,
+	 * they aren't actual posts as entered into the block editor.
+	 */
+	public function migrate_all() {
+		$offset = 0;
+		$posts  = $this->query_for_posts( $offset );
+
+		while ( ! empty( $posts ) ) {
+			foreach ( $posts as $post ) {
+				$this->migrate_single( $post );
```
lukecarbis, line 72: Should we break and return a `WP_Error` or something like that if this returns `false`?

lukecarbis (approved):
Looks good! Couple of small questions, but happy for this to be merged as is.

php/blocks/migration/class-post-type.php
kienstra: Thanks, good point. Yeah, it's probably not needed. We were using it with the AMP plugin so classes couldn't be extended, but I don't think people are going to be extending this.

php/blocks/migration/class-post-type.php
kienstra: Good question, that might depend on the error feedback in the UI.

If there are 25 `block_lab` posts, and only the 10th has an error in migration, should we keep migrating the other 15, or stop the whole migration and display an error in the UI?

I think this `block_lab` post migration is less prone to error than the normal `post` or `page` migration, but we won't know until it's used.

lukecarbis: That's a really good question. Let's leave this as it is now, so that it can continue if it hits an error, but keep that in mind as we develop the UI.

kienstra: Thanks, sounds good.

kienstra: If the migration of all of the `block_lab` posts failed, we'd definitely want some kind of feedback.

kienstra:
Hi @lukecarbis, 
Thanks for reviewing this.

If it's alright, I pushed [another commit](https://github.com/getblocklab/block-lab/pull/588/commits/f63f28b1f930122421bafd4e814505389623e969) that fixes an issue and adds more tests.

Also, the target branch is now `add/full-migration`, which is branched off `develop`.

Maybe that could be the feature branch for all of the migration work. It seems a little strange merging this into `develop` now, as there's no way to use the code yet.

php/blocks/migration/class-post-type.php
```diff
@@ -143,4 +119,20 @@ public function migrate_single( WP_Post $post ) {
...
+	/**
+	 * Gets the posts of the previous post_type.
+	 *
+	 * @return WP_Post[] The posts that were found.
+	 */
+	private function query_for_posts() {
+		$query = new WP_Query(
+			[
+				'post_type'      => $this->previous_post_type_slug,
+				'posts_per_page' => 10,
```
lukecarbis, line 135: I don't understand how this can work without an offset parameter?

php/blocks/migration/class-post-type.php
```diff
@@ -132,7 +108,7 @@ public function migrate_single( WP_Post $post ) {
 		$rows_updated = $wpdb->update(
 			$wpdb->posts,
 			[
-				'post_type'    => $this->new_post_type_slug,
+				'post_type'    => sanitize_key( $this->new_post_type_slug ),
```
lukecarbis, line 111: Good call.

tests/php/unit/blocks/migration/test-class-post-type.php
```diff
@@ -147,6 +148,97 @@ public function test_migrate_all() {
...
+
+		$queried_new_cpt_posts = new WP_Query(
+			[
+				'post_type'      => self::NEW_POST_TYPE_SLUG,
+				'posts_per_page' => -1,
+			]
+		);
+
+		// The block_lab CPT posts should now have the new post_type and migrated post_content.
+		$this->assertEquals( $number_of_block_lab_posts, $queried_new_cpt_posts->post_count );
```
lukecarbis, line 192: How does this test ensure that the same set of posts weren't just looped through during migration?

e.g.
initial state: `a`,`b`,`c`,`d`,`e`,`f`,`g`,`h`,`i`,`j`,`k`,`l`,`m`,`n`,`o`,`p`
end state: `a`,`b`,`c`,`a`,`b`,`c`,`a`,`b`,`c`,`a`,`b`,`c`,`a`,`b`,`c`,`a`

(both sets have a count of 16)

lukecarbis:
@kienstra Great idea merging into `add/full-migration`. 👍

php/blocks/migration/class-post-type.php
kienstra: Yeah, this is confusing.

When the posts are migrated, they have their `post_type` changed. So this `WP_Query` won't find posts that were migrated, as it's looking for another `post_type`.

When this did have an `offset`, the offset eventually got high enough that there were no posts.

php/blocks/migration/class-post-type.php
kienstra: Thanks

php/blocks/migration/class-post-type.php
lukecarbis: Ahhh – that makes total sense. Maybe leave a comment in there to that effect?

tests/php/unit/blocks/migration/test-class-post-type.php
lukecarbis: Feel free to ignore this if you think it's safe enough as is.

kienstra: Hm, this tests that the right amount of posts had their `post_type` changed, as they'll have the `self::NEW_POST_TYPE_SLUG`:

https://github.com/getblocklab/block-lab/blob/f63f28b1f930122421bafd4e814505389623e969/tests/php/unit/blocks/migration/test-class-post-type.php#L184-L189

Still, that `assertEquals()`  doesn't ensure that the `post_content` was changed.

php/blocks/migration/class-post-type.php
kienstra: Sure, good idea. I missed that in the first iteration of this 😄

kienstra: https://github.com/getblocklab/block-lab/pull/588/commits/3a3afefc054316cb48c20977b6b1c2f99cae1ba9 adds a comment about that

kienstra:
Also, we should be able to do automated e2e tests of the migration, as long as the new plugin is in a public repo that Travis can access.

It should be able to verify that the migration of a post worked and the new plugin's block looks right.

PR #606, opened 2020-06-09
Migrate `block-lab` namespace in `post_content`

kienstra:
#### Changes
* Adds PHP functions to migrate the entire site's `post_content` from the `block-lab/` block namespace to `genesis-custom-blocks/`
* There are more considerations after this PR, like a REST API or AJAX endpoint that the UI can use

#### Testing instructions
1. `wp db export # if you'd like, so you can restore to this backup of your DB `
2. `wp eval-file tests/bin/migrate-post-content.php`

wp-cli-file

3. Expected: The WP-CLI command results in 'Success', like above
4. Go to the 'Revision' links shown above
5. Expected: The revision diff only involves `block-lab`. Nothing else should have changed:
exampled-diff-here

6. Expected: If there's no `block-lab/` namespace in the content, the post should not be shown in the WP-CLI command output above, and there should of course be no revision.
7. It'd be great if you could go to 40-50 of those revision URLs to sanity-check them, assuming there are that many. It's tedious, but they're all just one revision ID higher than the next. 

8. To reverse this migration:
```sh
wp shell
( new Block_Lab\Blocks\Migration\Post_Content( 'genesis-custom-blocks', 'block-lab' ) )->migrate_all();
```

### DIFF
--- a/php/blocks/migration/class-post-content.php
+++ b/php/blocks/migration/class-post-content.php
@@ -86,7 +86,10 @@ class Post_Content {
 	public function migrate_single( $post_id ) {
 		$post = get_post( $post_id );
 		if ( ! isset( $post->ID ) ) {
-			return new WP_Error( 'Invalid post ID' );
+			return new WP_Error(
+				'invalid_post_id',
+				__( 'Invalid post ID', 'block-lab' )
+			);
 		}
 
 		$replacement_count = 0;
@@ -99,7 +102,10 @@ class Post_Content {
 		);
 
 		if ( 0 === $replacement_count ) {
-			return new WP_Error( 'Post content did not have blocks with the namespace' );
+			return new WP_Error(
+				'no_block_namespace_found',
+				__( 'Post content did not have blocks with the namespace', 'block-lab' )
+			);
 		}
 
 		return wp_update_post(
--- a/php/blocks/migration/class-post-content.php
+++ b/php/blocks/migration/class-post-content.php
@@ -15,6 +15,7 @@ use WP_Error;
  * Class Post_Content
  */
 class Post_Content {
+
 	/**
 	 * The previous namespace of the block.
 	 *
@@ -32,8 +33,8 @@ class Post_Content {
 	/**
 	 * Post_Content constructor.
 	 *
-	 * @param string $previous_block_namespace  Previous namespace of the blocks.
-	 * @param string $new_block_namespace       New namespace of the blocks.
+	 * @param string $previous_block_namespace Previous namespace of the blocks.
+	 * @param string $new_block_namespace      New namespace of the blocks.
 	 */
 	public function __construct( $previous_block_namespace, $new_block_namespace ) {
 		$this->previous_block_namespace = $previous_block_namespace;
--- a/php/blocks/migration/class-post-content.php
+++ b/php/blocks/migration/class-post-content.php
@@ -77,6 +77,7 @@ class Post_Content {
 		if ( ! isset( $post->ID ) ) {
 			return new WP_Error( 'Invalid post ID' );
 		}
+
 		$new_post_content = preg_replace(
 			'#(<!--\s+wp:)(' . sanitize_key( $this->previous_block_namespace ) . ')(/[a-z][a-z0-9_-]*)#s',
 			'$1' . sanitize_key( $this->new_block_namespace ) . '$3',
--- a/php/blocks/migration/class-post-content.php
+++ b/php/blocks/migration/class-post-content.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * Post_Content.
+ *
+ * @package   Block_Lab
+ * @copyright Copyright(c) 2020, Block Lab
+ * @license http://opensource.org/licenses/GPL-2.0 GNU General Public License, version 2 (GPL-2.0)
+ */
+
+namespace Block_Lab\Blocks\Migration;
+
+use WP_Post;
+
+/**
+ * Class Post_Content
+ */
+class Post_Content {
+	/**
+	 * The previous namespace of the block.
+	 *
+	 * @var string
+	 */
+	private $previous_block_namespace;
+
+	/**
+	 * The new namespace of the block.
+	 *
+	 * @var string
+	 */
+	private $new_block_namespace;
+
+	/**
+	 * Post_Content constructor.
+	 *
+	 * @param string $previous_block_namespace  Previous namespace of the blocks.
+	 * @param string $new_block_namespace       New namespace of the blocks.
+	 */
+	public function __construct( $previous_block_namespace, $new_block_namespace ) {
+		$this->previous_block_namespace = $previous_block_namespace;
+		$this->new_block_namespace      = $new_block_namespace;
+	}
+
+	/**
+	 * Migrates the block namespaces in post_content.
+	 *
+	 * Blocks are stored in the post_content of a post with a namespace,
+	 * like '<!-- wp:block-lab/test-image {"example-image":8} /-->'.
+	 * In that case, 'block-lab' needs to be changed to the new namespace.
+	 * But nothing else in the block should be changed.
+	 * The block pattern is mainly taken from Gutenberg.
+	 *
+	 * @see https://github.com/WordPress/wordpress-develop/blob/78d1ab2ed40093a5bd2a75b01ceea37811739f55/src/wp-includes/class-wp-block-parser.php#L413
+	 *
+	 * @param WP_Post $post The post to convert.
+	 * @return int|
+	 * On success: the post ID that was changed, on failure: 0 or WP_Error.
+	 */
+	public function migrate_single( WP_Post $post ) {
+		$new_post_content = preg_replace(
+			'#(<!--\s+wp:)(' . sanitize_key( $this->previous_block_namespace ) . ')(/[a-z][a-z0-9_-]*)#s',
+			'$1' . sanitize_key( $this->new_block_namespace ) . '$3',
+			$post->post_content
+		);
+
+		return wp_update_post(
+			[
+				'ID'           => $post->ID,
+				'post_content' => $new_post_content,
+			]
+		);
+	}
+}
--- a/php/blocks/migration/class-post-content.php
+++ b/php/blocks/migration/class-post-content.php
@@ -53,7 +53,7 @@ class Post_Content {
 	 * @see https://github.com/WordPress/wordpress-develop/blob/78d1ab2ed40093a5bd2a75b01ceea37811739f55/src/wp-includes/class-wp-block-parser.php#L413
 	 *
 	 * @param WP_Post $post The post to convert.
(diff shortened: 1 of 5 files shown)


tests/php/integration/test-repeater-template-output.php
```diff
@@ -182,9 +182,9 @@ public function get_block_config() {
 	 *
 	 * @covers \block_rows()
 	 * @covers \block_row()
-	 * @covers \reset_block_row()
-	 * @covers \block_row_field()
-	 * @covers \block_row_value()
+	 * @covers \reset_block_rows()
+	 * @covers \block_sub_field()
+	 * @covers \block_sub_value()
```
kienstra, line 187: I referenced functions that didn't exist 😄

tests/bin/migrate-post-content.php
```diff
@@ -0,0 +1,116 @@
+<?php
+/**
+ * Migrates the entire site's post content to the new namespace.
+ *
+ * For testing, do not use in production.
```
kienstra, line 5: This WP-CLI script is mainly for our own testing. I don't think we should publicize this. 

It also checks that you really want to migrate:

sure-migr

php/blocks/migration/class-post-content.php
```diff
@@ -0,0 +1,131 @@
...
+	 *
+	 * @return array The posts that were found.
+	 */
+	private function query_for_posts() {
+		global $wpdb;
+
+		$query_limit = 10;
+		return $wpdb->get_results(
+			$wpdb->prepare(
+				"SELECT * FROM {$wpdb->posts} WHERE post_type != %s AND post_content LIKE %s LIMIT %d",
```
kienstra, line 136: Like in #588, this shouldn't get posts that were already migrated. 

Their `post_content` should have changed to have the new namespace, so `post_content LIKE %s` shouldn't be satisfied.

php/blocks/migration/class-post-content.php
```diff
@@ -0,0 +1,131 @@
...
+				'revision',
```
kienstra, line 137: This doesn't migrate `revision` posts. I thought that'd be like overwriting their 'GIT history' of posts.

This migration creates a new revision for a migrated post:

rev-here

This should make it easier for us to debug if something went wrong.

kienstra:
Hi @lukecarbis and @RobStino,
Hope your day's off to a great start.

Could you please review this when you have a chance? The [top comment](https://github.com/getblocklab/block-lab/pull/606#issue-432064433) is up to date.

Thanks!

php/blocks/migration/class-post-content.php
kienstra: This queries for posts that have `wp:block-lab/` in the content. I don't think that's possible with `WP_Query`.

kienstra:
For reviewers in an environment without any Block Lab block:

1. Publish the 'Example Block':

publish-example-block

2. Create a few posts, and add that block to it. 
3. If you're adventurous, create a few more Block Lab blocks like the 'Example Block', and add them to posts.

php/blocks/migration/class-post-content.php
```diff
@@ -0,0 +1,131 @@
...
+	 *
+	 * @see https://github.com/WordPress/wordpress-develop/blob/78d1ab2ed40093a5bd2a75b01ceea37811739f55/src/wp-includes/class-wp-block-parser.php#L413
+	 *
+	 * @param int $post_id The ID of the post to convert.
+	 * @return int|WP_Error The post ID that was changed, or a WP_Error on failure.
+	 */
+	public function migrate_single( $post_id ) {
+		$post = get_post( $post_id );
+		if ( ! isset( $post->ID ) ) {
+			return new WP_Error( 'Invalid post ID' );
```
lukecarbis: `WP_Error` has an error "code" as the first parameter of its constructor. The message comes second. [Here's an example](https://github.com/WordPress/WordPress/blob/a576a13246f85dfa76183911041d666a5cd42818/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php#L175).

Also, should we translate error messages?

My suggestion is something like:
```PHP
return new WP_Error(
	'invalid_post_id',
	__( 'Invalid post ID', 'genesis_custom_blocks' )
);
```

php/blocks/migration/class-post-content.php
```diff
@@ -0,0 +1,131 @@
...
+		$new_post_content  = preg_replace(
+			'#(<!--\s+wp:)(' . sanitize_key( $this->previous_block_namespace ) . ')(/[a-z][a-z0-9_-]*)#s',
+			'$1' . sanitize_key( $this->new_block_namespace ) . '$3',
+			$post->post_content,
+			-1,
+			$replacement_count
+		);
+
+		if ( 0 === $replacement_count ) {
+			return new WP_Error( 'Post content did not have blocks with the namespace' );
```
lukecarbis: See `WP_Error` comment above.

php/blocks/migration/class-post-content.php
lukecarbis: Great thinking.

php/blocks/migration/class-post-content.php
```diff
@@ -0,0 +1,131 @@
...
+			foreach ( $posts as $post ) {
+				if ( isset( $post->ID ) ) {
+					$results[] = $this->migrate_single( $post->ID );
+				}
+			}
+
+			$posts = $this->query_for_posts();
+		}
+
+		return $results;
```
lukecarbis, line 69: I think I found a small problem with this looping `while` approach.

Suppose 1 post doesn't migrate. It returns a `WP_Error` due to an invalid ID, for example, or the `wp_update_post()` function breaks.

Next loop through, that same post that caused an error in the previous loop will be reattempted, since it will be the first result in `query_for_posts`. That's no big problem. But what if it breaks every time? What if the second and third and fourth time we try to update it, it always returns the same error?

In that case, it's still no big deal, it's just slightly less efficient, as we'll only be migrating 9 posts with every loop through instead of 10.

BUT… what if it's not 1 post, but all 10 posts! Or 5 posts in the first loop through and 5 posts in the second (causing a total of 10 errors after the second loop). Since our `$query_limit` is set to 10, we'll end up with an infinite loop.

Having said that, I think the fix for that problem potentially introduces a ton of complexity. Happy to move forward with this PR knowing that this super unlikely edge case exists, but not fixing it.

tests/php/unit/blocks/migration/test-class-post-content.php
```diff
@@ -0,0 +1,280 @@
...
+		$new_post = get_post( $post_id );
+		$this->assertEquals( $expected_post_content, $new_post->post_content );
+	}
+
+	/**
+	 * Test migrate_single with an invalid post ID.
+	 *
+	 * @covers \Block_Lab\Blocks\Migration\Post_Content::migrate_single()
+	 */
+	public function test_migrate_single_invalid_post_id() {
```
lukecarbis, line 220: Great idea.

lukecarbis:
Looks like e2e tests are failing?

lukecarbis (approved):
Nice work on this Ryan.

Migration script worked beautifully.

tests/php/unit/blocks/migration/test-class-post-content.php
kienstra: Thanks!

php/blocks/migration/class-post-content.php
kienstra: Good idea

php/blocks/migration/class-post-content.php
kienstra: Good suggestion, the message should be second.

php/blocks/migration/class-post-content.php
kienstra: That's a good catch, I'll think about that. It could definitely create an infinite loop if there are 10 failures.

kienstra: Really good point, I think that is a case we should handle.

What do you think about https://github.com/getblocklab/block-lab/pull/606/commits/f6aada67ce83125c46285ed04cce24375ed3389c?

It returns from `migrate_all()` if there are more than 20 `WP_Error`s in migration.

It'll prevent an infinite loop. 

If there was a `WP_Error`, something might have gone very wrong. Like `wp_insert_post()` [tried to insert an empty post](https://github.com/WordPress/wordpress-develop/blob/5e2ef3d93739610e97dd20f5f254a355a4eb5bd3/src/wp-includes/post.php#L3697). If there are over 20 of those, it's probably better to exit. 

It might be possible to keep track of failed post IDs and exclude them from the SQL query, maybe with `NOT IN`. But that might not be very efficient.

kienstra:
>Looks like e2e tests are failing?

Yeah, that's a flaky test case that I introduced. I'll have to fix it some time, whether it passes now or not.

php/blocks/migration/class-post-content.php
```diff
@@ -0,0 +1,137 @@
...
+		$results            = [];
+		$error_count        = 0;
+		$max_allowed_errors = 20;
+		$posts              = $this->query_for_posts();
+
+		while ( $posts && $error_count < $max_allowed_errors ) {
+			foreach ( $posts as $post ) {
+				if ( isset( $post->ID ) ) {
+					$migrated_post = $this->migrate_single( $post->ID );
+					$results[]     = $migrated_post;
```
kienstra, line 59: Before releasing this, we might remove `$results`, and only return `WP_Error` messages if there were any. If there are hundreds of posts migrated, this could be too big.

But for now, it lets us test this and enables the WP-CLI output of exactly what posts were migrated.

php/blocks/migration/class-post-content.php
kienstra: Thanks for suggesting this, https://github.com/getblocklab/block-lab/pull/606/commits/d708186bac9981ec3fc67f0f842ee4a6c0866d31 uses your suggestion.

php/blocks/migration/class-post-content.php
kienstra: Thanks, fixed in https://github.com/getblocklab/block-lab/pull/606/commits/d708186bac9981ec3fc67f0f842ee4a6c0866d31

php/blocks/migration/class-post-content.php
lukecarbis: Yeah, this is all pretty edge case. I'm happy with the 20 errors limit. 🚢

kienstra: Nice, thanks!

kienstra:
This PR can still be tested with the same [testing instructions](https://github.com/getblocklab/block-lab/pull/606#issue-432064433), just checkout the [add/full-migration](https://github.com/getblocklab/block-lab/tree/add/full-migration) branch.

Here are some [instructions](https://github.com/getblocklab/block-lab/pull/606#issuecomment-646743052) to populate Block Lab blocks if needed.

It'd be interesting to know roughly how long `wp eval-file tests/bin/migrate-post-content.php` takes in your local. 2-3 seconds, or 20 seconds?

PR #629, opened 2020-06-24
Add the migration page

kienstra:
#### Changes
* The markup and CSS is from Rob's Gist 😄  
* Migrates Block Lab blocks to Genesis Custom Blocks
* Adds a 'Migrate' submenu under `/wp-admin > 'Block Lab'`

migration-full

#### Testing instructions (updated July 27th)
1. Upload and activate the plugin [block-lab.zip](https://github.com/getblocklab/block-lab/files/4983292/block-lab.zip), or do `composer install && npm install && npm run build`
2. If needed, [populate some Block Lab blocks](https://github.com/getblocklab/block-lab/pull/606#issuecomment-646743052)
3. Either install the GCB plugin: [genesis-custom-blocks.zip](https://github.com/getblocklab/block-lab/files/4983329/genesis-custom-blocks.zip), or do steps 5-7
4. Don't activate it yet
5. `cd path/to/wp-content/plugins/`
6. `git clone https://github.com/studiopress/genesis-custom-blocks` # if you haven't done this already
7. `cd genesis-custom-blocks && composer install && npm install && npm run build` # So that it's available to be activated in the final step
8. Go to `/wp-admin` > Block Lab > Migrate
9. Go through the migration flow 
10. Expected: The Block Lab blocks should now appear in Genesis Custom Blocks:
bl-block-here

11. Expected: Block Lab blocks in posts should look the same other than icons, and should now be Genesis Custom Blocks blocks:

icon-genesis-here

12. Here's how to [reverse the migration](https://github.com/getblocklab/block-lab/pull/629#issuecomment-655846402)


kienstra:
There's a lot of work left here, but there's now a 'Migrate' submenu page:

migration-ui-here

kienstra:
If it's alright, I think React will be useful here, and I started the React setup for the Migrate page.

It'd be possible to do this in Vanilla JS, like how the CPT page is now.

But I'd be slower developing for that, mixing JavaScript and PHP. And with React, it'll be easier to test.

kienstra:
There's still a lot of work left, but here's what the migration looks like right now. 

Only 2 of the 5 steps are in place so far.

migration-for-now

lukecarbis:
Looking _amazing_, Ryan! 🤩

kienstra:
Thanks, Luke! Thanks for your great work on the UI.

kienstra:
There's still a lot of work left, but here's what it looks like now:

basic-bl-migration

lukecarbis:
Looking good Ryan! I'm sure @RobStino will drop in with a few design tweaks at some stage, too.

Interesting point on onboarding. Do you think we should somehow try to disable it for migrations? IMO, it's not a huge problem, so if it's going to cause a bunch of extra work, I wouldn't worry about it.

kienstra:
Ah, good point about the onboarding. I'll look at using a query var to disable the onboarding after the migration.

lukecarbis:
Only if it's easy.

kienstra:
OK, thanks.

kienstra:
## Does work
* The migration workflow, including migrating the `block_lab` CPT and the blocks in the `post_content`

migration-with-steps


## Doesn't work yet

* Automatically downloading GCB:

migrates-block

kienstra:
To test this, please also clone [GCB](https://github.com/studiopress/genesis-custom-blocks) and do `composer install && npm install && npm run build`. This is so that it can be activated in the last step of the UI.

You also might want to do `wp db export` so you can restore to this backup of your DB.

## Reversing the migration

To reverse the migration, ensure Block Lab is active and do:
```sh
wp shell
( new Block_Lab\Admin\Migration\Post_Content( 'genesis-custom-blocks', 'block-lab' ) )->migrate_all();
( new Block_Lab\Admin\Migration\Post_Type( 'genesis_custom_block', 'genesis-custom-blocks', 'genesis_custom_blocks', 'block_lab', 'block-lab', 'block_lab' ) )->migrate_all();

```

kienstra:
What do you think about the messages on submitting the Genesis Pro subscription key?

empty-here

There's no validation of the key, it just checks that it's a non-empty string.

As you know, the validation logic is in Genesis Custom Blocks.

There's also a message if `update_option()` returns `false`:

>There was an error saving the key

johnstonphilip:
Hey @kienstra - I took this for a spin and this works really well and looks great!

I hit a couple snags, though may not be related. 

1. I attempted to have both plugins active at the same time, which is throwing this:

```
Plugin could not be activated because it triggered a fatal error.


Fatal error: Cannot redeclare block_field() (previously declared in /Users/phil.johnston/Documents/wps/wpengine1/app/public/wp-content/plugins/block-lab/php/helpers.php:20) in /Users/phil.johnston/Documents/wps/wpengine1/app/public/wp-content/plugins/genesis-custom-blocks/php/Helpers.php on line 20

```

2. The "Developer Notice button didn't seem to work (nothing happens upon click)

kienstra:
Hi @johnstonphilip,
Thanks a lot for testing this, I really appreciate it.

1. Good point, I still need to fix this issue with both plugins being active.

2. Good catch also, https://github.com/getblocklab/block-lab/pull/629/commits/8f74facec192cb782ac82aac568b1f3141baa3a0#diff-1f12e601c690bb1256c9c71f5989a088R39 adds a link for the developer notice.

Thanks, and have a great weekend!

js/migration/components/steps/get-genesis-pro.js
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * @param {GetGenesisProProps} Props The component props.
+ * @return {React.ReactElement} The component to get Genesis Pro.
+ */
+const GetGenesisPro = ( { goToNext, isStepActive, isStepComplete, stepIndex } ) => {
+	// @todo: replace this.
+	const urlMigrateWithoutGenPro = 'https://example.com';
+
+	// @todo: Is this a stable URL? Or should it be changed to https://www.studiopress.com/genesis-pro/
+	const urlGetGenesisPro = 'https://my.wpengine.com/signup?plan=genesis-pro';
```
kienstra, line 40: Is this a stable URL? Or should it be changed to https://www.studiopress.com/genesis-pro/ ?

js/migration/components/steps/get-genesis-pro.js
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * The step to get Genesis Pro.
+ *
+ * @param {GetGenesisProProps} Props The component props.
+ * @return {React.ReactElement} The component to get Genesis Pro.
+ */
+const GetGenesisPro = ( { goToNext, isStepActive, isStepComplete, stepIndex } ) => {
+	// @todo: replace this.
+	const urlMigrateWithoutGenPro = 'https://example.com';
+	const urlGetGenesisPro = 'https://my.wpengine.com/signup?plan=genesis-pro';
```
kienstra, line 38: Is this a stable URL? Or should it be changed to https://www.studiopress.com/genesis-pro/ ?

kienstra:
### Request for review

Hi @lukecarbis and @RobStino,
Could you please review this? The [top comment](https://github.com/getblocklab/block-lab/pull/629#issue-439492116) is up to date.

This migration doesn't actually install Genesis Custom Blocks. That's in a ticket that's scheduled for next sprint, though I'll look at doing that soon.

Here are example error messages from post content migration:

migration-errors-here

Though ideally there won't be errors 😄 

There are also `@todo` comments here to add URLs, like the [Developer Notice](https://github.com/getblocklab/block-lab/pull/629/files#diff-1f12e601c690bb1256c9c71f5989a088R20). Once we know those, I'll add them in.

It'll be interesting how long the post content migration takes for you. It usually takes a few seconds for me.

Thanks, Rob and Luke!

js/migration/components/steps/get-genesis-pro.js
lukecarbis: In this case, sending people directly to the signup page (not another marketing page) is preferred, so leave as is.

js/migration/components/steps/get-genesis-pro.js
```diff
@@ -0,0 +1,234 @@
...
+		event.preventDefault();
+		if ( ! subscriptionKey ) {
+			setSubmissionMessage( __( 'The subscription key is empty.', 'block-lab' ) );
+			setKeySubmittedSuccessfully( false );
+			return;
+		}
+
+		// @ts-ignore global
+		if ( subscriptionKey === blockLabMigration.couponCode ) {
+			setSubmissionMessage( __( "That looks like your coupon code. Please click 'Get Genesis Pro' and enter your coupon code there.", 'block-lab' ) );
```
lukecarbis, line 68: Very, very, very nice. Brilliant work.

lukecarbis (changes requested):
All the code looks fantastic, of course.

I'm not seeing any migration menu in Block Lab though.
https://ufile.io/1n084xcy

js/migration/components/steps/get-genesis-pro.js
kienstra: Thanks, Luke!

kienstra:
Hi @lukecarbis,
Sorry the menu's not displaying. That makes it hard to test 😆 

Does this help?

```diff
diff --git a/php/admin/migration/class-submenu.php b/php/admin/migration/class-submenu.php
index d953c8e..4d1b1a7 100644
--- a/php/admin/migration/class-submenu.php
+++ b/php/admin/migration/class-submenu.php
@@ -135,7 +135,7 @@ class Submenu extends Component_Abstract {
         * @return bool Whether the user can view the migration page.
         */
        public function user_can_view_migration_page() {
-               return current_user_can( 'install_plugins' ) && current_user_can( self::MIGRATION_CAPABILITY );
+               return true;
        }
```

...just to narrow down the problem.

lukecarbis:
Ah, damn. Sorry man. Had the wrong branch checked out.

kienstra:
All good, there's a lot to juggle here with 2 plugins and all of the setup instructions.

lukecarbis (changes requested):
Screen Shot 2020-07-13 at 2 22 29 pm
The last paragraph should have no bottom margin.

Screen Shot 2020-07-13 at 2 34 44 pm
Made a bunch of changes to this section. @RobStino could you please review?

Screen Shot 2020-07-13 at 2 39 38 pm
Is this button supposed to work yet?

Screen Shot 2020-07-13 at 2 40 59 pm
Not yet responsive.

ezgif com-video-to-gif
Animations between steps are very quick / disorienting.

(happened to fast for a screenshot)
The loading bar during migration seemed blue instead of purple.

lukecarbis:
Maybe instead of a loading bar we should show a spinner, since it's all happening awfully fast. Maybe it's only happening fast because we're doing it locally?

kienstra:
>Is this button supposed to work yet?

Yeah, the 'Submit' button for the subscription key should work. Is there a console error?

kienstra:
>Maybe instead of a loading bar we should show a spinner, since it's all happening awfully fast. Maybe it's only happening fast because we're doing it locally?

Yeah, a spinner might be a good idea. Some steps will probably be really fast, especially the CPT migration. In my local, it almost was too fast to see.

Yeah, maybe the speed is from being in a local environment, but some calls will probably be fast normally.

lukecarbis:
I just tried saving the key again and it worked. Must have broken when I messed with the DOM for that mockup. My bad, sorry.

kienstra:
No problem, it could have been a bug, who knows 😄

RobStino (commented):
Hey @kienstra 
This is looking really good.
I tested both without and with a Block Lab Pro license key active and got expected behaviour.
The only thing that I spotted that might be weird is that I got this error in the block editor _after_ the migration.
image
Now I know that the GCB plugin wasn't installed as that part is yet to be sorted, but as the block had been migrated, I was unsure about seeing the `block-lab` slug in that error notice. Not sure how it works technically, or even if this is an issue, but wanted to call it out just in case.

There are a couple of things that need adjusting:
1. We need a better disabled-state for the buttons. They still look active.
2. The animation between steps is janky. I think that is my hacky CSS (height transitions in CSS are notoriously hard!). Is there something you could do with JS to handle transitions?
3. We should capture a task somewhere to update all the external links before launch.

RobStino:
For the disabled state button (#3 above), I mean the next-step buttons and their state prior to the checkbox being checked.
Here is some CSS that I think styles it appropriately:
```
color: var(--color-gray-500);
background-color: var(--color-gray-200);
box-shadow: none;
cursor: not-allowed;
```
For the "allowed state", we'll probably have to add:
```
cursor: pointer;

**Comment by kienstra** [comment]
Hi @RobStino,
Thanks for testing this and your comments.

>The only thing that I spotted that might be weird is that I got this error in the block editor after the migration.

>...I was unsure about seeing the block-lab slug in that error notice. 

Right, that shouldn't happen. The migration should have changed that from `block-lab/notice` to `genesis-custom-blocks/notice` in the `post_content`.

1. What post type was that block in? A post, page, etc..?

2. Whats's the version of WordPress, and is the Gutenberg plugin active?

3. Did other blocks migrate OK?

**Comment by nickcernis** [comment]
Hey, @kienstra. The migration works smoothly for me and looks great. Nice work!

A couple of things I dug up:

### Restoring from browser backups of an unsaved post breaks blocks

The only way I found of reproducing the “your site doesn't include support for the [blockname] block“ from [Rob's comment](https://github.com/getblocklab/block-lab/pull/629#pullrequestreview-447748164) was to restore an old browser save or autosave after doing the migration:

1. Add a custom block to a page and publish the page.
2. Add another custom block to the page, do not publish the page, and wait ~10 seconds for the browser to back up your changes to local storage.
2. Refresh the page. You should see “The backup of this post in your browser is different from the version below.” and a button to “Restore the backup”.
3. In a new tab, complete the block migration.
4. Return to the page (refresh again if you wish) and click “Restore the backup”.

The migrated block and the unsaved block will be replaced with the old cached blocks, because the backup was made with the old Block Lab block names.

This might not be something you need/want to account for, but just flagging it as a possibility in case it comes up in support.

I would expect to see a similar issue with any post autosaves (“There is an autosave of this post that is more recent than the version below. View the autosave.”) but that requires a manual compare-and-restore step where it's more obvious that the block names have changed because you get to see the diff first.

### Keyboard and screen reader accessibility

I did a basic test for keyboard and screen reader accessibility. These may be separate issues to follow up on:

1. Navigating the page at `wp-admin/edit.php?post_type=block_lab&page=block-lab-migration` by keyboard (in Chrome, click to focus the page, then repeatedly press tab) reveals hidden next steps before I have begun the first step. I would not expect to be able to access later steps until I have completed the first one. It could prove confusing for a screen reader user or someone whose only means of access is by keyboard.

    tab-through-steps

2. Step 2: the “ More details here” links could be more descriptive so that screen reader users know what they're clicking on. [Learn more about accessible link text](https://www.washington.edu/accessibility/links/).

3. There's no audible progress alert for screen reader users. When I click “migrate now” in step 3 I would not be able to tell that the step has ever completed if I can't see the screen. In Genesis, John solved this with `wp.a11y.speak()` calls to announce what's going on:
    - [wp-a11y-speak](https://make.wordpress.org/accessibility/handbook/markup/wp-a11y-speak/)
    - [Genesis a11y.speak source](https://github.com/studiopress/genesis/blob/75377bc8f28004454932c5cdf3733862a74029ec/lib/js/admin.js#L359)

4. The “…and step into the FUTURE ꜰᴜᴛᴜʀᴇ FUTURE.” text in step 4 is cute, but VoiceOver read this as, “and step into the future latin letter small capital f, latin letter small capital u, latin letter small capital t, latin letter small capital u, latin letter small capital r, latin letter small capital e future”, which might annoy screen reader users.

   I would either wrap the text in spans and use CSS to adjust their size (which VoiceOver will read as “step into the future future future” which still gives the echo effect) or just change it to “…and step into the future”, even if it loses some of the personality you were going for here.

    It's not so bad in this case but it's something to be aware of as [special character usage can get much worse for screen readers](https://twitter.com/kentcdodds/status/1083073242330361856)! 

5. The final “activate plugin” step redirects to the plugin page, where it's not clear what has happened as a screen reader user. If you can activate the plugin (and announce “Genesis Custom Blocks plugin activated” or similar) but stay on the migration page that may be less disorientating.

**Comment by kienstra** [comment]
Hi @nickcernis,
Thanks so much for your detailed testing. Great idea to do accessibility testing, I really appreciate it.

I'll work on applying your points.

**Comment by kienstra** [comment]
Hi @lukecarbis and @RobStino,
How does this look for the 'Get Genesis Pro' step?

It applies Luke's suggestions from https://github.com/getblocklab/block-lab/pull/629#pullrequestreview-446978121.

to-g-pro

**Comment by kienstra** [comment]
Also, what do you think about this use of a spinner instead of the `<progress>` bar?

The progress bar was good, but the post content migration ended up taking most of the time. The CPT migration was too fast to see in my local.

spinner-here

**Comment by kienstra** [comment]
Changing steps is now a little smoother:

chaning-steps

This [transitions the max-height](https://css-tricks.com/using-css-transitions-auto-dimensions/#technique-1-max-height).

It's hard to transition from a defined height to `height: auto`.

An alternative would be a [more complex JS solution](https://css-tricks.com/using-css-transitions-auto-dimensions/#technique-3-javascript).

**Comment by kienstra** [comment]
This step is a bit more responsive now:

them-tiles

**Comment by kienstra** [comment]
>3. We should capture a task somewhere to update all the external links before launch.

Sure, I opened BH-452 for that.

**Comment by kienstra** [comment]
This is ready for another review.

**Comment by RobStino** [comment]
Hey @kienstra ,
Thanks for all those edits. This looks good to me.
I feel like I might want to tweak a few things around padding and transition durations. I'll work on a separate PR for these though. That would be the best approach hey?

**Comment by nickcernis** [comment]
Accessibility is much better now. Nice job!

**Comment by kienstra** [comment]
Hi @RobStino,
Thanks a lot for looking at this.

>I feel like I might want to tweak a few things around padding and transition durations. I'll work on a separate PR for these though. That would be the best approach hey?

Sure, that sounds good. A PR to the `add/full-migration` branch would be great.

**Comment by kienstra** [comment]
Hi @nickcernis,
Thanks a lot for your detailed review. Sorry for not responding to you about [your suggestions](https://github.com/getblocklab/block-lab/pull/629#issuecomment-658062873).

Great point about restoring from browser backups. Thanks for your steps to reproduce, I also saw that.

Of your 5 suggestions on accessibility, I applied all of them except 5:

>The final “activate plugin” step redirects to the plugin page, where it's not clear what has happened as a screen reader user. If you can activate the plugin (and announce “Genesis Custom Blocks plugin activated” or similar) but stay on the migration page that may be less disorientating.

It's a really good suggestion. But due to a quirk in how Genesis Custom Blocks and Block Lab have some of the same functions, they can't be active at the same time. So as part of activating GBC, it has to deactivate Block Lab.

Thanks for your great suggestions.

**Comment by mindctrl** [comment]
"Backup your site" should probably be "Back up your site"... backup is a noun.

**Comment by kienstra** [comment]
Ah, good point!

**Comment by kienstra** [comment]
'backup' is now changed to 'back up':

back-up-site

**Review by lukecarbis — APPROVED** [review]
@kienstra 

Spinner looks great.

Can we change "Save your Genesis Pro Key / Submit" to "Paste your Genesis Pro Key / Save"

Like @RobStino suggested, I have a few animation / design tweaks in mind, but happy to move them to a separate PR.

**Comment by lukecarbis** [comment]
Feel free to merge with or without the suggested copy change.

**Comment by kienstra** [comment]
Hi @lukecarbis,
Thanks, good idea!

It's now changed to 'Paste your...' and 'Save':

paste-save-here

File: js/migration/components/intro.js [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+
+	return (
+		<>
+			<div>
+				<h1>{ __( 'Migrating to Genesis Custom Blocks', 'block-lab' ) }</h1>
+				<p>{ __( 'In April, the Block Lab team joined the Genesis team at WP Engine. With our full-time focus, we’re very excited about the future of custom block tooling in WordPress. You can read more about that moment in this', 'block-lab' ) } <a target="_blank" rel="noopener noreferrer" className="text-purple-600 underline hover:text-purple-700" href={ announcementUrl }>{ __( 'announcement post', 'block-lab' ) }.</a></p>
+				<p>
+					{ __( 'As part of this move, we have been working on a new plugin that is based on what we developed at Block Lab.', 'block-lab' ) }
+					&nbsp;
+					{ __( 'Genesis Custom Blocks is now the home of all our custom block efforts and what we have planned is very very cool!', 'block-lab' ) }
```
sleepykate, line 31: Optional polish: Change "very very" to "very, very"

sleepykate:
Optional polish:

- Change "ok" to "okay" throughout
- Change "&" to "and" throughout
- Do a quick pass on headers and CTAs for consistent capitalization (sentence case vs. title case)

kienstra:
Thanks, @sleepykate! Good ideas.

kienstra:
@RobStino and @lukecarbis,
What do you think about [those suggestions](https://github.com/getblocklab/block-lab/pull/629#issuecomment-665271278)? 

I could apply them today.

Thanks!

lukecarbis:
@sleepykate Thanks for these suggestions!

Personally I always prefer `okay` to `ok`, too. However, WordPress tends to go with OK (capitalised) ([examples here](https://translate.wordpress.org/projects/wp/dev/en-au/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=988&filters%5Btranslation_id%5D=2986019)).

On the other hand… all those `OK` instances are button actions, but ours are in the middle of sentences. Given this, I'd prefer to go with Katie's suggestion and **change them to `okay`**.

Regarding the ampersand, my opinion is that ampersands are good in headers, but not body text. So keep "Update Hooks & API" but change lines where ampersands are used in that way.

kienstra:
Thanks, @lukecarbis! I'll push those changes to https://github.com/getblocklab/block-lab/pull/660.

PR #635, opened 2020-06-28
Add a migration notice on certain pages

kienstra:
#### Changes
* Adds the migration and 'Not Now' notices
* On clicking 'Not Now', the migration notice never appears again for that user:
* Appears on Block Lab > Content Blocks, Block Lab > Settings, `/wp-admin/plugins.php`, and the Dashboard. 

#### Testing instructions
1. Go to Block Lab > Content Blocks, Block Lab > Settings, `/wp-admin/plugins.php`, and the Dashboard
2. Expected: the notice appears as shown above
3. Click 'Not Now'
4. Expected: the notice is hidden, and the 'Not Now' notice displays
5. Reload the page
6. Expected: The notice doesn't appear again
7. To make the notice appear again:
```sh
wp user list # Find your user ID from that
wp shell
update_user_meta( $user_id, 'block_lab_show_migration_notice', 1 ); // $user_id is the ID from the wp user list command.
```

### DIFF
--- a/js/admin.migration-notice.js
+++ b/js/admin.migration-notice.js
@@ -3,7 +3,8 @@
 document.addEventListener( 'DOMContentLoaded', function() {
 	const hiddenClass = 'bl-hidden';
 
-	// On clicking 'Not now', make an AJAX request to store the user meta to not display the notice again.
+	// In the main migration notice, On clicking 'Not Now',
+	// make an AJAX request to store the user meta to not display the notice again.
 	// Also, remove this notice and display another.
 	document.querySelector( '#bl-notice-not-now' ).addEventListener( 'click', function() {
 		const request = new XMLHttpRequest();
@@ -16,6 +17,8 @@ document.addEventListener( 'DOMContentLoaded', function() {
 
 		const notice = document.querySelector( '#bl-migration-notice' );
 		notice.parentNode.removeChild( notice );
+
+		// Display the 'Not Now' notice.
 		document.querySelector( '#bl-not-now-notice' ).classList.remove( hiddenClass );
 	} );
 
--- a/js/admin.migration-notice.js
+++ b/js/admin.migration-notice.js
@@ -3,7 +3,7 @@
 document.addEventListener( 'DOMContentLoaded', function() {
 	const hiddenClass = 'bl-hidden';
 
-	// In the main migration notice, On clicking 'Not Now',
+	// In the main migration notice, on clicking 'Not Now',
 	// make an AJAX request to store the user meta to not display the notice again.
 	// Also, remove this notice and display another.
 	document.querySelector( '#bl-notice-not-now' ).addEventListener( 'click', function() {
@@ -15,6 +15,7 @@ document.addEventListener( 'DOMContentLoaded', function() {
 		request.open( 'POST', ajaxurl, true );
 		request.send( data );
 
+		// Remove this notice.
 		const notice = document.querySelector( '#bl-migration-notice' );
 		notice.parentNode.removeChild( notice );
 
--- a/js/admin.migration-notice.js
+++ b/js/admin.migration-notice.js
@@ -1,0 +1,26 @@
+/* global XMLHttpRequest, FormData, ajaxurl */
+
+document.addEventListener( 'DOMContentLoaded', function() {
+	const hiddenClass = 'bl-hidden';
+
+	// On clicking 'Not now', make an AJAX request to store the user meta to not display the notice again.
+	// Also, remove this notice and display another.
+	document.querySelector( '#bl-notice-not-now' ).addEventListener( 'click', function() {
+		const request = new XMLHttpRequest();
+		const data = new FormData();
+		data.append( 'action', 'bl_dismiss_migration_notice' );
+		data.append( 'bl-migration-nonce-name', document.querySelector( '#bl-migration-nonce-name' ).value );
+
+		request.open( 'POST', ajaxurl, true );
+		request.send( data );
+
+		const notice = document.querySelector( '#bl-migration-notice' );
+		notice.parentNode.removeChild( notice );
+		document.querySelector( '#bl-not-now-notice' ).classList.remove( hiddenClass );
+	} );
+
+	// In the 'Not Now' notice, on clicking 'OK', hide the notice.
+	document.querySelector( '#bl-notice-ok' ).addEventListener( 'click', function() {
+		document.querySelector( '#bl-not-now-notice' ).classList.add( hiddenClass );
+	} );
+} );
--- a/php/admin/class-migration.php
+++ b/php/admin/class-migration.php
@@ -106,7 +106,10 @@ class Migration extends Component_Abstract {
 					printf(
 						/* translators: %1$s: the plugin name */
 						esc_html__( 'The Block Lab team have moved. For future updates and improvements, migrate now to the new home of custom blocks: %1$s.', 'block-lab' ),
-						'<strong>Genesis Custom Blocks</strong>'
+						sprintf(
+							 '<strong>%1$s</strong>',
+							esc_html__( 'Genesis Custom Blocks', 'block-lab' )
+						)
 					);
 					?>
 				</p>
@@ -197,7 +200,7 @@ class Migration extends Component_Abstract {
 		check_ajax_referer( self::NOTICE_NONCE_ACTION, self::NOTICE_NONCE_NAME );
 
 		if ( ! current_user_can( self::NOTICE_CAPABILITY ) ) {
-			wp_die( -1 );
+			wp_send_json_error();
 		}
 
 		update_user_meta( get_current_user_id(), self::NOTICE_USER_META_KEY, self::NOTICE_DISMISSED_META_VALUE );
--- a/php/admin/class-migration.php
+++ b/php/admin/class-migration.php
@@ -17,33 +17,61 @@ use Block_Lab\Component_Abstract;
 class Migration extends Component_Abstract {
 
 	/**
+	 * The action of the migration notice nonce.
+	 *
+	 * @var string
+	 */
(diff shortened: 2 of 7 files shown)


js/admin.migration-notice.js
```diff
@@ -0,0 +1,26 @@
+/* global XMLHttpRequest, FormData, ajaxurl */
+
+document.addEventListener( 'DOMContentLoaded', function() {
```
kienstra, line 3: Haha, it's strange writing vanilla JS again

php/admin/class-migration.php
```diff
@@ -0,0 +1,205 @@
...
+		?>
+		<div id="bl-migration-notice" class="notice notice-info bl-notice-migration">
+			<?php wp_nonce_field( self::NOTICE_NONCE_ACTION, self::NOTICE_NONCE_NAME, false ); ?>
+			<div class="bl-migration-copy">
+				<p>
+					<?php
+					printf(
+						/* translators: %1$s: the plugin name */
+						esc_html__( 'The Block Lab team have moved. For future updates and improvements, migrate now to the new home of custom blocks: %1$s', 'block-lab' ),
+						'<strong>Genesis Custom Blocks.</strong>'
```
kienstra, line 107: This isn't translated, as it's the name of the plugin. Maybe it should be 😄

php/admin/class-migration.php
```diff
@@ -0,0 +1,205 @@
...
+		// If the user has dismissed the notice, it shouldn't appear again.
+		if ( self::NOTICE_DISMISSED_META_VALUE === get_user_meta( get_current_user_id(), self::NOTICE_USER_META_KEY, true ) ) {
+			return false;
+		}
+
+		$screen = get_current_screen();
+		return (
+			( isset( $screen->base, $screen->post_type ) && 'edit' === $screen->base && 'block_lab' === $screen->post_type )
+			||
+			( isset( $screen->base ) && in_array( $screen->base, [ 'plugins', 'dashboard', 'block_lab_page_block-lab-settings' ], true ) )
```
kienstra, line 186: If you don't want this to display in the Dashboard, it'd be easy to remove `'dashboard'` here

php/admin/class-migration.php
```diff
@@ -0,0 +1,207 @@
...
+		?>
+		<div id="bl-migration-notice" class="notice notice-info bl-notice-migration">
+			<?php wp_nonce_field( self::NOTICE_NONCE_ACTION, self::NOTICE_NONCE_NAME, false ); ?>
+			<div class="bl-migration-copy">
+				<p>
+					<?php
+					printf(
+						/* translators: %1$s: the plugin name */
+						esc_html__( 'The Block Lab team have moved. For future updates and improvements, migrate now to the new home of custom blocks: %1$s.', 'block-lab' ),
+						'<strong>Genesis Custom Blocks</strong>'
```
kienstra: This isn't translated, as it's the name of the plugin. Maybe it should be 😄

kienstra:
#### Request for review

Hi @lukecarbis,
Could you please review this when you can? The [top comment](https://github.com/getblocklab/block-lab/pull/635#issue-441126183) is up to date.

Also, the notice doesn't have borders between the the buttons, like in the Migration UX document. 

Are those borders intended, or are they just from the document formatting?

borders-innotice

Thanks, Luke!

php/admin/class-migration.php
lukecarbis: I think it _should_ be translatable, but the translation should be "Genesis Custom Blocks" in every language.

So, yeah, switch to a `__()` here.

php/admin/class-migration.php
```diff
@@ -0,0 +1,207 @@
...
+					);
+					?>
+				</p>
+				<p>
+					<?php
+					printf(
+						'<a target="_blank" href="https://getblocklab.com/docs/genesis-custom-blocks">%1$s</a>',
```
lukecarbis: @RobStino This page doesn't exist yet. Do we have anything we can chuck up there, or do we at least have this page marked as a to do on your go to market list? If so, let's make sure the URL is correct.

php/admin/class-migration.php
```diff
@@ -0,0 +1,207 @@
...
+	 * Handles an AJAX request to not display the notice.
+	 *
+	 * This stores in the user meta the fact that the notice was dismissed,
+	 * so it's not displayed again.
+	 */
+	public function ajax_handler_migration_notice() {
+		check_ajax_referer( self::NOTICE_NONCE_ACTION, self::NOTICE_NONCE_NAME );
+
+		if ( ! current_user_can( self::NOTICE_CAPABILITY ) ) {
+			wp_die( -1 );
```
lukecarbis: Why use `wp_die()` over [`wp_send_json_error()`](https://developer.wordpress.org/reference/functions/wp_send_json_error/)?

lukecarbis (changes requested):
Looking good Ryan! Couple of small questions, and minor change requested.

lukecarbis:
Tested, and everything works as expected.

I'll let @RobStino comment on the design side of things, I suspect there's a few tweaks he'd like to make.

php/admin/class-migration.php
RobStino: No, this page doesn't exist yet, but I've got it marked as a to-do. I'll just need to make sure I use the correct URL.

php/admin/class-migration.php
kienstra: Good question, I'm fixing it 😄

kienstra: Fixed in https://github.com/getblocklab/block-lab/pull/635/commits/d2ce757313994d0d487d55cbc9c17afe8d9f4e95

kienstra:
Hi @lukecarbis,
Thanks a lot for reviewing this, good points. They're applied above.

php/admin/class-migration.php
```diff
@@ -0,0 +1,210 @@
...
+	public function ajax_handler_migration_notice() {
+		check_ajax_referer( self::NOTICE_NONCE_ACTION, self::NOTICE_NONCE_NAME );
+
+		if ( ! current_user_can( self::NOTICE_CAPABILITY ) ) {
+			wp_send_json_error();
+		}
+
+		update_user_meta( get_current_user_id(), self::NOTICE_USER_META_KEY, self::NOTICE_DISMISSED_META_VALUE );
+
+		wp_die( 1 );
```
lukecarbis: This should be `wp_send_json_success()`, I think.

lukecarbis (approved):
👍 from me. Just one minor suggestion, but no need for further review from me after that.

Would love @RobStino's signoff on this before it's merged, though.

php/admin/class-migration.php
kienstra: Right, usually AJAX callbacks would use that.

This is a little different in that it doesn't send anything back to the [JS logic that called it](https://github.com/getblocklab/block-lab/pull/635/files#diff-682871fb6fb8d69d6b84890ae345f685R16).

lukecarbis: I understand that we wouldn't be using any success response. It might be more efficient to just `wp_die()`.

Either way, we should try to be consistent. Use `wp_die()` for both the error and success cases (as it was prior to my `wp_send_json_error()` suggestion), or use `wp_send_json_success()` here. We don't have to provide any data, we can call that function without any parameters.

The only difference is that the unused output would be:
`{'success':true}`

Instead of:
`1`

kienstra: Sounds good, how about just `wp_die()` for both?

kienstra: Good point that it's still sending `1` or `-1` in the response.

kienstra: https://github.com/getblocklab/block-lab/pull/635/commits/4e8167a4e178d46c4bd26852b73e6663183f1667 changes this to `wp_send_json_success()`

lukecarbis: Thanks for 4e8167a – looks good. Would have been happy with `wp_die()`, but I like this approach better, personally, because it feels consistent with other AJAX responses.

kienstra: Thanks, it is more like other AJAX calls 😄

RobStino:
Hey Ryan. Thanks for the work on this.
If we can make some small adjustments to the design of the notice, that would be great. I'm keeping it super simple and inline with standard WP admin notifications.
Specific call outs:
- WP Admin Button styling to 2 primary actions
- Reduction of vertical spacing between text and "learn more" link
See below for my mock
image

The second stage of the notification for when someone clicks "not now" is good as is.

kienstra:
Hi @RobStino,
Thanks, that looks good. I'll ping you here when it's applied.

kienstra:
Hi @RobStino,
Thanks, I think your suggestion looks better.

How does it look with https://github.com/getblocklab/block-lab/pull/635/commits/0f1c6ea248101ac2b46e37ef2c995cb7af5b9573?

### Narrow screen

narrow-width

### Wider screen

wider-screen

There's no change in the 'Not now' notice:

not-now-notice

Also, I [added](https://github.com/getblocklab/block-lab/pull/635/files#diff-24db437d8589eed1f645b12ea7e577a0R92) a `@todo` as a note to myself to check the link.

lukecarbis:
I think it needs a touch more padding to the left of the Not Now button.

RobStino:
Hey. Could you add 16px more margin to the left of the "Not Now" button?

kienstra:
Sure!

kienstra:
@RobStino,
Good idea. How does it look now with https://github.com/getblocklab/block-lab/pull/635/commits/4137f1addd19f0bb1f2277fdd8b05e910bdecc2d?

notice-margin-left-full-screen

RobStino (approved):
Looks good to me! Thanks for the work.

kienstra:
Nice, thanks for reviewing it!

PR #653, opened 2020-07-20
In the migration, install and activate Genesis Custom Blocks

kienstra:
#### Changes
* As John suggested, changes the order of the GCB installation and block migration
* Installs and activates GCB via a REST API endpoint
* The endpoint is mainly taken from the Gutenberg [plugin installation route](https://github.com/WordPress/gutenberg/blob/8d64aa3092d5d9e841895bf2d495565c9a770238/lib/class-wp-rest-plugins-controller.php#L417)
* Ideally the JS file would just use that endpoint, but it's only in Gutenberg, not in Core

#### Testing instructions (updated 23 July)

1. Checkout this PR's branch, and also checkout https://github.com/studiopress/genesis-custom-blocks/pull/13, but don't activate GCB it yet
2. Go to `/wp-admin` > Block Lab > Migrate, and go through the process
3. Expected: 
expected-diff

4. Change the slug to `amp`, as that's available on wp.org to download:

```diff
diff --git a/php/admin/migration/class-api.php b/php/admin/migration/class-api.php
index 9d4b3ae..7e36d98 100644
--- a/php/admin/migration/class-api.php
+++ b/php/admin/migration/class-api.php
@@ -45,7 +45,7 @@ class Api extends Component_Abstract {
         * Api constructor.
         */
        public function __construct() {
-               $this->new_plugin_slug = 'genesis-custom-blocks';
+               $this->new_plugin_slug = 'amp';
                $this->new_plugin_file = "{$this->new_plugin_slug}/{$this->new_plugin_slug}.php";
        }
```
GCB is not on wp.org of course, so we can't really test downloading and installing it.

5. Go through the migration flow just like before
6. Expected: The plugin should have been installed and activated:

plugin-activated

7. To reverse the migration, you might need to do this:
```sh
wp shell
( new Block_Lab\Admin\Migration\Post_Content( 'amp', 'block-lab' ) )->migrate_all();
```

8. Here are more steps to [reverse the migration](https://github.com/getblocklab/block-lab/pull/629#issuecomment-655846402)

9. If you have time, testing [these steps](https://github.com/getblocklab/block-lab/pull/629#issue-439492116) again would help, verifying the 'Expected' steps 8 and 9

Thanks for your patience 😄


mindctrl:
One thing I just thought about with this migration flow. The activate plugin step is last. That means if that step fails, the customer's site is left in a broken state. Is there a reason for not installing the new plugin first? I saw mention of bootstrapping conflicts. Is that difficult to solve here? We had a similar issue in AB with AB Pro/GPB, but solved it in the main plugin file by moving most of the logic to a secondary loader file.

Even with the conflicts resolved, this still assumes the plugin will install and activate successfully _after_ the content is migrated. It would be nice to verify that you can complete the required steps before updating all the content.

kienstra:
Hi @mindctrl,
> Is there a reason for not installing the new plugin first? I saw mention of bootstrapping conflicts. Is that difficult to solve here? 

Great idea. That should be possible with https://github.com/studiopress/genesis-custom-blocks/pull/11. I'll look at it.

kienstra:
Hi @mindctrl 
What do you think about this?

1. Exchange steps 4 and 5
2. The step 'Install And Activate Genesis Custom Blocks' would install and activate the plugin with the REST API. So there'd be no reload. If either failed, we could prevent migrating in the next step:
exchanging-steps-here

3. The 'Migrate your Blocks' step would still have the CPT and `post_content` migration. But it wouldn't install GCB anymore. On success, it would have a button 'Go to Genesis Custom Blocks' that would deactivate BL and go to `/wp-admin` > Custom Blocks:

migration-here

mindctrl:
Hi @kienstra 

> If either failed, we could prevent migrating in the next step:

I like the idea of changing the order and confirming GCB can be and is successfully installed before migrating. That's probably the safest thing to do I think.

kienstra:
Thanks, John! Good idea to check the plugin can be activated first.

kienstra:
Hi @lukecarbis and @RobStino,
What do you think about https://github.com/getblocklab/block-lab/pull/653#issuecomment-662029626? 

As John brought up, it could prevent users from having errors in blocks if the plugin installation and activation don't work.

lukecarbis:
Great idea!

kienstra:
Nice, thanks! I'll apply that tomorrow.

kienstra:
Here's the notice that Block Lab displays if GCB is also active:

block-lab-displays-notice

>It looks like Block Lab is active. Please deactivate it or migrate, as it will not work while Genesis Custom Blocks is active.

Block Lab won't load its helper functions, like `block_value()` and `block_field()`. But it'll still load all other classes, and all of the menus will work.

php/admin/migration/class-post-type.php
```diff
@@ -84,20 +85,23 @@ public function __construct( $previous_post_type_slug, $previous_block_namespace
...
-		$success_count = 0;
-		$error_count   = 0;
+		$posts              = $this->query_for_posts();
+		$success_count      = 0;
+		$error_count        = 0;
+		$max_allowed_errors = 20;
+		$errors             = new WP_Error();
 
-		while ( ! empty( $posts ) ) {
+		while ( ! empty( $posts ) && $error_count < $max_allowed_errors ) {
```
kienstra, line 97: Like with the post content, this now exits if there are 20 or more errors.

php/admin/migration/class-submenu.php
```diff
@@ -36,23 +36,14 @@ class Submenu extends Component_Abstract {
...
-	const NONCE_ACTION_DEACTIVATE_AND_ACTIVATE = 'deactivate_bl_and_activate_new';
-
-	/**
-	 * The query var to disable onboarding in Genesis Custom Blocks.
-	 *
-	 * The user will have already created blocks, so onboarding doesn't seem useful.
-	 *
-	 * @var string
-	 */
-	const QUERY_VAR_DISABLE_ONBOARDING = 'disable_onboarding';
```
kienstra, line 55: This shouldn't be needed, as this activates GCB 'silently' via `activate_plugin()`, so it won't run the activation hook.

kienstra:
Hi @lukecarbis,

Thanks for your patience with all of this 😄

php/admin/migration/class-post-type.php
kienstra: Like you brought up before, if there are several errors, this can be an infinite loop.

lukecarbis (approved):
Amazing, as always. Wish I could find something to fault you for.

kienstra:
Thanks, Luke! Thanks a lot for your patience with all of the changes here 😄

PR #660, opened 2020-07-27
Handle migration of thousands of posts

kienstra:
#### Changes
* If there are thousands of posts to migrate, the `post_content` migration request will probably time out
* I tested migrating 10,000 posts with BL blocks, and it took about 8 minutes
* So if the request times out, keep making it again until either it succeeds, or fails with a code other than `invalid_json`

#### Testing instructions
Same as the steps in https://github.com/getblocklab/block-lab/pull/629#issue-439492116

### DIFF
--- a/js/migration/components/steps/migrate-blocks.js
+++ b/js/migration/components/steps/migrate-blocks.js
@@ -70,16 +70,23 @@ const MigrateBlocks = ( { isStepActive, isStepComplete, stepIndex } ) => {
 	 * Migrates the post content.
 	 */
 	const migratePostContent = async () => {
+		const timeoutErrorCode = 504;
+
 		await apiFetch( {
 			path: '/block-lab/migrate-post-content',
 			method: 'POST',
 		} ).then( () => {
 			speak( __( 'The migration was successful!', 'block-lab' ) );
 			setIsSuccess( true );
-		} ).catch( ( result ) => {
+		} ).catch( async ( result ) => {
+			if ( result.hasOwnProperty( 'code' ) && timeoutErrorCode === result.code ) {
+				await migratePostContent();
+			}
+
 			if ( result.hasOwnProperty( 'message' ) ) {
 				setErrorMessage( result.message );
 			}
+
 			speak( __( 'The migration failed in the post content migration', 'block-lab' ) );
 			setIsError( true );
 		} );
--- a/js/migration/components/steps/migrate-blocks.js
+++ b/js/migration/components/steps/migrate-blocks.js
@@ -70,7 +70,8 @@ const MigrateBlocks = ( { isStepActive, isStepComplete, stepIndex } ) => {
 	 * Migrates the post content.
 	 */
 	const migratePostContent = async () => {
-		const timeoutErrorCode = 504;
+		// Used for a 504 Gateway Timeout Error, but could also be for other errors.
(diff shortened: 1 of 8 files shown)


js/migration/components/steps/migrate-blocks.js
```diff
@@ -70,16 +70,22 @@ const MigrateBlocks = ( { isStepActive, isStepComplete, stepIndex } ) => {
...
 			path: '/block-lab/migrate-post-content',
 			method: 'POST',
 		} ).then( () => {
 			speak( __( 'The migration was successful!', 'block-lab' ) );
 			setIsSuccess( true );
-		} ).catch( ( result ) => {
-			if ( result.hasOwnProperty( 'message' ) ) {
+		} ).catch( async ( result ) => {
+			if ( result.hasOwnProperty( 'code' ) && timeoutErrorCode === result.code ) {
+				await migratePostContent();
```
kienstra, line 84: Migrating ~10,000 posts took about 8 minutes, and the REST API request to migrate the `post_content` timed out after a minute:

timed-out

So this looks for the `invalid_json` code that a timeout has, and makes the request again. 

The user won't see that the request failed and that it's making a new one.

There's some risk that something else causes an `invalid_json` code, and this keeps looping and never shows an error message. That wouldn't do any damage to posts, but it'd be a bad UX.

The tradeoff is that we can now migrate thousands of posts.

My local migrated about 1250 posts per minute.

php/admin/migration/class-post-content.php
```diff
@@ -109,13 +109,10 @@ public function migrate_single( $post_id ) {
...
 		}
 
-		$replacement_count = 0;
-		$new_post_content  = preg_replace(
+		$new_post_content = preg_replace(
 			'#(<!--\s+wp:)(' . sanitize_key( $this->previous_block_namespace ) . ')(/[a-z][a-z0-9_-]*)#s',
 			'$1' . sanitize_key( $this->new_block_namespace ) . '$3',
-			$post->post_content,
-			-1,
-			$replacement_count
```
kienstra, line 118: `$replacement_count` isn't used

php/admin/migration/class-post-type.php
```diff
@@ -137,7 +137,7 @@ public function migrate_single( WP_Post $post ) {
 		$block = json_decode( $post->post_content, true );
 		if ( JSON_ERROR_NONE !== json_last_error() || empty( $block ) ) {
 			return new WP_Error(
-				'invalid_json',
+				'block_invalid_json',
```
kienstra, line 140: `'invalid_json'` is also used when the server returns HTML instead of JSON.

kienstra:
Hi @lukecarbis,
Good to see you earlier. 

Could you please review this? It allows migrating thousands of posts. 

Thanks, Luke!

kienstra:
https://github.com/getblocklab/block-lab/pull/660/commits/61cf280c3d821cd0f8f5781b647f8f3d970f80cf applie's Katies suggestions in https://github.com/getblocklab/block-lab/pull/629#issuecomment-665271278

lukecarbis (approved):

kienstra:
There's a support topic from a users with ~3,000 posts: 
https://wordpress.org/support/topic/migration-issue-to-genesis-blocks/

PR #673, opened 2020-08-07
Bump BL version to 1.5.6

kienstra:
#### Changes
* Bumps BL to `1.5.6`
* Uses Nick's fix of the `isLarge` prop warning
* Doesn't include any of the migration to GCB, that's in a separate branch
* Updates almost all `npm` packages
* No new feature
* Heres [the diff](https://github.com/getblocklab/block-lab/compare/v1.5.5...bump/1.5.6) since the last release. A lot of it is new JS component and e2e tests, which of course aren't deployed to wp.org.

#### Testing instructions
1. `rm js/editor.blocks.asset.php js/editor.blocks.js css/blocks.editor.css` # remove built files, to ensure this is using the most recently built versions, not falling back to previous builds
2. `npm install && npm run dev`
3. Regression testing of plugin, especially in the block editor. It looked good in my testing.

PR #684, opened 2020-08-13
Hide the meta box arrows, as they don't work in Edit Block

kienstra:
@RobStino, thanks for [pointing this out](https://github.com/getblocklab/block-lab/pull/673#issuecomment-673273198).

#### Changes
* It looks like WP 5.5 has arrows in meta boxes that allow moving them:

arrows-here-

* They can cause the main block config editor to be in the side, where there isn't enough room:

clicking-twice

* It's important to fix this now. If users change the order, they won't be able to change it back once this PR is deployed. That's because moving the meta boxes persists across page loads, and this PR hides the arrows (though they could remove the `display: none` style rule and move them back).

#### Testing instructions
1. Go to `/wp-admin` > Block Lab > All Blocks, and click a block
2. Expected: there aren't any more arrows in the meta boxes
3. Expected: the styling of the rest of the meta box still looks good

### DIFF
--- a/css/admin.block-post.css
+++ b/css/admin.block-post.css
@@ -30,6 +30,10 @@
 	padding: 6px 10px;
 	margin-bottom: 6px;
 }
+.handle-order-higher,
+.handle-order-lower {
+	display: none
+}
 #block_fields a,
 #block_fields a:focus {
 	outline: none;
--- a/css/admin.block-post.css
+++ b/css/admin.block-post.css
@@ -32,7 +32,7 @@
 }
 .handle-order-higher,
 .handle-order-lower {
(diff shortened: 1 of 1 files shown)


kienstra:
Hi @RobStino and @lukecarbis,
Could you please review this? It'd be good to merge this before releasing Block Lab to wp.org.

Thanks!

lukecarbis (approved):
lgtm

lukecarbis:
Let's make sure we get this fix in GCB as well.

kienstra:
Sounds good, I also committed it to an open PR for GCB.

